PR #3, opened 2019-10-04
Better development env

slawosz:
More flexible configuration for development, with possibility to switch between docker or standalone services.

Additionally, adds `pry` gem for easier debugging.

### DIFF
--- a/psd-web/.env.development.example
+++ b/psd-web/.env.development.example
@@ -9,4 +9,4 @@ KEYCLOAK_CLIENT_SECRET=
 
 REDIS_URL=redis://localhost:6379
 
-ELASTICSEARCH_URL=http://elasticsearch:9200
+ELASTICSEARCH_URL=http://localhost:9200
--- a/psd-web/.env.development.example
+++ b/psd-web/.env.development.example
@@ -1,7 +1,7 @@
 NOTIFY_API_KEY=
 NOTIFY_SMS_TEMPLATE_ID=
 
-DATABASE_URL=postgres://localhost/cosmetics_dev?user=postgres
+DATABASE_URL=postgres://localhost/psd_dev?user=postgres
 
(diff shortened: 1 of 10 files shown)


slorek:
I think we should use either `REDIS_URL` or `REDIS_HOST`, but not both (seems like `_URL` is not used). Same with the database config. Production uses URLs for everything so I think we should use that, e.g. `REDIS_URL`, `DATABASE_URL`, `ELASTICSEARCH_URL`.

psd-web/.env.test.example
```diff
@@ -0,0 +1,13 @@
...
+DATABASE_URL=postgres://localhost/psd_test?user=postgres
+
+KEYCLOAK_AUTH_URL=http://localhost:8080/auth
+KEYCLOAK_CLIENT_ID=psd-app
+KEYCLOAK_CLIENT_SECRET=
+
+REDIS_URL=redis://localhost:6379
+REDIS_HOST=localhost
+
+ELASTICSEARCH_URL=http://elasticsearch:9200
```
StupidCodeFactory:
```suggestion
ELASTICSEARCH_URL=http://localhost:9200
```

psd-web/.env.development.example
```diff
@@ -0,0 +1,12 @@
...
+
+DATABASE_URL=postgres://localhost/psd_dev?user=postgres
+
+KEYCLOAK_AUTH_URL=http://localhost:8080/auth
+KEYCLOAK_CLIENT_ID=psd-app
+KEYCLOAK_CLIENT_SECRET=
+
+REDIS_URL=redis://localhost:6379
+
+ELASTICSEARCH_URL=http://elasticsearch:9200
```
StupidCodeFactory:
```suggestion
ELASTICSEARCH_URL=http://localhost:9200
```

PR #7, opened 2019-10-08
WIP: Review environment POC

slawosz:
Still need some cleaning. For now is quite dangerous, as every restart removes db - not to run on production.

`./psd-web/deploy-review.sh`

### DIFF
--- a/psd-web/app/models/business.rb
+++ b/psd-web/app/models/business.rb
@@ -4,7 +4,7 @@ class Business < ApplicationRecord
   include Documentable
   include AttachmentConcern
 
-  index_name [Rails.env, "businesses"].join("_")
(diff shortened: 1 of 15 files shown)


slawosz:
@slorek 
```
I think we can probably save a lot of time reusing the existing database instances and just creating a new database on there.
```
What you mean exactly? Not to create db when it is unnecessary? For now, we could make it optional, for later (env started by travis) it might be harder.

```
And we need to be able to set the Keycloak OAuth callback URL programmatically.
```
For now it looks like keycloak is receiving callback url with client request, so hopefully this step is not needed. I will review it though.

slawosz:
@slorek Actually, db selection on travis might be easy.

PR #10, opened 2019-10-11
24-use-redis-as-rails-cache-store-in-production

StupidCodeFactory:
https://trello.com/c/ORi61T9d/24-use-redis-as-rails-cache-store-in-production

## Description
Add a shared cache, it uses the Sidekiq one.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


infrastructure/ci/build-docker-images.sh
```diff
@@ -13,7 +13,7 @@ function docker_tag_exists {
 }
 
 echo "Logging into DockerHub"
-docker login -u $DOCKER_USERNAME -p $DOCKER_PASSWORD
+echo $DOCKER_PASSWORD docker login -u $DOCKER_USERNAME --password-stdin
```
slorek: Won't this break the build as this script is run by Travis?

psd-web/config/environments/production.rb
```diff
@@ -49,7 +49,9 @@
   config.log_tags = [:request_id]
 
   # Use a different cache store in production.
-  config.cache_store = :memory_store
+  config.cache_store = :redis_cache_store, {
+    url: Rails.application.config_for(:redis_queue)
```
slorek: I think we should rename `redis_queue` to `redis_queue_and_cache_store` or just `redis_store`.

infrastructure/ci/build-docker-images.sh
StupidCodeFactory: I doubt that's it -but that does not look right though it's missing a `|`

StupidCodeFactory: yes that is it 🤦‍♂

PR #17, opened 2019-10-15
Remove Visual Studio IDE config

slorek:
IDE config shouldn't be committed to the repository. Also I don't believe anybody in the current team uses Visual Studio.


StupidCodeFactory:
@doctorpod I believe you do use vs code. maybe add these folders to your global .gitignore
```shell
git config --global core.excludesfile ~/.gitignore
```

PR #18, opened 2019-10-15
WIP: Fix ElasticSearch indices not being created in tests

slorek:
This fixes ElasticSearch indices not being created before tests run, at the expense of longer startup time for the test suite.

Here's an example of failing tests: https://travis-ci.org/UKGovernmentBEIS/beis-opss-psd/jobs/598170140

### DIFF
--- a/psd-web/app/models/concerns/indexable.rb
+++ b/psd-web/app/models/concerns/indexable.rb
@@ -2,6 +2,6 @@ module Indexable
   extend ActiveSupport::Concern
 
   included do
-    [ENV.fetch("ES_NAMESPACE", "default_namespace"), Rails.env, self.class.to_s.gsub('::', '').downcase].join("_")
+    index_name [ENV.fetch("ES_NAMESPACE", "default_namespace"), Rails.env, self.to_s.gsub('::', '').downcase].join("_")
   end
 end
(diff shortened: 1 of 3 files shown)


psd-web/app/models/concerns/indexable.rb
```diff
@@ -2,6 +2,30 @@ module Indexable
...
+
+  def update_search_index
+    __elasticsearch__.update_document
+    self.class.__elasticsearch__.refresh_index! if Rails.env.test?
+  end
+
+  def delete_search_index
+    __elasticsearch__.delete_document
+    self.class.__elasticsearch__.refresh_index! if Rails.env.test?
   end
```
StupidCodeFactory: I'd prefer have module in the test folder, and include it at inside this module in the `test_helper.rb`

```ruby
module IndexRefresh
 def insert_search_index
    super
    self.class.__elasticsearch__.refresh_index!
  end

  def update_search_index
    super
    self.class.__elasticsearch__.refresh_index!
  end

  def delete_search_index
    super
    self.class.__elasticsearch__.refresh_index! if Rails.env.test?
  end
end
```

slorek: It’s just that code for the minute, I don’t think it should be necessary.

PR #21, opened 2019-10-16
bundle udpate all the things

StupidCodeFactory:
update all of the dependencies
## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


psd-web/.rubocop.yml
```diff
@@ -1,2 +1,24 @@
...
+  Enabled: false
+
+Style/TrailingCommaInHashLiteral:
+  Enabled: false
+
+Style/TrailingCommaInArguments:
+  Enabled: false
+
+Style/StringLiterals:
+  EnforcedStyle: single_quotes
```
slorek: 😱

slorek: What's the rationale for this? Seems to requite a lot of file changes.

StupidCodeFactory: Contrary to rubocop's documentation which claims it defaults to enforcing single quotes for string literals, it actually enforced double quote which led to as much change as enforcing single quotes. 
The other option would to not enforce this cop but I would rather have it and also enforcing single quotes (less clutter visually to my eyes and I don't have to press the shift key).  
I'm not massively fussed about either way though.
What's your preference @doctorpod @slorek ?

slorek: If we're going to enforce one (which we probably should) I would prefer double quotes because in my experience that's what most teams do these days.

See https://anti-pattern.com/always-use-double-quoted-strings-in-ruby and https://www.viget.com/articles/just-use-double-quoted-ruby-strings/ - these posts pretty much started the trend.

I also used to prefer single until then, but then you do get a mix of single and double throughout the codebase (i.e. when string interpolation is useful), so after a period of adjustment double quotes is my default now.

PR #24, opened 2019-10-17
Fix product listing page

StupidCodeFactory:
**CHANGES:**
1. make countrie not read from the cache - Just hold it in memory
2. fix rails cache store config

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


psd-web/app/models/country.rb
```diff
@@ -0,0 +1,11 @@
+class Country
+
+  PATH_TO_COUNTRIES_LIST =
+    'node_modules/govuk-country-and-territory-autocomplete/dist/location-autocomplete-canonical-list.json'.freeze
```
StupidCodeFactory: This is a bit 😞 but hey 🤷‍♂

PR #32, opened 2019-10-23
Sub nav component, applied to investigation show page

samlionheart:
- Added sub nav component to shared-web
- Applied it to the case page
Sub_Nav_3

- Fixed "Unassigned" being the wrong font size


edwardhorsford:
Looks like the attachments tab has the url `/documents` - can this be made to match the tab name?

I don't think I'd intended to change 'Activities' to 'Timeline' as part of this either. Though it is something I'd like to do at some point.

psd-web/app/controllers/investigations_controller.rb
```diff
@@ -35,6 +35,7 @@ def index
   # GET /cases/1
   # GET /cases/1.json
   def show
+    @current = request.fullpath.split("/").last
```
StupidCodeFactory: I'm not sure what you're trying to do here? wouldn't `param[:id]` do the trick?

StupidCodeFactory: also `@current` is really not a great variable name. Is it an investigation as the controller name would suggest?

psd-web/app/helpers/investigations/display_text_helper.rb
```diff
@@ -3,6 +3,26 @@ def image_document_text(document)
...
+
+  def investigation_sub_nav_tabs
+    mapping = {
+      products: :products,
+      businesses: :businesses,
+      attachments: :attachments,
+      activity: :activity
+    }
+
+    render "investigations/tabs/" + mapping.fetch(@current.to_sym, :overview).to_s
```
StupidCodeFactory:
```suggestion
    render "investigations/tabs/" + @current || 'overview')
```
and remove the `mapping`

shared-web/app/views/components/_hmcts_sub_nav.html.slim
```diff
@@ -0,0 +1,7 @@
+
+nav.hmcts-sub-navigation
+  ul.hmcts-sub-navigation__list
+    - items.each do |item|
+      li.hmcts-sub-navigation__item
+        a.hmcts-sub-navigation__link aria-current=("page" if item[:active]) href=item[:href] id="#{item[:text]}_id"
+          = item[:html] || item[:text] + "#{item[:count]}"
```
StupidCodeFactory: these two files could probably in psd-web rather than shared-web

slorek:
Can you remove the extra commits already in master again on this branch please?

StupidCodeFactory:
@samlionheart @slorek I'm closing this.
It's included in this PR https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/37
I've had to make a few tweaks the way the JS was loaded, i.e not in standard rails place.

PR #35, opened 2019-10-24
Change default OPSS team

doctorpod:
## Description
Ticket: https://trello.com/c/6TOv2SJI

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/spec/models/team_spec.rb
+++ b/psd-web/spec/models/team_spec.rb
@@ -1,8 +1,6 @@
 require "rails_helper"
 
 RSpec.describe Team do
-  subject { described_class }
-
   describe ".get_visible_teams" do
     before do
       allow(Rails.application.config).to receive(:team_names).and_return(
@@ -24,7 +22,7 @@ RSpec.describe Team do
       let(:user) { double("User", is_opss?: true) }
 
       it "returns all important teams" do
-        expect(subject.get_visible_teams(user).map(&:name)).to eq(important_team_names)
+        expect(described_class.get_visible_teams(user).map(&:name)).to eq(important_team_names)
       end
     end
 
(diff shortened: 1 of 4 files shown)


slorek:
Can you add a test for this behaviour?

psd-web/spec/models/team_spec.rb
```diff
@@ -0,0 +1,39 @@
+require "rails_helper"
+
+RSpec.describe Team do
+  subject { described_class }
```
StupidCodeFactory: nope `subject` !== `described_class`

just use `described_class` in the test instead of `subject` 
By convention, a `subject` is an instance of `described_class`, almost 99%. 
Sometime in controller tests  subject is assigned the return value, i.e `response` of the HTTP call, that's really almost the only time you a `subject` _is not_ and instance of `described_class`

psd-web/spec/models/team_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    let(:important_team_names) do
+      %w{bish bosh bash}
+    end
+
+    context "OPSS user" do
+      let(:user) { double("User", is_opss?: true) }
+
+      it "returns all important teams" do
+        expect(subject.get_visible_teams(user).map(&:name)).to eq(important_team_names)
```
StupidCodeFactory:
```suggestion
        expect(described_class.get_visible_teams(user).map(&:name)).to eq(important_team_names)
```

psd-web/spec/models/team_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      end
+    end
+
+    context "Non-OPSS user" do
+      let(:user) { double("User", is_opss?: false) }
+
+      it "returns first important team" do
+        expect(subject.get_visible_teams(user).map(&:name)).to eq([important_team_names.first])
```
StupidCodeFactory:
```suggestion
        expect(described_class.get_visible_teams(user).map(&:name)).to eq([important_team_names.first])
```

PR #37, opened 2019-10-24
Reformat case overview page

StupidCodeFactory:
**CHANGES:**
1. Subnav component
  I've added a few `#show` or `#index` to the relevant controller, it's a bit more resilient implementation than the previous one as it does not try to second guess what page we're on base on the URL.
2. refactored of overview page with Ed. 
    I've added Draper gem which makes it **much** easier to tests fiddly HTML component with some logic without having to resort to slow request/controller/integration tests.
3. fix the asset pipeline by placing the assets in the standard location and unpinned sprocket.(that's where most of the changes come)
4. Also removed the `def model_name` hack in `Investigation` subclasses, explicitly passing investigation_path were needed in views and controller redirects, and `scope: : investigation` in forms.
5. Fixed a few flaky system tests by using capybara aware `assert_*` methods. 
https://www.rubydoc.info/gems/capybara/Capybara/Minitest/Assertions#assert_css-instance_method

review app:https://psd-pr-37-web.london.cloudapps.digital/

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


psd-web/app/views/components/_govuk_summary_list.html.slim
```diff
@@ -0,0 +1,19 @@
+- classes ||= ""
+- dl_wrapper_class = ["govuk-summary-list"]
+- dl_wrapper_class << classes
+- dl_wrapper_class
```
StupidCodeFactory: I know this is very gross, @edwardhorsford wants to have this component match the GDS javascript API one. 
Let's address this when we get to implement it.

psd-web/test/fixtures/organisations.yml
```diff
@@ -0,0 +1,7 @@
+south_hampton:
```
slorek: Southampton is one word.

psd-web/test/fixtures/organisations.yml
```diff
@@ -0,0 +1,7 @@
...
+  name: Southamptom Council
```
slorek: *Southampton

psd-web/test/fixtures/organisations.yml
```diff
@@ -0,0 +1,7 @@
...
+  path: /Organisations/Southamptom Council
```
slorek: *Southampton

psd-web/test/fixtures/users.yml
```diff
@@ -0,0 +1,11 @@
+south_hampton:
```
slorek: *`southampton`

psd-web/bin/rails
```diff
@@ -1,4 +1,9 @@
 #!/usr/bin/env ruby
+begin
```
slorek: Not sure on loading `spring` by default especially as these scripts is used for the production processes.

psd-web/app/views/components/_govuk_summary_list.html.slim
edwardhorsford: More broadly, if we're aiming for a rails port of the design system, the api should be compatible.

psd-web/bin/rails
StupidCodeFactory: Spring is only loaded install in `developement` hence won't affect production.
See https://github.com/rails/spring/blob/ca86653ba6d784f0348d5dd0280e030dd690af19/lib/spring/application.rb#L182-L188

psd-web/app/views/components/_govuk_summary_list.html.slim
frankieroberto: Feel free to contribute a `govukSummaryList` helper to https://github.com/UKGovernmentBEIS/govuk-design-system-rails/tree/master/app/helpers/govuk_design_system 😄

psd-web/app/decorators/investigation/allegation_decorator.rb
```diff
@@ -0,0 +1,31 @@
+class Investigation < ApplicationRecord
```
slorek: 🤔 is it necessary to monkey-patch the `Investigation` class in order to namespace the decorator? The corresponding models are defining the namespace in the class name, which I think is safer if not ideal.

psd-web/app/decorators/investigation/enquiry_decorator.rb
```diff
@@ -0,0 +1,4 @@
+class Investigation < ApplicationRecord
```
slorek: Do we need empty decorator classes?

psd-web/app/helpers/application_helper.rb
```diff
@@ -1,4 +1,8 @@
 module ApplicationHelper
+  def govuk_hr
```
slorek: This would be more appropriate as a helper in https://github.com/UKGovernmentBEIS/govuk-design-system-rails

psd-web/app/helpers/documents_helper.rb
```diff
@@ -3,6 +3,9 @@ module DocumentsHelper
 
   def set_parent
     @parent = Investigation.find_by!(pretty_id: params[:investigation_pretty_id]) if params[:investigation_pretty_id]
+    @parent = Investigation.find_by!(pretty_id: params[:allegation_id]) if params[:allegation_id]
+    @parent = Investigation.find_by!(pretty_id: params[:project_id]) if params[:project_id]
+    @parent = Investigation.find_by!(pretty_id: params[:inquiry]) if params[:inquiry]
```
slorek: Seems like a lot of redundant database queries are going to be generated here. Is it possible to refactor to escape early?

psd-web/app/models/investigation/enquiry.rb
```diff
@@ -1,42 +1,36 @@
-class Investigation::Enquiry < Investigation
-  include DateConcern
+class Investigation < ApplicationRecord
```
slorek: Same comment regarding monkey-patching as above. I think this is unsafe. Ideally we should have a module `Investigation`, but I understand that's not possible without renaming the `Investigation` class. But I don't think we should override the class definition elsewhere.

psd-web/app/models/investigation/project.rb
```diff
@@ -1,23 +1,17 @@
...
+      "project"
+    end
 
-private
+  private
 
-  def create_audit_activity_for_case
-    AuditActivity::Investigation::AddProject.from(self)
+    def create_audit_activity_for_case
+      AuditActivity::Investigation::AddProject.from(self.decorate)
```
slorek: Should we pass the decorated version here or have the recipient call `decorate` if it needs it?

psd-web/app/models/user.rb
```diff
@@ -84,7 +84,7 @@ def name
   end
 
   def display_name(ignore_visibility_restrictions: false)
-    display_name = name
+    display_name = name || ""
```
slorek: I think you got this from merging my changes, but it is out of date and should be removed.

psd-web/app/views/components/_govuk_summary_list.html.slim
```diff
@@ -0,0 +1,20 @@
+ruby:
```
slorek: I think we would prefer to write new templates as ERb as part of a gradual migration from Slim, it's not necessary this time but to bear in mind in future.

psd-web/bin/rails
```diff
@@ -1,4 +1,9 @@
...
+  load File.expand_path('../spring', __FILE__)
+rescue LoadError => e
+  raise unless e.message.include?('spring')
```
slorek: This seems a bit clunky/risky. Is there a better way?

psd-web/bin/rspec
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env ruby
```
slorek: Can we have Spring loaded explicitly, perhaps with an env var? I find it a bit annoying personally. And it would prevent trying to load a (supposedly) non-existent library in production. This way you can load it if it's useful to you.

psd-web/test/fixtures/users.yml
```diff
@@ -0,0 +1,11 @@
+southampton:
+  email: southampton@example.com
```
slorek: Use `beis.gov.uk` for test fixtures as `example.com` is a real domain.

StupidCodeFactory: it's actually safer to use example.com has it is made for that purpose.
http://example.com/

psd-web/bin/rails
StupidCodeFactory: This is generated by rails when you install spring. I'll just remove it and have in my branches.

psd-web/app/decorators/investigation/enquiry_decorator.rb
StupidCodeFactory: Yes we do, has it's infered from the class name. Draper allows to explicitly decorate an object but that would mean adding more code in the controller to decorate with the correct decorator. 
With STI it's easier to have empty decorator.

psd-web/app/decorators/investigation/allegation_decorator.rb
StupidCodeFactory: It's not really monkey pathing. I'm just following the namespacing the standard way to namespace classes in rails

psd-web/app/models/user.rb
StupidCodeFactory: > Just some minor comments, no biggies.
> 
> I think the Draper gem looks useful and I'm happy to go with it, but we should have a follow-up team discussion about how we want to test the presentation layer. The Cosmetics team are going with request specs asserting text on the page. I'd like to have a consistent approach.

So asserting text on the page, make tests slower (capybara has to scan the entire HTML document), I've managed to trim down the time the tests take to run by a fairly significant time by switching `have_content` to `have_css` in several projects.
 I'm happy to go along with just asserting the text on the page but I find it less robust as you could have false positive (that has happened to me in the past several time and regressions where introduced.)

psd-web/app/decorators/investigation/allegation_decorator.rb
StupidCodeFactory: It is necessary and safer, otherwise you get TypeError superclass mismatch based on the load order of the file.

psd-web/app/helpers/application_helper.rb
StupidCodeFactory: ok I'll move that over to the other gem.

psd-web/test/fixtures/users.yml
slorek: TIL...

psd-web/app/models/user.rb
slorek: I think it is a good solution personally.

psd-web/app/helpers/documents_helper.rb
StupidCodeFactory: this won't make any extra db queries, only one of those params will be present.
This whole `document` is a bit of pain anyway it tries to be agnostic of the whole attachments but in reality it makes the code hard to understand work with. 

I could add `parent ||=`

psd-web/app/decorators/investigation/allegation_decorator.rb
slorek: It's the redefinition of `Investigation` inheriting from `ApplicationRecord` that concerns me. The class definition is being overridden in other files. Not a biggie though.

psd-web/app/models/investigation/enquiry.rb
StupidCodeFactory: it's just regular namespacing.

StupidCodeFactory: It's actually safer to it this way, due to how ruby lookup constants.

psd-web/app/models/investigation/project.rb
StupidCodeFactory: So I've asked myself the same question. 
In the first place, I would argue that this is misplaced and it is not the role of this class to know about auditing and even less about how to audit.
Ideally that's should done in some kind of `Command|Service` like class

psd-web/app/views/components/_govuk_summary_list.html.slim
StupidCodeFactory: yeah, we're going to have to resort to some dirty code to create these fancy highly customisable templates. 

I would argue we should just create the HTML as needed.

psd-web/test/fixtures/users.yml
edwardhorsford: `test.com` is a real domain, but afaik `example.com` is reserved.

edwardhorsford:
@nasirkhanbeis points out that we should comma separate the product categories rather than commas separated with `and` as the last item. This is because some of the categories themselves have `and` in them.

So: `a, b, c` rather than `a, b and c`.

StupidCodeFactory:
> @nasirkhanbeis points out that comma separating doesn't work well - as the source text includes commas already.
> 
> Suggest changing the product categories to be a [govuk-list](https://design-system.service.gov.uk/styles/typography/#lists).
> 
> Screenshot 2019-11-19 at 11 12 49

That's ineeded a good spot

psd-web/test/system/products_test.rb
```diff
@@ -0,0 +1,42 @@
+require "application_system_test_case"
```
slorek: We shouldn’t be adding any new system tests. They run so slowly and we should write new tests in Rspec as well. It looks like this could be easily translated to a feature spec.

psd-web/test/system/products_test.rb
```diff
@@ -0,0 +1,42 @@
...
+                    href: investigation_products_url(@investigation)
+      p.assert_link businesses_link,
+                    href: investigation_businesses_url(@investigation)
+      p.assert_link attachments_link,
+                    href: investigation_attachments_url(@investigation)
+      p.assert_link activity_link,
+                    href: investigation_activity_url(@investigation)
+    end
+
+    click_link products_link
```
slorek: I’m not sure these assertions are really adding a lot of value versus the amount of time four extra requests will take. Given we have already asserted the links are present on the page and they are matched to name routes can we safely assume that the routing works?

edwardhorsford:
I've spotted that the tabs on the business and products sections are not working - something up with the JS I guess.

psd-web/test/system/products_test.rb
StupidCodeFactory: At that time there was no RSpec installed so I went along with Minitest. I'll translate this to a feature test.

psd-web/test/system/products_test.rb
StupidCodeFactory: So I changed the tile HTML, it was to ensure the correct heading was there and that the links also worked. I can remove. It's indeed safe to assume those links works

StupidCodeFactory: @slorek  So I've tried to convert it. However, in your other PR, you've added all the goodies to log in and factory_bot. I've tried to rip some bits of it. It's not super practical. 
I think it makes sense to do that once at least you PR or both is/are merged. Would you mind going along with this for the time being?

slorek: Yes that makes sense.

psd-web/app/controllers/documents_controller.rb
```diff
@@ -23,17 +23,25 @@ def update
     return render :edit unless file_valid?
 
     @file.blob.save
-    AuditActivity::Document::Update.from(@file.blob, @parent, previous_data) if @parent.is_a? Investigation
-    redirect_to @parent
+    if @parent.is_a? Investigation
```
slorek: We have two different patterns next to each other here. Above we have:

`return render :edit unless file_valid?`

So we could do this here:

`return redirect_to @parent unless @parent.is_a? Investigation`

And remove the `if/else` block.

Generally I prefer this escape early pattern but I think it should at least be consistent.

Same in the `destroy` action.

psd-web/app/controllers/documents_flow_controller.rb
```diff
@@ -27,8 +27,12 @@ def update
     return redirect_to next_wizard_path unless step == steps.last
 
     attach_blobs_to_list(@file_blob, file_collection)
-    AuditActivity::Document::Add.from(@file_blob, @parent) if @parent.is_a? Investigation
-    redirect_to @parent
+    if @parent.is_a? Investigation
```
slorek: Same comment here as above; two different patterns used, escape early and if/else.

edwardhorsford:
As per standup, 'created by' still doesn't seem to work.

@nasirkhanbeis also notices that date created appears to be hardcoded.

psd-web/app/decorators/investigation_decorator.rb
```diff
@@ -0,0 +1,114 @@
...
+  def investigation_summary_list(include_actions: true, classes: "")
+    rows = [
+      {
+        key: { text: "Status", classes: classes },
+        value: { text: status, classes: classes },
+        actions: []
+      },
+      {
+        key: { text: "Created by", classes: classes },
+        value: { text: investigation.source&.name, classes: classes },
```
slorek: I think we should be using `display_name`?

psd-web/app/helpers/investigations/display_text_helper.rb
```diff
@@ -70,7 +104,7 @@ def correspondence_should_be_hidden(result, source, investigation)
 
   # rubocop:disable Rails/OutputSafety
   def investigation_assignee(investigation, classes = "")
-    out = [investigation.assignee ? investigation.assignee.name.to_s : "Unassigned"]
+    out = [investigation.assignee ? investigation.assignee.name.to_s : tag.div("Unassigned", class: classes)]
```
slorek: `display_name`.

psd-web/app/decorators/investigation_decorator.rb
slorek: There is also the `show` method on `UserSource`.

psd-web/spec/decorators/investigation_decorator_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+    end
+  end
+
+  describe "#investigation_summary_list" do
+    fixtures(:sources)
+    let(:investigation_summary_list) { subject.investigation_summary_list }
+
+    it "has the expected fields" do
+      expect(investigation_summary_list).to summarise("Status", text: investigation.status)
+      expect(investigation_summary_list).to summarise("Created by", text: investigation.source.name)
```
slorek: `source.show`.

psd-web/spec/decorators/investigation_decorator_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+      expect(investigation_summary_list).to summarise("Assigned to", text: /Unassigned/)
+      expect(investigation_summary_list).
+        to summarise("Date created", text: investigation.created_at.beginning_of_month.strftime("%e %B %Y"))
+      expect(investigation_summary_list).to summarise("Last updated", text: time_ago_in_words(investigation.updated_at))
+      expect(investigation_summary_list).to summarise("Trading Standards reference", text: investigation.complainant_reference)
+    end
+
+    context "whithout complainant reference" do
```
slorek: *without

psd-web/app/models/investigation.rb
```diff
@@ -229,6 +228,11 @@ def send_confirmation_email
       ).deliver_later
     end
   end
+
+
+  require_dependency "investigation/allegation"
```
slorek: Why is this needed?

psd-web/app/controllers/investigations/activities_controller.rb
```diff
@@ -1,74 +1,81 @@
...
-    when "visibility"
-      redirect_to visibility_investigation_path(@investigation)
-    when "alert"
-      redirect_to new_investigation_alert_path(@investigation)
-    else
-      @activity_type_empty = true
+      @breadcrumbs = {
+        items: [
+          { text: "Cases", href: investigations_path(previous_search_params) },
+          { text: @investigation.pretty_description }
```
slorek: Should `pretty_description` be in the decorator?

psd-web/config/initializers/date_and_time_formats.rb
```diff
@@ -0,0 +1,2 @@
+Time::DATE_FORMATS[:govuk] = "%e %B %Y"
```
slorek: I think these might be useful in the design system gem. But not necessary for now.

psd-web/app/decorators/investigation_decorator.rb
slorek: `source&.show`

psd-web/config/initializers/date_and_time_formats.rb
edwardhorsford: The design system doesn't include anything for date formatting. But yes we should have a govuk-date formatter.

StupidCodeFactory: so I can update the to the latest design system gem has it is not backward compatible. 
So things that should be in the design system gem, will stay in this repo until it is made compatible or we have the version of govuk frontend

psd-web/app/controllers/investigations/activities_controller.rb
StupidCodeFactory: Hm yes absolutely! I'll implement that change.

psd-web/app/models/investigation.rb
StupidCodeFactory: In dev, with Investigation.import at the bottom of the file it breaks the how classes are loaded. This ensures when finishing loading the Investigation model We also load the subclasses so it doesn't throw a TypeError.
I'd rather ditch the `import` and have run a rake task when I need to refresh the index.  rather than this hacky code to index at code loading time.

psd-web/app/controllers/documents_controller.rb
StupidCodeFactory: Yes, good spot. I prefer early return aas well

psd-web/app/controllers/investigations/activities_controller.rb
```diff
@@ -1,74 +1,82 @@
...
-    authorize @investigation, :show?
-  end
+    def set_investigation_with_associations
+      @investigation = Investigation
+                         .eager_load(:source,
+                                     products: { documents_attachments: :blob },
+                                     investigation_businesses: { business: :locations },
+                                     documents_attachments: :blob)
+                         .find_by!(pretty_id: params[:investigation_pretty_id])
+
```
slorek: In the `set_investigation` method above we call `investigation.decorate` but in this one we don't and it is called in the view instead. It's a minor inconsistency - is it possible to decorate it here instead of the view?

slorek:
Just a couple of minor comments. I still think we might need to use `display_name` for the assignee.

StupidCodeFactory:
> Just a couple of minor comments. I still think we might need to use `display_name` for the assignee.

so I used display_name but @edwardhorsford raised it as a bug so I've put it back the way it was.

psd-web/app/controllers/investigations/activities_controller.rb
StupidCodeFactory: Which view are you refering to?

edwardhorsford:
One thing I've noticed - all pages (previously tabs) have the same title "View case" - this is neither specific to the case nor specific to the page - both of which are needed for accessibility.

I think the titles should probably combine the h1 with the case type and case number. I suggest we leave as is for this PR and raise a separate ticket to cover remedial fixes.

edwardhorsford:
A minor bug - there's an empty excess `li`, and the `img` should be inside of the first `li`.

psd-web/app/helpers/investigations/display_text_helper.rb
StupidCodeFactory: @edwardhorsford raised a bug when I used to display name so reverted it back. He wants to use `display_name` as a separate tickets.

psd-web/app/controllers/investigations/activities_controller.rb
StupidCodeFactory: It's not decorated in the view. It was not decorated at all which caused a bug.

PR #38, opened 2019-10-24
Cache user roles

slorek:
## Description
* Cache user roles so we don't send unnecessary requests to Keycloak every time.
* Move Rspec tests to main app since we are not sharing the code anymore.

This will significantly reduce the number of requests made to Keycloak.

I'm not happy with the way the test suite changes work. The way the test suite stubs user roles, in combination with the use of `ActiveHash` which is read-only, makes it very difficult to alter the existing tests without resorting to stubbing of the Keycloak library.

I decided to move the Rspec suite to the main app because I needed to use various Rails libraries which we don't have available within the gem, and we are going to move `shared-web` into the main app anyway as we did on cosmetics. It means we can start writing Rspec tests instead.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


nasirkhanbeis:
@slorek could you please spin up review app if we need regression testing?

slawosz:
Looks good overall :+1:

psd-web/spec/shared/web/user_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    context "when the user's roles are not cached" do
+      before { expect(Shared::Web::KeycloakClient.instance).to receive(:get_user_roles).and_return(keycloak_roles) }
+
+      it "returns the roles from Keycloak" do
+        expect(user.roles).to eq(keycloak_roles)
+      end
+
+      it "caches the roles" do
+        user.roles
```
StupidCodeFactory:
```suggestion
        expect { user.roles }.to change { Rails.cache.read("user_roles_#{id}") }.from(nil).to(keycloak_roles)
```

shared-web/app/models/shared/web/user.rb
```diff
@@ -41,8 +43,13 @@ def self.current=(user)
       end
 
       def has_role?(role)
-        access_token = User.current.access_token if current_user?
-        KeycloakClient.instance.has_role?(id, role, access_token)
+        roles.include?(role)
+      end
+
+      def roles
+        @roles ||= Rails.cache.fetch("user_roles_#{id}", expires_in: 30.minutes) do
```
slawosz: In order to avoid code fragmentation and possible cache issues, I would create a wrapper. This will also help us when we will need to do cache invalidation, or change underlying cache engine.
```ruby
@roles ||= CacheWrapper.cache_user_roles(id, expires_in: 30.minutes) do
    KeycloakClient.instance.get_user_roles(id)
end
```
and then implementation:

```ruby
class CacheWrapper
  def self.cache_user_roles(id, expires_in: 30.minutes) 
    cache("user_roles_#{id}", expires_in) { yield }
  end
  
  def self.cache(key, expires_in)
    Rails.cache.fetch(key, expires_in: 30.minutes) { yield }
  end
end
```

slorek:
Review app: https://psd-cache-roles-web.london.cloudapps.digital

shared-web/app/models/shared/web/user.rb
slorek: @slawosz thanks, what does that provide which `Rails.cache` doesn't?

slawosz: We explicitly know where we cache, so as I wrote above, its easier to control. But its just a suggestion.

slorek: Thanks. Perhaps we can discuss this suggestion in person so I can understand it.

PR #39, opened 2019-10-25
Deploy rewiev app on PR

slawosz:


psd-web/Gemfile
```diff
@@ -16,7 +16,7 @@ gem "rubyzip"
 gem "scout_apm"
 gem "slim-rails"
 gem "sprockets", "= 4.0.0.beta10"
-gem "strong_migrations"
+gem "strong_migrations", "0.4.0"
```
StupidCodeFactory: what's the problem with 0.4.1 - it's been working great so far?

slawosz: It causes some error when migrations are being run from scratch.

PR #43, opened 2019-10-28
only run tests on CI

StupidCodeFactory:
Only run tests with coverage on CI 
## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


slorek:
Thanks! Can you provide some rationale for changing this (so that everyone knows why we did it)? Will this prevent the huge coverage block from being displayed at the end of the test output?

StupidCodeFactory:
@slorek  Apologies - This wasn't indeed very clear why. Yes, it does remove the annoying table displayed at the end of the test run.

PR #44, opened 2019-10-31
Fix teams array being null

StupidCodeFactory:
Fixes this sentry error
https://sentry.io/organizations/beis/issues/997569880/?environment=staging&project=1329381&referrer=slack

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/app/models/user.rb
+++ b/psd-web/app/models/user.rb
@@ -85,7 +85,7 @@ class User < Shared::Web::User
   end
 
   def team_names
-    teams.map(&:name).join(", ")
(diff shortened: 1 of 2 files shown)


slorek:
Thanks. This is a long running issue, but do we understand why we get a nil team in the first place? That's the question which prevented me from doing the quick fix.

slorek:
This should be fixed by https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/56

PR #45, opened 2019-10-31
Remove shared web

slorek:
## Description
This removes the `shared-web` Rails engine which was adding friction to the development process. We have already removed it from Submit Cosmetics Product Notifications in https://github.com/UKGovernmentBEIS/beis-opss/pull/1374.

Important: when this is merged, you will no longer need to copy `shared-web` into `vendor` manually. Make sure any changes you're working on separately within `shared-web` are retained when you rebase against this.

The GOV.UK Design System components have been extracted to a new Gem at https://github.com/UKGovernmentBEIS/govuk-design-system-rails.

I also reverted the `sprockets` gem version change which was introduced because there is a breaking change in v4.0.0 which requires a `manifest.js` file which we don't have. This caused failing tests.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


psd-web/Gemfile
```diff
@@ -38,10 +56,12 @@ group :development, :test do
   gem "rubocop"
   gem "rubocop-performance"
   gem "ruby-debug-ide"
+  gem "sass-rails", "~> 5.0"
   gem "selenium-webdriver"
   gem "simplecov"
   gem "simplecov-console"
   gem "slim_lint"
   gem "solargraph"
+  gem "sprockets", "3.7.2"
```
StupidCodeFactory: beta 10 works - I know this is related to brakeman complaining about a version under 4.0.0.beta8 that has a vulnerability issue.
But so far the beta10 was working. Has this changed?

psd-web/app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,39 @@
+class SessionsController < ApplicationController
+  skip_before_action :authenticate_user!
+
+  def new
+    redirect_to keycloak_login_url(request.original_fullpath)
+  end
+
+  def signin
+    request_and_store_token(auth_code, params[:request_path])
+    redirect_path = main_app.root_path
```
StupidCodeFactory:
```suggestion
```

psd-web/app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    redirect_path = params[:request_path] if is_relative(params[:request_path])
```
StupidCodeFactory:
```suggestion
    redirect_path = is_relative(params[:request_path]) ? params[:request_path] : root_path
```

psd-web/app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    redirect_to redirect_path
+  rescue RestClient::ExceptionWithResponse => e
+    redirect_to keycloak_login_url(params[:request_path]), alert: signin_error_message(e)
+  end
+
+  def logout
+    KeycloakClient.instance.logout(refresh_token)
+    redirect_to main_app.root_path
```
StupidCodeFactory:
```suggestion
    redirect_to root_path
```

psd-web/app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  def signin_error_message(error)
+    error.is_a?(RestClient::Unauthorized) ? "Invalid email or password." : JSON(error.response)["error_description"]
+  end
+
+  def auth_code
+    params.require(:code)
+  end
+
+  def is_relative(url)
+    url =~ /^\/[^\/\\]/
```
StupidCodeFactory: I'd rather have something like 

```ruby 
def allow_redirect?(url)
  ENV['HTTP_HOST'] == URI(url).host
end
```

psd-web/app/models/concerns/date_concern.rb
```diff
@@ -0,0 +1,176 @@
+module DateConcern
```
StupidCodeFactory: this entire file seems rather suspicious. I'd love to create a ticket and to examine this and rip it off and re-write it in a legible way.

psd-web/app/controllers/sessions_controller.rb
slorek: Let's improve it separately, I'm just copying out the files from shared-web as-is in this PR.

psd-web/app/controllers/sessions_controller.rb
slorek: Good spot thanks.

psd-web/Gemfile
slorek: Here's what happens when I add:

```ruby
gem "sprockets", "= 4.0.0.beta10"
```
to the Gemfile and run tests:

```ruby
Sprockets::Railtie::ManifestNeededError:
  Expected to find a manifest file in `app/assets/config/manifest.js`
  But did not, please create this file and use it to link any assets that need
  to be rendered by your app:

  Example:
    //= link_tree ../images
    //= link_directory ../javascripts .js
    //= link_directory ../stylesheets .css
  and restart your server
```

psd-web/app/models/concerns/date_concern.rb
slorek: Go ahead! :)

slorek:
Review app URL: https://psd-pr-45-web.london.cloudapps.digital

psd-web/Gemfile
StupidCodeFactory: 🤔 ok don't bother - I have all the assets stuff fixed in my reformat overview page. 
The config wasn't correct in the first place.

slorek:
This was approved by Nasir via Slack.

PR #47, opened 2019-10-31
Remove hack hijacking investigation subclasses model name

StupidCodeFactory:
now elasitcseaerch indices are manually provided which remove the need for this
hack.
This is inconveniant because this modifies routes/form generation.
I need this to be merge for the reformat overview page and make each tabs a
seperate action.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


StupidCodeFactory:
I'm closing this - it turns out to not be totally as straight forward as it should have been. Ill create a ticket to sort this out as tech debt

PR #48, opened 2019-10-31
Static analysis outside docker

slorek:
## Description
This runs the static analysis checks directly rather than inside the Docker context, which makes them much faster and gives us feedback sooner.

This is a port to PSD of @frankieroberto's changes in https://github.com/UKGovernmentBEIS/beis-opss/pull/1390.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


slorek:
Closed in favour of https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/66

PR #49, opened 2019-10-31
Merge web/worker apps

slorek:
## Description
There's no difference in the codebase between worker and web applications, just the way the Docker images are built and the deploy script is slightly different. This removes the `psd-worker` and `db` directories from the root of the project.

Also fixes a bug in the review app deployment script whereby it was looking for a `cosmetics` database.

Also fixes looking for antivirus/keycloak services tagged with `prod` branch name which will become out of date since we are no longer using that branching strategy.

Also removed the redundant `docker-compose` files and cleaned up the deploy scripts and Docker-only env files.

Also ensures the worker application is deployed in review apps.

Also fixed a failing test which should never have passed.

I have decided not to move the contents of `psd-web` into the root of the project in this PR because this could cause disruption to other in-flight work, but this is definitely the next intended step.

This does not change any functionality so does not require regression testing. We just need to check the deployment works after merging.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


infrastructure/ci/build-docker-images.sh
```diff
@@ -3,33 +3,17 @@
 # tagged with the build ID to DockerHub.
 set -ex
 
+echo "Logging into DockerHub"
+docker login -u $DOCKER_USERNAME -p $DOCKER_PASSWORD
```
StupidCodeFactory:
```suggestion
echo $DOCKER_PASSWORD | docker login -u $DOCKER_USERNAME --password-stdin
```
removes an anoying warning

slorek:
Workers are not currently deployed in review apps. This was going to be the next thing to fix, but will add it now to this PR.

PR #50, opened 2019-11-01
Resend invitiations

samlionheart:
https://trello.com/c/sQhCv7Tz/28-5-resending-account-invites

Tests weren't working locally so hoping for Travis to run them correctly.

https://psd-pr-50-web.london.cloudapps.digital/


psd-web/app/controllers/teams_controller.rb
```diff
@@ -57,12 +68,17 @@ def invite_existing_user_if_able(user)
...
+    if (@team.users.include? user)
+      if user.name.present?
+        @new_user.errors.add(:email_address,
+                            "#{@new_user.email_address} is already a member of #{@team.display_name}")
       return
+      else
+        resend_invitation_to_user(user.email)
+      end
+    else
+      invite_user user
```
StupidCodeFactory: I would **very much** love to see moved out of the controller, this is far too much logic in the controller. The controller is only here to dispatch the request to the relevant business logic and handle the response based on the outcome of this action.
Maybe something some service/domain/command class under `app/services/user/invite.rb` (with its own set of tests)

psd-web/config/routes.rb
```diff
@@ -118,6 +118,7 @@
     member do
       get :invite_to, path: "invite"
       put :invite_to, path: "invite"
+      put :resend_invitation
```
StupidCodeFactory: This should probably be its own `Teams::ResendInvitationsController` with a single `#create` action. 
Teams controller is not be concerned with invitations as they are two distinct concerns.

nasirkhanbeis:
@samlionheart  resend invitation link is throwing 500 error atm in above review app.

PR #52, opened 2019-11-04
Remove ActiveHash

slorek:
## Description
This change is a prerequisite to using asynchronous polling of user groups (teams).

NOTE: ActiveHash has a limitation whereby it cannot be a child of an ActiveRecord object (referenced via `has_many`). Therefore there will be some test failures e.g. `NoMethodError: undefined method `relation_delegate_class' for User:Class` until _all_ of the models have been converted.

## To Do
- [x] Convert all models to ActiveRecord
- [ ] Sidekiq jobs to sync users
- [ ] Ensure database is populated when loading application
- [x] Merge `user_attributes` table with `users`

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


slorek:
Closing as Travis has stopped doing checks. Re-opening separately.

PR #55, opened 2019-11-06
Resend account invite govuk visually hidden

samlionheart:
For accessibility, I was previously applying aria-label to the link_to

Now it has a span in the link_to text, so that screen readers read out "Resend invitation to #{user.email}"

Before it could miss the aria-label, as well as reading out "Resend invitation, Resend invitation to #{user.email}"


nasirkhanbeis:
can't really read the code but seems @edwardhorsford is happy so I'm happy for to merge.

PR #23, opened 2019-10-16
WIP: Update keycloak gem, add token class

samlionheart:
Ticket: https://trello.com/c/dpJZjxOi

WIP -  Needs Travis testing, as well as tests on the Token class.

Updated keycloak gem to 3.0.0

Add token class, that replaces the need for default_call in keycloak_client.rb

Token class checks if a token exists and if it has expired, generates a new token accordingly.

No longer logs out after each request to keycloak.

PR #56, opened 2019-11-06
Refactor teams/users to use asynchronous loading and ActiveRecord

slorek:
## Description
This changes the relationship between the application and Keycloak in order to improve performance for users.

Currently, the application implements `Team`, `Organisation`, `User` and `TeamUser` models using [ActiveHash](https://github.com/zilkey/active_hash), an in-memory emulation of ActiveRecord, to encapsulate data retrieved from Keycloak.

Unfortunately, the application has to keep in sync with Keycloak and so it has to pull all the data about all organisations, teams, users and team-user relationships *during the request cycle*, which we can see from Scout logs takes 30 seconds (and used to be a lot longer before we implemented [token re-use](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/23)). And in order to keep it up to date it has to do this [every 5 minutes](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/clients/keycloak_client.rb#L202).

This change converts those models to `ActiveRecord`, and moves that syncing process to an asynchronous job, so the user doesn't see it and are not impacted by the time it takes. The job runs every 5 minutes also, so the data is as fresh as the old method.

Currently, there's no way to delete a user/team within the app so it was not necessary to check for data which no longer exists in Keycloak. Deleting such data could result in errors with old cases.

This also fixes a few bugs:

* User was able to log in even if they had no teams (or PSD-related teams)
* Newly invited team member not appearing in team list immediately (due to cache implementation)
* A bug related to a user or team being deleted in Keycloak, but the application still depending on that data (e.g. case history)
* Keycloak requests for all groups/users were not re-using the same token
* In some cases Keycloak requests would fail but the reason was not obvious

There's quite a lot of gaps in testing of this functionality which have been noted, but the purpose of this change is to improve performance and has already taken some time, so I have not had the opportunity to improve the test coverage in this PR.

## To Do
- [x] Convert all models to ActiveRecord
- [x] Sidekiq jobs to sync users
- [x] Merge `user_attributes` table with `users`

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


psd-web/app/clients/keycloak_client.rb
```diff
@@ -2,9 +2,9 @@ module Keycloak
   module Client
     def self.get_installation
       @realm = "opss"
-      @auth_server_url = ENV["KEYCLOAK_AUTH_URL"]
-      @client_id = ENV["KEYCLOAK_CLIENT_ID"]
-      @secret = ENV["KEYCLOAK_CLIENT_SECRET"]
+      @auth_server_url = ENV.fetch("KEYCLOAK_AUTH_URL")
+      @client_id = ENV.fetch("KEYCLOAK_CLIENT_ID")
+      @secret = ENV.fetch("KEYCLOAK_CLIENT_SECRET")
```
StupidCodeFactory: Can we add those in `config/application.rb` under `config.x.keycloak.auth_url`, `config.x.keycloak.client_id` and `config.x.keycloak.client_secret`?

psd-web/app/controllers/concerns/authentication_concern.rb
```diff
@@ -24,8 +24,24 @@ def set_current_user
     return unless user_signed_in?
 
     user_info = KeycloakClient.instance.user_info(access_token)
-    User.current = ::User.find_or_create(user_info)
-    User.current.access_token = access_token
+
+    begin
+      User.current = User.find_or_create_by(id: user_info[:id]) do |user|
+        teams = Team.where(id: user_info[:groups])
+        organisation = Organisation.find_by(id: user_info[:groups]) || teams.first&.organisation
```
StupidCodeFactory: I know for sure that this is not going to work `Organisation.find_by(id: user_info[:groups])`
groups is an array containing the `path`

psd-web/app/models/organisation.rb
```diff
@@ -1,32 +1,13 @@
...
+  validates :id, presence: true, uuid: true
 
-    if options.has_key?(:conditions)
-      where(options[:conditions])
-    else
-      @records ||= []
+  def self.load_from_keycloak(orgs = KeycloakClient.instance.all_organisations)
+    orgs.each do |org|
+      record = find_or_create_by(id: org[:id])
+      record.update(org.slice(:name, :path))
```
StupidCodeFactory:
```suggestion
      record.update!(org.slice(:name, :path))
```

psd-web/app/models/organisation.rb
```diff
@@ -1,32 +1,13 @@
...
-    self.load
+  validates :id, presence: true, uuid: true
 
-    if options.has_key?(:conditions)
-      where(options[:conditions])
-    else
-      @records ||= []
+  def self.load_from_keycloak(orgs = KeycloakClient.instance.all_organisations)
+    orgs.each do |org|
+      record = find_or_create_by(id: org[:id])
```
StupidCodeFactory:
```suggestion
      record = create_or_find_by!(id: org[:id])
```
no biggy but find_or_create_by is not atomic, create_or_find_by is

psd-web/app/models/team.rb
```diff
@@ -1,53 +1,26 @@
...
-  def self.all(options = {})
-    self.load
-
-    if options.has_key?(:conditions)
-      where(options[:conditions])
-    else
-      @records ||= []
+  def self.load_from_keycloak(teams = KeycloakClient.instance.all_teams(Organisation.ids))
+    teams.each do |team|
+      record = find_or_create_by(id: team[:id])
```
StupidCodeFactory:
```suggestion
      record = create_or_find_by!(id: team[:id])
```

psd-web/app/models/team.rb
```diff
@@ -1,53 +1,26 @@
...
+      record.update(team.slice(:name, :path, :team_recipient_email, :organisation_id))
```
StupidCodeFactory:
```suggestion
      record.update!(team.slice(:name, :path, :team_recipient_email, :organisation_id))
```

psd-web/app/models/user.rb
```diff
@@ -1,87 +1,50 @@
...
 
-    if options.has_key?(:conditions)
-      where(options[:conditions])
-    else
-      @records ||= []
+      user
+    end
+
+    users.reject { |user| user[:organisation].blank? }.each do |user|
+      record = find_or_create_by(id: user[:id])
```
StupidCodeFactory:
```suggestion
      record = create_or_by!(id: user[:id])
```

psd-web/app/models/user.rb
```diff
@@ -1,87 +1,50 @@
...
+      record.update(user.slice(:name, :email, :organisation))
```
StupidCodeFactory:
```suggestion
      record.update!(user.slice(:name, :email, :organisation))
```

psd-web/app/models/user.rb
```diff
@@ -1,87 +1,50 @@
...
+      record.teams = user[:teams]
```
StupidCodeFactory: Do we need to save this?

psd-web/app/models/user.rb
```diff
@@ -103,23 +66,13 @@ def roles
...
-    teams = Team.where(id: groups)
-    organisation = Organisation.find_by(id: groups) || Organisation.find_by(id: teams.first&.organisation_id)
-    attributes.merge(organisation_id: organisation&.id)
-  end
-
-  def self.populate_name(attributes)
-    attributes[:name] ||= ""
-    attributes
+  def name
+    super.to_s
```
StupidCodeFactory: this looks suspicious? what was the problem here? Can name contain something else that must behave like a string?

psd-web/app/validators/uuid_validator.rb
```diff
@@ -0,0 +1,7 @@
+class UuidValidator < ActiveModel::EachValidator
```
StupidCodeFactory:
```suggestion
class UUIDValidator < ActiveModel::EachValidator
```
and in `config/initializers/inflections.rb`add:
```ruby
inflect.acronym 'UUID'
```

psd-web/db/migrate/20191104144011_create_organisations.rb
```diff
@@ -0,0 +1,11 @@
+class CreateOrganisations < ActiveRecord::Migration[5.2]
+  def change
+    create_table :organisations, id: "uuid", default: nil do |t|
+      t.string :name
+      t.string :path
+      t.timestamps
+
+      t.index :path
```
StupidCodeFactory: Can we enforce `null: false` on any of those if not all?

psd-web/db/migrate/20191104163921_create_users.rb
```diff
@@ -0,0 +1,52 @@
+class CreateUsers < ActiveRecord::Migration[5.2]
+  class User < ApplicationRecord; end
+  class UserAttributes < ApplicationRecord; end
+
+  def up
+    create_table :users, id: "uuid", default: nil do |t|
+      t.string :name
+      t.string :email
```
StupidCodeFactory: same here it would be great to be able to enforce `null: false` at some point in a later migration.

psd-web/app/models/user.rb
slorek: There's a bunch of other code which assumes it is an empty string (which it is in the existing implementation) unfortunately. It can be `nil` in the case of a user invited but not yet registered.

psd-web/db/migrate/20191104163921_create_users.rb
slorek: Yeah I was just discussing this with Ed. It might be a good idea but it might cause some issues with existing data. I think we should improve validations separately so we can control the impact more easily.

psd-web/db/migrate/20191104144011_create_organisations.rb
slorek: As below

psd-web/app/models/user.rb
slorek: I think it's auto-saved.

psd-web/app/controllers/concerns/authentication_concern.rb
slorek: It's an array of IDs according to my testing.

nasirkhanbeis:
@slorek Any idea why its returning error when I try open pre-existing cases?

psd-web/app/validators/uuid_validator.rb
slorek: Nice!

psd-web/app/models/user.rb
slorek: I don't want to change the interface too much in this PR because of the poor test coverage and data; all tests were passing before I added this `name` method but manual testing revealed an issue :/

psd-web/app/models/organisation.rb
slorek: I'll implement this but also add some exception handling as it could potentially break the app if there is an exception thrown during the loop.

psd-web/app/models/organisation.rb
slorek: Thanks, good tip.

psd-web/app/models/team.rb
slorek: Thanks, good tip.

psd-web/app/models/user.rb
slorek: 👍

psd-web/app/models/team.rb
slorek: 👍

psd-web/app/models/user.rb
slorek: 👍

nasirkhanbeis:
@slorek  As mentioned earlier in the standup- looks like caching issue although account is setup with name and password but it still shows the link "Resend invitation" however after a sometime on page refresh it is all good. this is something still asynchronous I think, other than this all good.

slorek:
These issues are expected behaviour. The presence of the `name` attribute is currently used to determine whether the user is enabled, which is clunky. Name is set in a Keycloak account page when signing up, so it will take up to 5 minutes to sync with the app, which is the same as the current implementation.

psd-web/app/controllers/concerns/authentication_concern.rb
slorek: I've done some more testing and you are correct, thanks.

psd-web/app/models/organisation.rb
slorek: `create_or_find_by!` was added in Rails 6 but we are running 5.2.3 still. However I will add validations and use `find_or_create_by!` as it will prevent defective data being persisted in our app.

psd-web/app/models/team.rb
slorek: `create_or_find_by!` was added in Rails 6 but we are running 5.2.3 still. However I will add validations and use `find_or_create_by!` as it will prevent defective data being persisted in our app.

psd-web/db/migrate/20191104163921_create_users.rb
slorek: I decided to add very basic validations to this PR to prevent defective data from Keycloak entering the application database.

psd-web/db/migrate/20191104144011_create_organisations.rb
slorek: I decided to add very basic validations to this PR to prevent defective data from Keycloak entering the application database.

psd-web/app/models/user.rb
slorek: Confirmed.

psd-web/app/models/user.rb
slorek: This should be improved when we introduce the additional user flag to determine if they're signed up.

PR #61, opened 2019-11-08
Compare whitelist domains insensitively and strip whitespace

slorek:
…telisted domains

## Description
https://trello.com/c/xHjhwZgT/187-email-whitelist-not-checked-case-insensitively

This ensures email domains are compared case insensitively and adds beis.gov.uk. Also some of the whitelisted domains had trailing whitespace which would have not allowed users from those domains to be invited.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


psd-web/test/controllers/teams_controller_test.rb
```diff
@@ -1,8 +1,8 @@
-ENV["EMAIL_WHITELIST_ENABLED"] = "true"
 require "test_helper"
 
 class TeamsControllerTest < ActionDispatch::IntegrationTest
   setup do
+    ENV["EMAIL_WHITELIST_ENABLED"] = "true"
```
StupidCodeFactory: could have something a bit more sensible in `config/application.rb`, all the config from ENV var would be listed in the same place and the test won't have to modify the ENV, which I always find a little intrusive.
```ruby
config.email_whitelist_enabled = ENV.fetch("EMAIL_WHITELIST_ENABLED", "true") == "true"
```
and in the test:
```ruby
allow(Rails.application.config).to receive(:email_whitelist_enabled).and_return(true)
```

slorek: Good idea

nasirkhanbeis:
Looks good to me!

PR #63, opened 2019-11-11
Add home page

doctorpod:
Ticket: https://trello.com/c/nCIzIIXe/

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


doctorpod:
Review app: https://psd-pr-63-web.london.cloudapps.digital/

nasirkhanbeis:
@doctorpod looking good to me

edwardhorsford:
Spotted an issue with this.

The main page content is not inside of a `main` element. There is a main element, but it's lower down the page and empty. The content from the h1 to the the footer (but not including the footer) should be within a `main` element.

edwardhorsford:
@doctorpod can you also update the trello story acceptance criteria if any of them have been done?

psd-web/app/controllers/homepage_controller.rb
```diff
@@ -1,6 +1,22 @@
 class HomepageController < ApplicationController
+  skip_before_action :authenticate_user!, only: :landing_page
+  skip_before_action :authorize_user, only: :landing_page
+  skip_before_action :has_accepted_declaration
```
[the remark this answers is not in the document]
doctorpod: We need this line as it skips the original implementation. If we remove it, it will run before each action, even the home_page action when user not logged in.

edwardhorsford:
All looks good to me. The only thing I query is that the mailto email address uses uppercase `OPSS` whereas I think it should be lowercase. @LMallows can confirm.

LMallows:
> All looks good to me. The only thing I query is that the mailto email address uses uppercase `OPSS` whereas I think it should be lowercase. @LMallows can confirm.

correct, the email should be opss.enquiries@beis.gov.uk

StupidCodeFactory:
> All looks good to me. The only thing I query is that the mailto email address uses uppercase `OPSS` whereas I think it should be lowercase. @LMallows can confirm.

I've implemented the change

psd-web/spec/decorators/investigation_decorator_spec.rb
```diff
@@ -4,7 +4,7 @@
   include ActionView::Helpers::DateHelper
   include ActionView::Helpers::TextHelper
 
-  fixtures(:investigations, :investigation_products, :products)
+  fixtures(:investigations, :investigation_products, :products, :sources)
```
slorek: I don' think this change is necessary. I already fixed the failing test separately.

psd-web/test/system/homepage_test.rb
```diff
@@ -0,0 +1,27 @@
+require "application_system_test_case"
```
slorek: We should not be adding new system tests. Prefer RSpec feature specs.

psd-web/spec/features/home_page_spec.rb
```diff
@@ -0,0 +1,30 @@
+require "rails_helper"
+
+RSpec.describe "HomePage", :with_keycloak_config, :with_elasticsearch do
```
StupidCodeFactory: actually I'll change this to a request spec

slorek: Do we really need Elasticsearch for this?

psd-web/spec/requests/home_page_spec.rb
```diff
@@ -0,0 +1,30 @@
+require "rails_helper"
+
+RSpec.describe "HomePage", :with_keycloak_config, :with_elasticsearch do
```
slorek: Why do we need `:with_elasticsearch` if we don't care what cases are rendered on the page? It might slow the test down or make it more brittle.

psd-web/app/views/layouts/application.html.slim
```diff
@@ -47,22 +49,51 @@ html.govuk-template.app-html-class[lang="en"]
...
+              .govuk-grid-column-two-thirds-from-desktop
+                h1.govuk-heading-xl Product safety database
+                p.govuk-body-l class="govuk-!-margin-bottom-2"
+                  | Report, track and share product safety information with the product safety community.
+                .landing-masthead-links
+                  = link_to "Sign in to your account", keycloak_login_url(params[:request_path]),
+                      class: "govuk-button", role: "button", draggable: false
+                  p
+                    | or&nbsp;
+                    = mail_to "opss.enquiries@beis.gov.uk", "email OPSS"
```
slorek: This e-mail address is used several times in the app, sometimes shown as OPSS.enquiries@beis.gov.uk. Perhaps we should have a configuration variable (not necessarily as part of this change).

edwardhorsford: Yes, we use this all over the place and a variable (with correct lowercase casing) would be good.

PR #64, opened 2019-11-11
Set Sentry environment name

frankieroberto:
This sets a `SENTRY_CURRENT_ENV` environment variable when deploying review apps, so that we know which review app an exception occurred in (especially within the worker app as that doesn’t send a hostname).

Port from cosmetics project of:
* https://github.com/UKGovernmentBEIS/beis-opss/pull/1456
* https://github.com/UKGovernmentBEIS/beis-opss/pull/1435


slorek:
Do we need to alter the user provided service as well before deployment?

frankieroberto:
@slorek yep. And annoyingly that's a manual job as I don't think those services are specified in code anywhere.

slorek:
I've updated the `psd-sentry-env` user-provided service.

PR #69, opened 2019-11-15
Format free text field

StupidCodeFactory:
## Description

1. Add `simple_format` in the relevant decorators.
2. move away from fixtures in favour of FactoryBot in investigation_decorator_spec.rb

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/app/decorators/product_decorator.rb
+++ b/psd-web/app/decorators/product_decorator.rb
@@ -4,12 +4,12 @@ class ProductDecorator < ApplicationDecorator
 
   def summary_list(include_batch_number: false)
     rows = [
-      { key: { text: "Product name" }, value: { text: product.name } },
+      { key: { text: "Product name" }, value: { text: object.name } },
       { key: { text: "Barcode / serial number" }, value: { text: product_code } },
       { key: { text: "Type" }, value: { text: product_type } },
       include_batch_number ? { key: { text: "Batch number" }, value: { text: batch_number } } : nil,
       { key: { text: "Category" }, value: { text: category } },
-      { key: { text: "Webpage" }, value: { text: product.webpage } },
+      { key: { text: "Webpage" }, value: { text: object.webpage } },
       { key: { text: "Country of origin" }, value: { text: country_from_code(country_of_origin) } },
       { key: { text: "Description" }, value: { text: description } }
     ]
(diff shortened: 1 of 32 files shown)


psd-web/app/decorators/product_decorator.rb
```diff
@@ -1,18 +1,23 @@
 class ProductDecorator < ApplicationDecorator
   delegate_all
   decorates_association :investigations
+
   def summary_list(include_batch_number: false)
     rows = [
       { key: { text: "Product name" }, value: { text: product.name } },
```
slorek: Just for my understanding: is `product` an alias for `object` inferred from the name of the class? I couldn't see documentation of this usage.

psd-web/spec/decorators/investigation_decorator_spec.rb
```diff
@@ -1,46 +1,50 @@
...
-    let(:investigation) { investigations(:enquiry) }
+    let(:investigation) { create(:enquiry) }
+
     context "with no product" do
       it { is_expected.to_not be_display_product_summary_list }
     end
 
     context "with products" do
-      before { investigation.products << products(:one) }
+      before { investigation.products << create(:product) }
```
slorek: Minor improvement might be to use `let(:products) { [create(:product)] }` here and modify the `let(:investigation)` block above.

psd-web/spec/factories/products.rb
```diff
@@ -0,0 +1,35 @@
+FactoryBot.define do
+  factory :product do
+    name { "product name" }
+    description { "product description" }
+    category { Rails.application.config.product_constants["product_category"].sample }
+    product_type { "product_type" }
+  end
+
+  factory :iphone, parent: :product do
```
slorek: I wonder if we should prepend the name of these with `product_` so it's clearer what they are.

psd-web/app/decorators/product_decorator.rb
StupidCodeFactory: yep - I tend to use `object` over the inferred name from the class as it's more portable. I think I moved this code over from the investigation decorator where I would have had to anyway reference it with `product`.
I will change this to object for consistency

psd-web/spec/decorators/investigation_decorator_spec.rb
StupidCodeFactory: so I tried to use `let(:products)` but at that point the investigation had already been created. I didn't want to do an an `object.reload` in the decorator. Adding it this way the investigation object is aware of its new products.

PR #70, opened 2019-11-18
Fix page layout on error pages not being rendered

slorek:
## Description
This fixes the page layout not being rendered on error pages.

Before:

Screenshot 2019-11-18 at 14 53 34

After (logged out):

Screenshot 2019-11-18 at 14 51 37

After (logged in)

Screenshot 2019-11-18 at 14 51 24

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/app/controllers/application_controller.rb
+++ b/psd-web/app/controllers/application_controller.rb
@@ -27,7 +27,7 @@ class ApplicationController < ActionController::Base
   end
 
   def nav_items
-    return nil if hide_nav? # On some pages we don't want to show the main navigation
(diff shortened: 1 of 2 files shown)


edwardhorsford:
Thanks Steve 👍

nasirkhanbeis:
looking good

edwardhorsford:
I had a look on the review app - for the two error pages I can easily trigger (page not found and not authorised), both look good to me.

StupidCodeFactory:
:pensive: I don't quite understand why we have a controller for the erros instead of simple html pages.

PR #71, opened 2019-11-19
Increase site width to 1200px

slorek:
## Description
This increases the site width to 1200px and sets max-width of content items to 722px.

I think there may be some further tweaks required but will let @edwardhorsford check and specify these.

Before:

Screenshot 2019-11-19 at 14 45 25

After:

Screenshot 2019-11-19 at 14 47 21

## Checklist:

### General testing
- [ ] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [ ] Works keyboard only
- [x] Passes automated checker (automated in build or manual)


edwardhorsford:
I see two possible bugs - please don't merge before I have a chance to dig in a bit.

edwardhorsford:
Have updated the list to include more _content-like_ elements from the design system.

Also moved the teams page away from using a `govuk-list`.

edwardhorsford:
Have updated the product and business templates to ensure tab content is always in a grid column - so these are not so long now.

edwardhorsford:
@LMallows can you check this?

PR #73, opened 2019-11-19
Hide invited users

slorek:
## Description
This hides users who have not yet registered (but may have received an invitation email) from the case assignment and filter lists, and also prevents them receiving alert emails.

https://trello.com/c/sHYQIJom/197-assign-case-someone-in-your-team-someone-else-drop-down-to-display-only-team-member-names-rather-team-names-too

https://trello.com/c/KUR6Njtu/201-only-send-notifications-alerts-whatever-to-users-who-have-finished-registration-not-invited-users

I've introduced FactoryBot to this codebase and also added some helpers to find content on a page in request specs via a CSS selector.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


nasirkhanbeis:
Hi @slorek I had a quick look and seems it is not listing all the users from the team 
image

edwardhorsford:
I don't know if it's related (might be a separate bug), but I'm seeing OPS&S listed several times in the 'other person or team' autocomplete in the case filter page - there's two issues here
1. we should only be showing teams, and OPS&S is an org
2. there's three of them (or more?)

Screenshot 2019-11-20 at 10 14 29

slorek:
> Hi @slorek I had a quick look and seems it is not listing all the users from the team
> image

@nasirkhanbeis this is expected behaviour due to the database migration. It has created a new database, separate to the other review apps. In this database there's currently only four activated users. You need to sign in and accept the declaration to become activated. You will then appear in the lists.

slorek:
@edwardhorsford I've tracked this down, it is intended (and existing) behaviour, but definitely not right. This happens when it tries to show a team name belonging to another organisation. In this case it shows only the organisation name. I think we need to decide what the correct behaviour should be in this scenario.

nasirkhanbeis:
@slorek I thought we agreed not to consider declaration acceptance flag as declaration may change anytime which then user won't be able see/assign to others user until other user as logged-in and accepted.  I think we need to consider as active user once they complete setup and logged at least for once . This may include historical login  irrespective they accepted new declaration or not @edwardhorsford  pls confirm if you agree

slorek:
> @slorek I thought we agreed not to consider declaration acceptance flag as declaration may change anytime which then user won't be able see/assign to others user until other user as logged-in and accepted. I think we need to consider as active user once they complete setup and logged at least for once . This may include historical login irrespective they accepted new declaration or not @edwardhorsford pls confirm if you agree

We agreed that the *first* time they accept the declaration they will be marked as activated. It is *not* tied to the current state of declaration acceptance.

edwardhorsford:
> @edwardhorsford I've tracked this down, it is intended (and existing) behaviour, but definitely not right. This happens when it tries to show a team name belonging to another organisation. In this case it shows only the organisation name for several OPSS teams because the user is not in the OPSS organisation. I think we need to decide what the correct behaviour should be in this scenario.

This is a bug and not intended behaviour - but unrelated to this ticket. We've got another bug open about this happening - so we can add to that. I know how this came about - there was a misunderstanding about what a story was asking for and it seems to have manifested as this feature where we don't show team names to non-OPS&S people.

For clarity, the intended functionality is that we show a list of users and teams.

edwardhorsford:
@slorek I've added the above to a trello bug ticket. It seems unrelated to this work, so this seems good to me.

psd-web/spec/factories/organisations.rb
```diff
@@ -0,0 +1,7 @@
+FactoryBot.define do
+  factory :organisation do
+    id { SecureRandom.uuid }
```
StupidCodeFactory: you can remove this. This will be autogenerated by the DB

psd-web/spec/factories/teams.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :team do
+    id { SecureRandom.uuid }
```
StupidCodeFactory: you can remove this. This will be autogenerated by the DB

psd-web/spec/factories/teams.rb
```diff
@@ -0,0 +1,9 @@
...
+    name { "test" }
+    team_recipient_email { "test@example.com" }
+    organisation
```
StupidCodeFactory: So this is how one slowly makes its test suite **very** slow. Now every time you create a `team` you'll be creating an `organisation`. Make sure you wrap associations in a block. 
```ruby
organisation { create :organistation } 
```

psd-web/spec/factories/users.rb
```diff
@@ -0,0 +1,37 @@
...
+    has_accepted_declaration { false }
+    has_been_sent_welcome_email { true }
+    has_viewed_introduction { false }
+    account_activated { false }
+
+    transient do
+      roles { [:psd_user] }
+    end
+
+    trait :activated do
```
StupidCodeFactory: So you see here if we had Service `User::Accept`,  you could just use this in the factory:
```ruby
factory :user do
  # attributes
  after(:create) do |user|
    # mock email service or any external call that maybe made
    User::Accept.call(user) 
   end
end
```
With this approach, we have to maintain the logic of an activated user in two places, this doesn't scale.
When we update the logic this will get missed and we will have weird, out of date, nowhere close to the real world factories.

psd-web/spec/models/alert_spec.rb
```diff
@@ -0,0 +1,17 @@
+require "rails_helper"
+
+RSpec.describe Alert do
+  describe "#send_alert_email" do
+    subject(:alert) { described_class.new(summary: "test", description: "test") }
+
+    before do
+      @unactivated_user = create(:user, email: "unactivated@example.com")
+      @activated_user = create(:user, :activated, email: "activated@example.com")
```
StupidCodeFactory: What's wrong with `let` declaration. Let's are cached between each test instances, will raise NameError if misspelled (not the case with instance variable).

```ruby
let(:unactivated_user) { create(:user, email: "unactivated@example.com") } 
let(:activated_user) { create(:user, :activated, email: "activated@example.com") }
```

psd-web/spec/models/user_spec.rb
```diff
@@ -1,21 +1,87 @@
...
-  before do
-    allow(ENV).to receive(:fetch).with("KEYCLOAK_AUTH_URL").and_return("test")
-    allow(ENV).to receive(:fetch).with("KEYCLOAK_CLIENT_ID").and_return(client_id)
-    allow(ENV).to receive(:fetch).with("KEYCLOAK_CLIENT_SECRET").and_return(client_secret)
-    allow(KeycloakToken).to receive(:new).and_return(token_stub)
+RSpec.describe User do
+  describe ".activated" do
+    before do
+      @unactivated_user = create(:user)
+      @activated_user = create(:user, :activated)
```
StupidCodeFactory: same here. use `let`

psd-web/spec/models/user_spec.rb
```diff
@@ -1,21 +1,87 @@
...
+    it "returns only users with activated accounts" do
+      expect(User.activated.to_a).to eq [@activated_user]
+    end
+  end
+
+  describe "#activate!" do
+    subject(:user) { build(:user) }
+
+    it "sets the activated flag on the user" do
+      expect(user.account_activated?).to be_falsy
```
StupidCodeFactory: why check for falsey instead of false? also there is nicer way to right these tests.
```suggestion
      expect(user).not_to be_account_activated
```

psd-web/spec/models/user_spec.rb
```diff
@@ -1,21 +1,87 @@
...
+      user.activate!
+      expect(user.account_activated?).to be_truthy
```
StupidCodeFactory:
```suggestion
      expect(user).to be_account_activated
```

psd-web/spec/models/user_spec.rb
```diff
@@ -1,21 +1,87 @@
...
+  describe ".get_team_members" do
+    let(:team) { create(:team) }
+    let(:user) { create(:user, :activated, teams: [team]) }
+    let(:investigation) { create(:allegation) }
+    let(:team_members) { described_class.get_team_members(user: user) }
+
+    before do
+      @another_active_user = create(:user, :activated, organisation: user.organisation, teams: [team])
+      @another_inactive_user = create(:user, organisation: user.organisation, teams: [team])
+      @another_user_with_another_team = create(:user, teams: [create(:team)])
```
StupidCodeFactory: `let!` those instance variable.

psd-web/spec/models/user_spec.rb
```diff
@@ -1,21 +1,87 @@
...
+
+    it "does not return other users on other teams" do
+      expect(team_members).not_to include(@another_user_with_another_team)
+    end
+  end
+
+  describe ".get_assignees" do
+    before do
+      @active_user = create(:user, :activated)
+      @inactive_user = create(:user)
```
StupidCodeFactory: Both `let!`

psd-web/spec/models/user_spec.rb
```diff
@@ -1,21 +1,87 @@
...
+    end
   end
 
-  let(:client_id) { "123" }
-  let(:client_secret) { "secret" }
-  let(:token_stub) { OpenStruct.new(access_token: "test") }
-  let(:id) { SecureRandom.uuid }
+  describe "#roles", with_keycloak_config: true do
+    let(:id) { SecureRandom.uuid }
+    subject(:user) { described_class.new(id: id) }
```
StupidCodeFactory: now that we have `FactoryBot` we can just use `build(:user)` or `build_stubbed(:user)`

psd-web/spec/requests/assign_investigation_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+describe "Assigning an investigation", type: :request, with_keycloak_config: true do
+  let(:team) { create(:team) }
+  let(:user) { create(:user, :activated, teams: [team]) }
+  let(:investigation) { create(:allegation) }
+
+  before do
+    @another_active_user = create(:user, :activated, organisation: user.organisation, teams: [team])
+    @another_inactive_user = create(:user, organisation: user.organisation, teams: [team])
+    @another_active_user_another_team = create(:user, :activated, organisation: user.organisation, teams: [create(:team)])
+    @another_inactive_user_another_team = create(:user, organisation: user.organisation, teams: [create(:team)])
```
StupidCodeFactory: `let!` all of these

psd-web/spec/requests/filter_investigations_spec.rb
```diff
@@ -0,0 +1,51 @@
+require "rails_helper"
+
+class DummyElasticSearchResult
+  attr_accessor :_id
+  def merge(_); end
+end
+
+describe "Filtering the investigations list", type: :request, with_keycloak_config: true do
```
StupidCodeFactory:
```suggestion
RSpec.describe "Filtering the investigations list", type: :request, with_keycloak_config: true do
```

psd-web/spec/requests/filter_investigations_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  let(:team) { create(:team) }
+  let(:user) { create(:user, :activated, teams: [team]) }
+  let(:investigation) { create(:allegation) }
+
+  before do
+    stub_elasticsearch_results
+    @another_active_user = create(:user, :activated, organisation: user.organisation, teams: [team])
+    @another_inactive_user = create(:user, organisation: user.organisation, teams: [team])
```
StupidCodeFactory: `let!` those

psd-web/app/models/user.rb
```diff
@@ -122,13 +130,13 @@ def is_team_admin?
 
   def self.get_assignees(except: [])
     users_to_exclude = Array(except)
-    self.all - users_to_exclude
+    self.activated - users_to_exclude
   end
 
   def self.get_team_members(user:)
```
slorek: I think this method can be potentially refactored to a standard ActiveRecord relationship, but I'll look at it separately.

psd-web/spec/requests/filter_investigations_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    sign_in(as_user: user)
+  end
+
+  def stub_elasticsearch_results
+    result = DummyElasticSearchResult.new
+    result._id = investigation.id
+
+    results = OpenStruct.new(results: [result])
+    allow_any_instance_of(InvestigationsController).to receive(:search_for_investigations).and_return(results)
```
StupidCodeFactory: There is _almost_ *never* a good reason to use `(allow|expect)_any_instance_of`.

```ruby
allow(Investigation).to receive(:full_search).with(anyargs).and_return(results)
```

psd-web/spec/requests/team_spec.rb
```diff
@@ -0,0 +1,56 @@
+require "rails_helper"
+
+describe "Your team", type: :request, with_keycloak_config: true do
+  let(:team) { create(:team) }
+  let(:user) { create(:user, :activated, teams: [team]) }
+
+  before do
+    @another_active_user = create(:user, :activated, email: "active.sameteam@example.com", organisation: user.organisation, teams: [team])
+    @another_inactive_user = create(:user, email: "inactive.sameteam@example.com", organisation: user.organisation, teams: [team])
+    @another_user_another_team = create(:user, :activated, email: "active.otherteam@example.com", organisation: user.organisation, teams: [create(:team)])
```
StupidCodeFactory: let!

psd-web/spec/requests/user_declaration_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    before do
+      expect(user.has_accepted_declaration).to be_falsy
+      expect(user.account_activated).to be_falsy
+      post declaration_accept_path, params: { agree: "checked" }
+      user.reload
+    end
+
+    context "when they accept the declaration" do
+      it "sets the user accepted declaration flag" do
+        expect(user.has_accepted_declaration).to be_truthy
```
StupidCodeFactory:
```suggestion
        expect(user).to be_has_accepted_declaration
```

psd-web/spec/requests/user_declaration_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      end
+
+      it "sets the user account activated flag" do
+        expect(user.account_activated).to be_truthy
```
StupidCodeFactory:
```suggestion
        expect(user).to be_account_activated
```

psd-web/spec/support/elasticsearch.rb
```diff
@@ -0,0 +1,6 @@
+RSpec.configure do |config|
+  config.before :each do
+    elasticsearch_url = ENV.fetch("ELASTICSEARCH_URL", "http://elasticsearch:9200")
+    stub_request(:any, /#{Regexp.quote(elasticsearch_url)}/).to_return(body: "Elasticsearch disabled in Rspec", status: 200)
```
StupidCodeFactory: Hm before(:each) might be a lot. I'd like webmock to tell me when I'm calling out to ES and then stub it out on a per-test basis.

psd-web/spec/support/response_matcher.rb
```diff
@@ -0,0 +1,31 @@
+require "nokogiri"
+
+module ResponseMatcher
+  class CssMatcher
```
StupidCodeFactory: :scream: why not use the normal have_css?  
just add `require 'capybara/rspec'` and all the have_css, have_link have_checked... goodies with be added to specs of `type: :request`

psd-web/app/models/user.rb
```diff
@@ -12,6 +12,14 @@ class User < ApplicationRecord
   attr_accessor :access_token # Used only in User.current thread context
   attr_writer :roles
 
+  def self.activated
```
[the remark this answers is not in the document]
slorek: Scopes are just syntactic sugar on class methods. Both work, I just prefer this way as it’s easier to find.

psd-web/spec/models/alert_spec.rb
slorek: Because `let` is lazily evaluated and we need to ensure these are setup before making the request. I think `let!` would work but in previous projects there was a Rubocop rule preventing its usage so I avoid it now.

psd-web/spec/models/user_spec.rb
slorek: Yes I forgot to check the existing tests.

psd-web/spec/requests/assign_investigation_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  context "when the user wishes to assign to another user in the same team" do
+    before { get "/cases/#{investigation.pretty_id}/assign/choose" }
+
+    it "shows other active users" do
+      expect(response.body).to have_css("#investigation_select_team_member option[value=\"#{@another_active_user.id}\"]")
+    end
+
+    it "does not show other inactive users" do
+      expect(response.body).not_to have_css("#investigation_select_team_member option[value=\"#{@another_inactive_user.id}\"]")
+    end
```
StupidCodeFactory: maybe collapse those two tests

psd-web/spec/requests/assign_investigation_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  context "when the user wishes to assign to someone else" do
+    before { get "/cases/#{investigation.pretty_id}/assign/choose" }
+
+    it "shows other active users" do
+      expect(response.body).to have_css("#investigation_select_someone_else option[value=\"#{@another_active_user_another_team.id}\"]")
+    end
+
+    it "does not show other inactive users" do
+      expect(response.body).not_to have_css("#investigation_select_someone_else option[value=\"#{@another_inactive_user_another_team.id}\"]")
+    end
```
StupidCodeFactory: same here those test can be collapsed into one

you can add the `:aggregate_failure` as metadata

nasirkhanbeis:
Hi @slorek  looking good as expected now, however just found a minor issue
all pre-existing users show up with resend invitation link except admin
image

nasirkhanbeis:
> Hi @slorek looking good as expected now, however just found a minor issue
> all pre-existing users show up with resend invitation link except admin
> image

Please confirm if this again new db issue(as none users accepted declaration) - In this case it is vital to check in prod as soon we release this fix.

slorek:
@nasirkhanbeis correct, because this is a new database version it has created a new database from the seeds in the review app. Users need to accept the declaration again to remain active. This will not happen when deployed to product as there is a database migration to update the existing users (tested locally), but we should double check when deploying.

psd-web/app/controllers/declaration_controller.rb
```diff
@@ -12,6 +12,7 @@ def accept
       return render :index
     end
     User.current.has_accepted_declaration!
+    User.current.activate!
     send_welcome_email unless User.current.has_been_sent_welcome_email
```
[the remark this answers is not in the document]
slorek: @StupidCodeFactory Agreed, I have refactored this as you suggested. I've also removed the redundant methods on the User object.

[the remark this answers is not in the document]
slorek: I also added test coverage around this which didn't exist.

psd-web/spec/factories/organisations.rb
slorek: ActiveRecord validates the ID before save:

`Validation failed: Id can't be blank, is not a valid UUID`

Since the UUIDs come from Keycloak I wasn't comfortable removing this validation as I don't want the DB to generate new UUIDs which don't match with keycloak's (which is the source of truth).

psd-web/spec/factories/teams.rb
slorek: See above.

psd-web/spec/factories/teams.rb
slorek: Isn't it the same - it will create an Organisation in both cases. I think the block syntax is if you want to pass in custom attributes or do something other than just create a standard instance of the named factory?

psd-web/spec/factories/users.rb
slorek: So Rspec mocks aren't available inside factories. We could add a flag to the service to control whether the email is sent, but I also wonder whether relying on the service implementation might give us false positives (in the case where the service wasn't working, for example). What do you think?

slorek: Also it will result in extra database queries which slow the tests down.

psd-web/spec/models/alert_spec.rb
slorek: I refactored this test because the `before` block was redundant anyway.

psd-web/spec/models/user_spec.rb
slorek: Removed the before block as it was redundant.

psd-web/spec/models/user_spec.rb
slorek: This was redundant anyway after I added a trait to the factory so we can be explicit in the state of the user at the time it is created.

psd-web/spec/models/user_spec.rb
slorek: Agree, we should not use instance variables in Rspec.

psd-web/spec/requests/filter_investigations_spec.rb
slorek: Oops.

psd-web/spec/support/elasticsearch.rb
slorek: Good point, I'll convert it to a shared context.

psd-web/spec/requests/filter_investigations_spec.rb
slorek: Good point, must have been tired.

psd-web/spec/support/response_matcher.rb
slorek: Capybara is removed from Rspec request tests now. There is an open question as to whether we should convert these to feature tests.

psd-web/spec/factories/teams.rb
StupidCodeFactory: I don't think it's the same, I could be wrong though, I think rubocop facotry_bot linter requires you to use the block version for this reason.

```ruby 
org = create :organisation
Organisation.count # => 1
create :team, organisation: org
Organisation.count # => 2
```
that's the way to verify

StupidCodeFactory: if you spin an irb and require 'factory_bot-rails' 
you should be able to execute the following and check whether I'm right or wrong

psd-web/spec/factories/users.rb
StupidCodeFactory: Hm ok, (maybe a load order problem), I know it's a bit tricky sometimes to get things working in factories.
Let's leave it like this for now, I'll investigate this when it merged and I'll fiddle around with the factories and see what I can do.

StupidCodeFactory: It could also be possible to inject a Test mailer into the service class so that in tests we use that. 
Unfortunately, Rails is pretty bad at encouraging dependency injection which is a life saver when building complex testable software.

slorek: If we could inject a test double that would work too, but not possible. When I tried to do this yesterday I found this: https://github.com/thoughtbot/factory_bot/issues/703

psd-web/spec/factories/teams.rb
slorek: Confirmed it does not create an extra record, count remains the same.

slorek: Checked Rubocop cops for FactoryBot here: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot. There is a copy for statically defined attributes but this is not applicable here.

psd-web/spec/support/response_matcher.rb
slorek: Removed and converted request specs to feature specs as agreed.

psd-web/spec/factories/users.rb
slorek: Good spot! That worked. Will refactor accordingly.

PR #82, opened 2019-12-02
Fix failing RSpec tests

slorek:
## Description
This fixes two intermittently failing RSpec tests.

They previously failed intermittently because the database was not clean and sometimes the wrong object was unwittingly being referenced due to data conflicts.

In fact, neither of these tests should have passed if run properly in isolation, so previous passes were false positives.

I've added the [Database Cleaner gem](https://github.com/DatabaseCleaner/database_cleaner) to ensure RSpec tests are cleaned up properly after each test. Both tests failed reliably after adding this.

I also changed the CI environment setup so the unrelated test data seeds are no longer loaded; this data is unrelated to the automated tests and is intended for manual testing/demo purposes.

Because this was causing failures, I also refactored the `User#display_name` method to better reflect current business logic and to remove the assumption that `User.current` is set. This is not true in the RSpec suite. In the `test` suite they have [manually set this](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/test/test_helper.rb#L78) in every test, but I don't think this is a good solution and exposes poor design. Instead, the 'current' user can be injected into the method as an argument. I also added instance variable caching to this method to hopefully improve performance; it is responsible for a lot of database queries currently according to Scout.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/spec/rails_helper.rb
+++ b/psd-web/spec/rails_helper.rb
@@ -41,7 +41,7 @@ RSpec.configure do |config|
   # If you're not using ActiveRecord, or you'd prefer not to run each of your
   # examples within a transaction, remove the following line or assign false
   # instead of true.
-  config.use_transactional_fixtures = true
+  # config.use_transactional_fixtures = true
 
   # RSpec Rails can automatically mix in different behaviours to your tests
   # based on their file location, for example enabling you to call `get` and
--- a/psd-web/spec/support/database_cleaner.rb
+++ b/psd-web/spec/support/database_cleaner.rb
@@ -1,0 +1,12 @@
+RSpec.configure do |config|
+  config.before(:suite) do
+    DatabaseCleaner.strategy = :transaction
+    DatabaseCleaner.clean_with(:truncation)
+  end
+
+  config.around(:each) do |example|
+    DatabaseCleaner.cleaning do
+      example.run
+    end
+  end
+end
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -95,7 +95,7 @@ jobs:
         RAILS_ENV: test
       run: |
         cd psd-web
-        bin/rails db:setup
+        bin/rails db:create db:schema:load
     - name: Run Tests
       env:
         ELASTICSEARCH_URL: localhost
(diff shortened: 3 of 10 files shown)


psd-web/spec/rails_helper.rb
```diff
@@ -41,7 +41,7 @@
   # If you're not using ActiveRecord, or you'd prefer not to run each of your
   # examples within a transaction, remove the following line or assign false
   # instead of true.
-  config.use_transactional_fixtures = true
+  # config.use_transactional_fixtures = true
```
StupidCodeFactory: we do want to user transactional fixtures

psd-web/spec/support/database_cleaner.rb
```diff
@@ -0,0 +1,12 @@
+RSpec.configure do |config|
+  config.before(:suite) do
```
StupidCodeFactory: this shouldn't be needed in rails 5.2

I'm wondering whether the problem doesn't come from the minitests fixtures

`https://api.rubyonrails.org/v5.2.3/classes/ActiveRecord/FixtureSet.html`
```ruby
class FooTest < ActiveSupport::TestCase
  self.use_transactional_fixtures = false
end
```

StupidCodeFactory:
This is a quick fix and I think we should fix those properly- rails' team made a lot of effort to make the use of transaction work accross the board whether in system test or simple 
unit tests

psd-web/spec/support/database_cleaner.rb
slorek: It's definitely coming from the RSpec suite. I didn't run the minitest suite between runs when testing the problem.

psd-web/spec/rails_helper.rb
slorek: It conflicts with database_cleaner - we get 19 failures with it set to `true`.

slorek:
I did a bit of reading into the RSpec implementation and there are some limitations in the way it cleans up data using transactional fixtures only. I'm not sure it's feasible but I'd be happy if we can make it work later.

PR #85, opened 2019-12-03
Fix flaky RSpec spec

slorek:
## Description
This addresses a flaky RSpec spec which could fail, depending on the order in which the two associated products were returned. The existing test only matched against the first returned product, but in some cases where the other product was returned first the matcher would fail because it was being compared case-sensitively. The [decorator modifies the case of the text](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/decorators/investigation_decorator.rb#L127) on the page from the raw database value.

This change simply modifies the test to compare case insensitively, and also adds some regex escaping to guard against potential failures in future where the data might contain special characters, which might not be spotted or provide false positives.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/spec/decorators/investigation_decorator_spec.rb
+++ b/psd-web/spec/decorators/investigation_decorator_spec.rb
@@ -28,7 +28,9 @@ RSpec.describe InvestigationDecorator, with_stubbed_elasticsearch: true do
 
     it "has the expected fields" do
       expect(product_summary_list).to summarise("Product details", text: "2 products added")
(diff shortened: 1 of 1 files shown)


slorek:
Minitest suite has failed repeatedly but this is unrelated to the change and will be addressed separately. GitHub actions has got into a state where the tests can't be rerun.

PR #86, opened 2019-12-03
Fix e-mail not being sent when creating a case

slorek:
## Description
This fixes a regression [introduced recently](https://github.com/UKGovernmentBEIS/beis-opss-psd/commit/c78dd635bff891bb8a4834221326b492d74ca9ee#diff-c573245a97e3475b3a065cd10502cd47L14) which prevented e-mails from being sent on case creation. It also adds a new test to prevent future regressions.

I have not attempted to improve the implementation which is pretty nasty with callbacks and dependency on `User.current`, the purpose of this is to fix the regression only and prevent it happening again.

It does also provide a starting point to build up test coverage of the `Investigation` model and subclasses.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/spec/models/investigation_spec.rb
+++ b/psd-web/spec/models/investigation_spec.rb
@@ -1,0 +1,38 @@
+require "rails_helper"
+
+# TODO: Refactor Investigation model to remove callback hell and dependency on User.current
+RSpec.shared_examples "an Investigation" do
+  describe "record creation", with_stubbed_elasticsearch: true, with_keycloak_config: true do
+    let(:user) { create(:user) }
+    let(:investigation) { build(factory) }
+
+    before do
+      User.current = user
+      NotifyMailer.stub(:investigation_created) { double("mailer", deliver_later: true) }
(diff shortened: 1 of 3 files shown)


psd-web/spec/models/investigation_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+# TODO: Refactor Investigation model to remove callback hell and dependency on User.current
+RSpec.shared_examples "an Investigation" do
+  describe "record creation", with_stubbed_elasticsearch: true, with_keycloak_config: true do
+    let(:user) { create(:user) }
+    let(:investigation) { build(factory) }
+
+    before do
+      User.current = user
+      NotifyMailer.stub(:investigation_created) { double("mailer", deliver_later: true) }
+      investigation.save # Need to trigger save after stubbing the mailer due to callback hell
```
StupidCodeFactory:
```suggestion
      investigation.save! # Need to trigger save after stubbing the mailer due to callback hell
```

PR #87, opened 2019-12-04
Fix incorrect test, caused by factories having common names

slorek:
## Description
The previous implementation of this spec was incorrect and giving false positives because the factories of different classes had the same `name` value. This fixes the test and also ensures factories have unique names.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


slorek:
System test failure is unrelated to this change. GitHub checks are stuck and cannot be rerun. Therefore I'm merging this change.

PR #88, opened 2019-12-04
Improve display of cases on case list pages

edwardhorsford:
Fixes some broken css on the case cards and generally improves the information display when cases are restricted or closed.

## Description
We had styles in our codebase for restricted / closed / status - but these weren't working as they relied on extending a compound `&` selector, which I don't think is possible.

I've flattened out the styles a bit and cleaned them up.

Other changes:
* Restricted cases where user is non-authorised now display better
* Restricted and closed cases gain a badge to indicate that status - this means users can more easily tell if a case is restricted.
* Improved focus indication on focussed case cards.

## Before
Screenshot 2019-12-04 at 11 15 19

## After
Non-authorised
Screenshot 2019-12-04 at 13 03 44

Authorised
Screenshot 2019-12-04 at 13 04 22

Mobile
Screenshot 2019-12-04 at 13 07 25

## Testing
I've tested locally, but I don't know if any new tests need adding for these changes.

Suggest @nasirkhanbeis tests the various scenarios - cases closed or open, restricted or unrestricted, and for authorised and non-authorised users.


edwardhorsford:
I should note this also changes the display of case 'cards' when viewed on product and business 'case' tabs:
Screenshot 2019-12-04 at 13 26 25

edwardhorsford:
Focus state:
Screenshot 2019-12-04 at 14 26 32

nasirkhanbeis:
I tested all possible scenarios - looks fine as expected. Just spotted an unrelated bug ,wondering if it's a regression in this branch or not  as I see the link "Back to Allegation" seems broken. 
I tried to replicate this on stage and seems there's no link at all. So steps to replicate this is something like this.
1. Open the case , go to products tab - view product (if there's  a product) Now click Back to Allegation which returns error as below 
image
@edwardhorsford @slorek

edwardhorsford:
I had a chat with @nasirkhanbeis where he demoed the issue.

I can see three things wrong, but I'm not sure they're related to this PR:
1. When viewing a product from a case, we seem to be sometimes showing a back link and sometimes a breadcrumb. I can't work out why it varies.
2. When a back link is shown, it's pointing at an invalid url
3. When visiting that url, a 404 isn't shown, a different error is shown.

Hopefully one of @slorek or @StupidCodeFactory can have a look and verify if these might be related to this pr - else we can raise as separate bugs.

PR #92, opened 2019-12-05
Show case status on search results cards

edwardhorsford:
In #88 I missed that search results use a different template. This adds the status tags to the search results.

Screenshot 2019-12-05 at 16 33 15


LMallows:
Looks good.

nasirkhanbeis:
@edwardhorsford Looking good as expected.
Tested as TS and opss user on cases list page, search page, product/business view page, filtered list page. All good.

PR #93, opened 2019-12-06
Eager load user organisation and teams for assignee filter list to eliminate n+1 database queries

slorek:
## Description
In Scout we are seeing ([example here](https://scoutapm.com/apps/156303/endpoints/Q29udHJvbGxlci9pbnZlc3RpZ2F0aW9ucy9pbmRleA==/trace/1704969432)) over 1,300 database queries taking over 450ms, on the case overview page. The Scout backtrace shows this is coming from rendering the list of users in the 'assigned to' filter list. There are a lot of n+1 database queries created by calling the `display_name` method on the lists of users and teams to render these forms.

The number of database queries is similar to the number of users +teams in the database, so the problem is getting worse as we add more users:

```
irb(main):001:0> User.count
=> 1084
irb(main):002:0> Team.count
=> 217
```

This change uses eager loading to eliminate most of those extra database queries and hopefully significantly improve page load time.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/app/models/team.rb
+++ b/psd-web/app/models/team.rb
@@ -38,7 +38,7 @@ class Team < ApplicationRecord
     self.ensure_names_up_to_date
   end
 
-  def self.get_assignees
+  def self.all_with_organisation
     all.includes(:organisation)
   end
 
--- a/psd-web/app/models/team.rb
+++ b/psd-web/app/models/team.rb
@@ -38,8 +38,12 @@ class Team < ApplicationRecord
     self.ensure_names_up_to_date
   end
 
+  def self.get_assignees
+    all.includes(:organisation)
+  end
+
   def display_name(ignore_visibility_restrictions: false, current_user: User.current)
-    return name if (current_user && (current_user.organisation == organisation)) || ignore_visibility_restrictions
+    return name if (current_user && (current_user.organisation_id == organisation_id)) || ignore_visibility_restrictions
 
     organisation.name
   end
--- a/psd-web/app/models/user.rb
+++ b/psd-web/app/models/user.rb
@@ -129,7 +129,7 @@ class User < ApplicationRecord
 
   def self.get_assignees(except: [])
     user_ids_to_exclude = Array(except).collect(&:id)
-    self.activated.where.not(id: user_ids_to_exclude).preload(:organisation, :teams)
+    self.activated.where.not(id: user_ids_to_exclude).eager_load(:organisation, :teams)
   end
 
   def self.get_team_members(user:)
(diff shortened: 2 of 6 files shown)


psd-web/app/models/user.rb
```diff
@@ -129,7 +129,7 @@ def is_team_admin?
 
   def self.get_assignees(except: [])
     users_to_exclude = Array(except)
-    self.activated - users_to_exclude
+    self.activated.includes(:organisation, :teams) - users_to_exclude
```
frankieroberto: Consider using `eager_load` (which does a left outer join) or `preload` (which does a separate single query), depending on which seems to be faster? `includes` chooses between the two depending on which it thinks will be best, but I’ve found in the past that it’s not always right...

frankieroberto: You could also do something like

```ruby
user_ids_to_exclude = Array(except).collect(&:id)
self.activated.where.not(id: user_ids_to_exclude).includes(:organisation, :teams)
```

to avoid loading the users to exclude?

psd-web/app/models/team.rb
```diff
@@ -38,8 +38,12 @@ def self.load_from_keycloak(teams = KeycloakClient.instance.all_teams(Organisati
     self.ensure_names_up_to_date
   end
 
+  def self.get_assignees
```
frankieroberto:
```suggestion
  def self.with_organisation
```
...might be a more descriptive name? (Feels a bit unconventional to use `get` within a method name in Ruby.)

psd-web/app/models/user.rb
slorek: Good points. I think `includes` automatically defaults to a separate query unless there are conditions e.g. `where`. So adding the `where` would force it to do a left join which I think would be slower , hence the default behaviour of `includes` being a separate query. The only usage I can find of `get_assignees` is to exclude `User.current` so I don't think it would be worth the performance hit.

psd-web/app/models/team.rb
slorek: Thanks, I was a bit uncomfortable with the naming as well.

psd-web/app/models/user.rb
frankieroberto: I _think_ you could force a single query via:

```ruby
self.activated.where.not(id: user_ids_to_exclude).eager_load(:organisation, :teams)`
```

...although it might actually still be quicker to do 2 separate queries, depending on indexing or table size.

Either way, 2 simple queries or 1 join query are both miles faster than n+1.

slorek: I did some quick benchmarking and this way seems significantly faster, thanks!

PR #96, opened 2019-12-09
Fix activities storing wrong html

StupidCodeFactory:
https://trello.com/c/6JF6vzok/215-fix-activities-storing-wrong-html-markdown

## Description
When I introduced the decorators it passed in the activities models a decorated instance of investigations which broke the body description as it would have the decorated version of the description instead of the _raw_ version.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


StupidCodeFactory:
> Anything we can do for faulty data stored in the interim?

we could right a migration I guess, load and resave the audit. I didn't do it as only adds an extra <p> and still looks ok

StupidCodeFactory:
> Anything we can do for faulty data stored in the interim?

PR #98, opened 2019-12-10
Cleanup case templates

edwardhorsford:
Improve the markup used in our case templates:

* Make sure things are correctly in the govuk-grid
* Add hidden text to identical links for accessibility.
* Improve page titles so they're unique to the thing they represent.


nasirkhanbeis:
Improve page titles so they're unique to the thing they represent- looks good! just need to test hidden text before I want to sign off this PR

psd-web/app/views/investigations/tabs/_businesses.html.slim
```diff
@@ -1,21 +1,45 @@
...
-      h3.govuk-heading-s
+      h2.govuk-heading-m
         = ib.relationship.titleize
+
       = business_summary_list ib.business
-      p
-        = link_to "View business", ib.business, class: "govuk-link psd-block-link"
-        = link_to "Remove business", remove_investigation_business_path(@investigation, ib.business)
+
+      - view_business_link_content = capture do
```
slorek: We should create a view helper to construct these links with hidden content.

slorek:
I'm rewriting the failing system tests in RSpec, but this change depends on https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/119.

slorek:
I've replaced the unreliable system test with new RSpec feature specs which cover the existing test scenarios plus more assertions on the intermediary steps.

psd-web/spec/features/add_attachment_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    expect_case_attachments_page_to_show_entered_information
+
+    click_link "Activity"
+
+    expect_case_activity_page_to_show_entered_information
+  end
+
+  def expect_to_be_on_add_attachment_page
+    expect(page).to have_current_path("/cases/#{investigation.pretty_id}/documents/new/upload")
+    expect(page).to have_selector("h1", text: "Add attachment")
```
StupidCodeFactory: I think the latter is sufficient.

psd-web/spec/features/edit_attachment_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    expect_case_attachments_page_to_show_entered_information
+
+    click_link "Activity"
+
+    expect_case_activity_page_to_show_entered_information
+  end
+
+  def expect_to_be_on_edit_attachment_page
+    expect(page).to have_current_path("/cases/#{investigation.pretty_id}/documents/#{document.to_param}/edit")
+    expect(page).to have_selector("h2", text: "Edit document details")
```
StupidCodeFactory: same here I think checking the `h2` with the text is sufficient to prove we're on the correct page, even more so when actually performing other actions in the test that would fail otherwise.

slorek: I felt that also checking the path was a good idea because some of these wizards do not necessarily update the page title between pages. Some of the page titles are much less specific, and it would be inconsistent to check only for page titles sometimes and paths other times.

slorek:
> look good. I'd just prefer to use `have_css`||`have_xpath` over `have_selector` it's clearer what type of selector you're using.

Fair enough. We're using `have_selector` a lot elsewhere already, so i think we should be consistent. Let's change them to `have_css` in one go in a separate PR and then agree to use these going forward.

PR #99, opened 2019-12-10
Various bug fixes

slorek:
## Description
This fixes various bugs encountered:

* Redirect loop to /403 page when trying to sign in as a user with no groups assigned.
* Page title suffix not displayed
* Blank page title on home page
* Home page masthead always displayed when signed out, when it should only be shown on the homepage (ref: @edwardhorsford )
* Help pages content not being displayed when signed out (regression introduced by https://github.com/UKGovernmentBEIS/beis-opss-psd/commit/6bad43e32d65f121a3b593c469fbf11dec6fab6f)
* Error page content not displayed when signed out (regression introduced by https://github.com/UKGovernmentBEIS/beis-opss-psd/commit/6bad43e32d65f121a3b593c469fbf11dec6fab6f)
* Remove "Back to list" link on /403 error page when signed out

Also introduces:

* Add Sign in link in header on help pages when signed out
* Add feature specs for above problematic use cases on forbidden error page and help pages
* Improved realism of RSpec Keycloak mocking so that sign in links can be rendered in feature specs and user groups can be specified when signing in

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


psd-web/app/views/layouts/application.html.slim
```diff
@@ -44,17 +46,18 @@ html.govuk-template.app-html-class[lang="en"]
...
             phaseBanner: User.current && "beta",
             navigation: User.current && nav_items,
-            secondary_navigation: User.current && secondary_nav_items,
+            secondary_navigation: secondary_nav_items,
             assetPath: "media/images"
 
-    - if User.current
+    = yield :masthead
+
+    - content = yield
```
StupidCodeFactory: that's going to render everything twice

StupidCodeFactory: Let's make a non logged in layout?

slorek: I've changed it so `yield` is only called once. Good spot.

edwardhorsford:
I see two bugs with this:

## Missing beta badge
When not on the homepage, the header should have a 'beta' badge next to the service name (we don't have it on the homepage as we have a full beta banner then).

Previously:
Screenshot 2019-12-10 at 17 19 57

## Intro slides
When a trading standards user first signs in, they should be shown the introduction slides. I *think* this no longer shows them - as it takes them to the `/cases` path rather than the root. I *think* it previously worked by sending the user to either root or `/cases` based on whether they were an opss user or not - and now possibly they all go to `/cases`?

This is a bit tricky to test as you basically need to keep creating new ts users to replicate (or clear whatever flag).

slorek:
Thanks. I think these are unrelated to my changes but I’ll look at them separately tomorrow.

PR #100, opened 2019-12-11
Fix beta banner and improve test coverage of home page/declaration/intro journeys

slorek:
## Description
This fixes the "beta" banner not being displayed when signed out (it should only be hidden on the home page as it shows the full banner there - ref: @edwardhorsford ).

I also expanded the test coverage of the home page to cover all of the known use cases. This was to try to reproduce and fix https://trello.com/c/edoGrjPO/243-intro-slides-not-always-shown, but the problem actually laid elsewhere (PR will be raised separately). Nevertheless this improves the test coverage so is valuable to include.

Also fixed the 'sign out' and 'your account' links which were inadvertently removed from the non-OPSS user home page when signed in.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


slorek:
@nasirkhanbeis this was introduced by the home page change and as mentioned in the description will be fixed in a separate PR.

nasirkhanbeis:
> @nasirkhanbeis this was introduced by the home page change and as mentioned in the description will be fixed in a separate PR.

sorry my bad, you mentioned it already. looks good otherwise.

psd-web/spec/features/home_page_spec.rb
```diff
@@ -0,0 +1,109 @@
+require "rails_helper"
+
+RSpec.feature "Home page", :with_keycloak_config, :with_elasticsearch do
+  context "User signed out" do
+    scenario "shows the home page" do
+      sign_out
+      visit "/"
```
StupidCodeFactory:
```suggestion
      visit root_path
```

psd-web/spec/features/home_page_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+
+      expect(page).not_to have_css(".psd-header .govuk-phase-banner__content__tag")
+      expect(page).to have_css(".govuk-phase-banner")
+
+      expect(page).to have_text("Report, track and share product safety information with the product safety community.")
```
StupidCodeFactory: No biggy but `have_(text|content)` matchers are much slower than `have_css` counterpart. If it's only a couple here and there it not an issue but as the tests suite grows this can easily add minutes to the test suite.

psd-web/spec/features/home_page_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+    before do
+      sign_in(as_user: create(:user, user_state, role, has_accepted_declaration: has_accepted_declaration, has_viewed_introduction: has_viewed_introduction))
+      visit "/"
+
+      expect(page).to have_css(".psd-header .govuk-phase-banner__content__tag")
+      expect(page).not_to have_css(".govuk-phase-banner")
+
+      expect(page).to have_link("Sign out")
+      expect(page).to have_link("Your account")
+      expect(page).not_to have_link("Sign in")
```
StupidCodeFactory: Not really a massive fan of expectations in the `before` block (nor anything that is not setup).
An I'd rather have those expect in method that describes this group of expectations.
It seems that we are `def expect_to_be_signed_in`

psd-web/spec/features/home_page_spec.rb
slorek: Following your comment I did try changing them all over to have_css but the execution time of the tests was the same. If the page was very large with many more elements and complicated CSS perhaps this would have a noticeable impact.

PR #101, opened 2019-12-12
Fix sign in callback redirect and add request specs for this functionality

slorek:
## Description
This fixes a bug whereby non-OPSS users would be redirected to /cases after sign in, rather than the correct path which could be either the introduction slides or the root path (which is a different page with a link to /cases), depending on whether the user has previously viewed the slides.

The home page controller correctly manages this so we now redirect to this controller.

Also added request specs around all of the scenarios handled by the callback action. This functionality wasn't covered by any existing unit test.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


nasirkhanbeis:
@slorek  can you pls spin up a review app? 
image

StupidCodeFactory:
> @slorek can you pls spin up a review app?
> image

@nasirkhanbeis this happens when the PR is being deployed

slorek:
Review app deploy is broken for now, working on a fix.

PR #107, opened 2019-12-13
Move deploys to GitHub actions

frankieroberto:
This moves the deployments from happening via Travis to GitHub Actions.

In addition it:

* combines the `manifest.prod.yml` and `manifest.yml`, for simplicity and to minimise differences between staging and production.
* defines a single application within `manifest.yml` which contains both the web and worker processes, to avoid having to push separate apps.
* uses the cf7 CLI instead of the version 6 CLI. This means that `--strategy rolling` can be used instead of having to do manual blue-green deployments.


psd-web/manifest.yml
```diff
@@ -1,20 +1,23 @@
 ---
-defaults: &defaults
+applications:
+- name: ((app-name))
   buildpacks:
     - ruby_buildpack
-  memory: 1G
+  memory: 2G
   instances: 1
```
slorek: The main reason we had a separate `prod` file was because the `instances` on prod is set to 4. We don't need that on staging.

frankieroberto: Ah ok. I’ll make that a variable so it can be set differently for each environment.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,46 @@
...
+      env:
+        CF_CLI_VERSION: 7.0.0-beta.25
+      run: |
+        curl -L "https://packages.cloudfoundry.org/stable?release=linux64-binary&version=${CF_CLI_VERSION}" | tar -zx -C /tmp
+        sudo cp /tmp/cf7 /usr/local/bin/cf7
+    - name: Deploy to staging
+      env:
+        SPACE: staging
+        APP_NAME: psd-web
+        DOMAIN: www.product-safety-database.service.gov.uk
```
slorek: The domains need to be swapped between staging/prod namespaces.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,46 @@
...
+        cf7 auth
+        cf7 target -o 'beis-opss' -s $SPACE
+        chmod +x ./cosmetics-web/deploy.sh
+        ./cosmetics-web/deploy.sh
+        cf7 logout
+    - name: Deploy to production
+      env:
+        SPACE: prod
+        APP_NAME: psd-web
+        DOMAIN: staging.product-safety-database.service.gov.uk
```
slorek: See above.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,46 @@
...
+        CF_USERNAME: ${{ secrets.PaaSUsernameProduction }}
+        CF_PASSWORD: ${{ secrets.PaaSPasswordProduction }}
+      run: |
+        cf7 api api.london.cloud.service.gov.uk
+        cf7 auth
+        cf7 target -o 'beis-opss' -s $SPACE
+        ./cosmetics-web/deploy.sh
```
slorek: `psd-web`

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,46 @@
...
+        APP_NAME: psd-web
+        DOMAIN: www.product-safety-database.service.gov.uk
+        CF_USERNAME: ${{ secrets.PaaSUsernameStaging }}
+        CF_PASSWORD: ${{ secrets.PaaSPasswordStaging }}
+      run: |
+        cf7 api api.london.cloud.service.gov.uk
+        cf7 auth
+        cf7 target -o 'beis-opss' -s $SPACE
+        chmod +x ./cosmetics-web/deploy.sh
+        ./cosmetics-web/deploy.sh
```
slorek: `psd-web`

psd-web/deploy.sh
```diff
@@ -1,84 +1,14 @@
...
-# Unmap the live hostanme from the old app
-cf unmap-route $APP $DOMAIN --hostname $HOSTNAME
+# Deploy the app
+cf7 push -f $MANIFEST_FILE $APP_NAME --var psd-host=$DOMAIN --strategy rolling
 
-# Remove the old app and rename the new one
-cf delete -f $APP
-cf rename $NEW_APP $APP
+# Remove the copied infrastructure env files to clean up
+rm -R cosmetics-web/env/
```
slorek: `psd-web`

psd-web/manifest.yml
slorek: Do we need `instances` at the top level? I think it should be in `processes`.

.github/workflows/deploy.yml
frankieroberto: Opps, 🤦‍♂

psd-web/manifest.yml
slorek: Same with `memory`.

psd-web/deploy.sh
frankieroberto: 👍

psd-web/manifest.yml
```diff
@@ -24,28 +27,14 @@ web_defaults: &web_defaults
...
-    - psd-notify-env
-    - psd-rails-env
-    - psd-sentry-env
-    - psd-keycloak-env
+    - psd-web-scout-env
+  processes:
+    - type: web
+      command: export $(./env/get-env-from-vcap.sh) && STATEMENT_TIMEOUT=60s bin/rake cf:on_first_instance db:migrate keycloak:sync && bin/rails server -b 0.0.0.0 -p $PORT -e $RAILS_ENV
+      instances: 1
+      memory: 1G
```
slorek: 2G

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,54 @@
...
+        CF_USERNAME: ${{ secrets.PaaSUsernameStaging }}
+        CF_PASSWORD: ${{ secrets.PaaSPasswordStaging }}
+      run: |
+        cf7 api api.london.cloud.service.gov.uk
+        cf7 auth
+        cf7 target -o 'beis-opss' -s $SPACE
+        chmod +x ./psd-web/deploy.sh
+        ./psd-web/deploy.sh
+        cf7 logout
+    - name: Deploy to production
```
StupidCodeFactory: can we have this as separate job which only run once the staging job has successfully run?

frankieroberto: @StupidCodeFactory that should happen already. There's one job, but two steps, and each step is only run if the previous step completed successfully.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,54 @@
...
+
+jobs:
+  deploy:
+    name: Deploy to Staging and Production
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Install cf client
+      env:
+        CF_CLI_VERSION: 7.0.0-beta.25
```
slorek: `beta28` is the current version.

frankieroberto: updated, ta

slorek:
This looks OK, would like to see the review app deploy successfully and test that before approval.

frankieroberto:
@slorek review app is deploying fine now: https://psd-pr-107.london.cloudapps.digital – the `psd-pr-107` app in cloudfoundry now contains separate web and worker instances. Can you login and verify that the worker instance is processing the jobs ok?

slorek:
Confirmed Sidekiq web is accessible and the jobs appear to be processing without errors.

StupidCodeFactory:
@dependabot-bot rebase

PR #108, opened 2019-12-16
Replace existing Trading Standards case creation system test with RSp…

slorek:
## Description
The existing system test for Trading Standard user creating a case is unreliable. Additionally it is testing with fixture data that already exists in the database, so the quality of the test is questionable. In order to build on this feature with confidence and in line with our strategy to move to RSpec I have re-implemented the test in RSpec.

I've basically reproduced the test steps as they were but added a lot more rigorous testing and guaranteed random fake data which doesn't already exist. It's not really possible to break this test down into smaller scenarios without making it very slow. It also runs much faster (3.5s vs 11s). I decided not to use FactoryBot to construct the test data because that doesn't really represent the way the user enters data. It would be like working backwards.

Removing the old test should also make our CI test runs much more reliable, since the old one would randomly fail. I'm not quite sure why yet, but I suspect it's related to assumptions with the fixture data, which is the reason other tests were randomly failing.

I'm not entirely sure if this is the best way to construct a long user journey test; I've tried to compose it of somewhat-reusable step methods so that different scenarios can be added in future. However the file is very long (twice as many lines as the old one). I would appreciate any feedback to improve the way the test is constructed.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


nasirkhanbeis:
Thanks @slorek for talking through tests in this PR, from testing perspective I'm happy as it covers pretty much happy path test with all the critical forms which is good.
Any error validations around forms will be covered in further feature changes as we go along. Good work! @slorek

PR #109, opened 2019-12-17
Improve product page display and link to add images

edwardhorsford:
This PR:

* Cleans up the display of the product page.
* * Swap image and heading so the page looks cleaner when there are no images.
* * Promote the product type and category more - as they're often useful for identifying what the product is (names are often not clear).
* * Promote the date created and updated more
* Removes the duplicate summary table.
* Shows a link to prompt users to add an image.
* Removes the ability for the summary list to exclude the batch number - as a byproduct this will now appear on case page product tab. I've attempted to update the tests but please check if they're ok.

Tests:
* Could possibly do with a test for the new decorator I made?
* I guessed at the format for the updated date - and it seems to work - but is it a reliable thing to use?

Components
* I'd like to make a pattern out of displaying the date created and updated (not quite this design) - so would like to replace what I've got here with some helper or something.

## Before:

Without an image:
2019 12 17-13-42-09-View product - Product safety database - GOV UK
With an image:
2019 12 17-13-44-45-Improve product page display and link to add images - DO NOT MERGE by edwardhorsford · Pull Request -109 · UKGovernmentBEIS-beis-opss-psd

##After:
Without an image:
2019 12 17-13-43-44-View product - Product safety database - GOV UK
With an image:
2019 12 17-13-47-11-View product - Product safety database - GOV UK


psd-web/app/decorators/product_decorator.rb
```diff
@@ -20,4 +20,23 @@ def summary_list(include_batch_number: false)
...
+    # nil if neither exists
+    if product_type.present? && category.present?
+      product_type + " (" + category.downcase + ")"
+    elsif product_type.present?
+      product_type
+    elsif category.present?
+      category
+    else
+      nil
+    end
```
StupidCodeFactory:
```ruby
def all_categories
  product_and_category = [ product_type, category ].flatten
  return "#{product_and_category.first} (#{product_and_category.last})" if product_and_category.length > 1
  product_and_category.first
end
```

StupidCodeFactory: Also the name of the method does not describe what if does, what about `product_type`?

psd-web/app/views/products/show.html.slim
```diff
@@ -2,8 +2,30 @@
...
+      span.govuk-caption-l
+        | Product
+      = @product.name
+
+    - if @product.all_categories.present?
+      p.govuk-body-l
+        = @product.all_categories
+
+    p.govuk-caption-m
+      = "Created: " + @product.created_at.strftime("%d %B %Y")
```
StupidCodeFactory:
```suggestion
      = "Created: " + @product.created_at.to_s(:govuk)
```

psd-web/app/views/products/show.html.slim
```diff
@@ -2,8 +2,30 @@
...
+      - if @product.updated_at != @product.created_at
+        br
+        = "Updated: " + @product.updated_at.strftime("%d %B %Y")
```
StupidCodeFactory:
```suggestion
        = "Updated: " + @product.updated_at.to_s(:govuk)
```

psd-web/app/views/products/show.html.slim
edwardhorsford: I copied what was already in the page - but can update to this.

psd-web/app/decorators/product_decorator.rb
edwardhorsford: We already have something called `product_type`.

I feel like your suggested method might be harder to read. Thoughts?

What does it return if neither exist? nul?

edwardhorsford: Have renamed to `combined_categories`

StupidCodeFactory: I've updated the previous snippet, I've removed the early "return"  (which is nice as it avoids doing the else), those are pretty idiomatic in ruby, in fact the linter might actually ask to do an early return.
so if `product_and_category` is empty it `product_and_category.first` will return `nil` (null in ruby)

So for the name, I'd suggest `product_type_and_category_label`. `combined_categories` is still incorrect as it doesn't describe what is being done inside the function.

nasirkhanbeis:
Tested this in review app, all good!

PR #113, opened 2019-12-23
Rework blue intro slides and make govuk-frontend compliant WIP

edwardhorsford:
Replaces our previous blue intro slides sass with some simpler styles adapted from govuk-frontend and that are more BEM-compliant.

Makes `--inverted` modifiers for components we use on the masthead and intro slides.

Has a minor visual impact:

Before:
Screenshot 2019-12-23 at 13 33 56

After:
Screenshot 2019-12-23 at 13 34 10


psd-web/app/assets/stylesheets/button.scss
```diff
@@ -9,3 +9,13 @@
   @include remove-default-button-styling;
   @include govuk-link-common;
 }
+
+// Larger buttons like govuk-start buttons, but without the chevron
+.app-button--l {
+  // Below styles from gouuk-button--start
```
frankieroberto: Minor typo:
```suggestion
  // Below styles from govuk-button--start
```

psd-web/app/assets/stylesheets/inverted.scss
```diff
@@ -0,0 +1,79 @@
+// Blue card styles - used by masthead and intro slides
+
+.app-inverted {
+  color: govuk-colour("white");
+  background-color: govuk-colour("blue");
+
+  * {
```
frankieroberto: This feels kinda dangerous. Might be better to explicitly set the foreground colour on any child elements that need it (even if this is more verbose)?

psd-web/app/views/introduction/report_products.html.slim
```diff
@@ -19,4 +19,5 @@
           relevant product safety legislation as soon as possible.
 
       = link_to "Continue", introduction_track_investigations_path,
-              class: "govuk-button", role: "button", draggable: false
+              class: "govuk-button app-button--inverted app-button--l \
```
frankieroberto: You could use our Rails helper here:

```ruby
govukButton(
  href: introduction_track_investigations_path, 
  text:  "Continue",  
  classes: "app-button--inverted app-button--l govuk-!-margin-bottom-0"
)
```

(Same options as the [Nunjucks macro](https://design-system.service.gov.uk/components/button/))

edwardhorsford: Suggest leaving this for a separate pr.

psd-web/app/assets/stylesheets/inverted.scss
edwardhorsford: Have made individual modifiers for each element.

PR #115, opened 2019-12-23
Add extra questions to corrective actions forms

slorek:
## Description
https://trello.com/c/MuvHX29n/206-5corrective-action-add-new-questions-and-update-content

This implements three of the four new questions specified in the above ticket, plus the content changes.

I've also replaced the existing system/controller/model tests with RSpec equivalents.

## Checklist:

### General testing
- [ ] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only
- [x] Disable css - does content make sense and appear in a logical order?


psd-web/app/views/investigations/corrective_actions/_form.html.slim
```diff
@@ -44,8 +45,32 @@
...
+                { text: CorrectiveAction.human_attribute_name("duration.temporary"), value: "temporary" },
+                { text: CorrectiveAction.human_attribute_name("duration.unknown"), value: "unknown" }]
+
+= render "form_components/govuk_select",
+         choices: corrective_action_geographic_scope,
+         key: :geographic_scope,
+         form: form,
+         is_autocomplete: true,
+         show_all_values: true,
+         formGroup: { classes: "govuk-!-width-one-half" },
```
edwardhorsford: I think we should remove this width class for now as it also applies to the label text. We'll need a separate piece of work to enable the autocomplete width to be set properly.

psd-web/app/views/investigations/corrective_actions/confirmation.html.slim
```diff
@@ -25,6 +25,17 @@ h1.govuk-heading-l
...
+          td.govuk-table__cell
+            = CorrectiveAction.human_attribute_name("measure_type.#{@corrective_action.measure_type}")
+        tr.govuk-table__row
+          th.govuk-table__header scope="row" Corrective action in place
+          td.govuk-table__cell
+            = CorrectiveAction.human_attribute_name("duration.#{@corrective_action.duration}")
+        tr.govuk-table__row
+          th.govuk-table__header scope="row" Geographic scope
+          td.govuk-table__cell = @corrective_action.geographic_scope
         - if @corrective_action.details.present?
```
edwardhorsford: Not something you changed, but I think this is broken.

psd-web/config/locales/en.yml
```diff
@@ -125,4 +134,14 @@ en:
...
               blank: "Enter a valid product category"
-
+    attributes:
+      corrective_action:
+        measure_type:
+          one: "Type of measure"
+          mandatory: "Mandatory"
+          voluntary: "Voluntary"
+        duration:
+          one: "Corrective action in place"
```
edwardhorsford: Duration of action

psd-web/spec/features/add_corrective_action_spec.rb
```diff
@@ -0,0 +1,147 @@
...
+
+  def expect_confirmation_page_to_show_entered_data
+    expect(page.find("th", text: "Summary")).to have_sibling("td", text: summary)
+    expect(page.find("th", text: "Date decided")).to have_sibling("td", text: date.strftime("%d/%m/%Y"))
+    expect(page.find("th", text: "Legislation")).to have_sibling("td", text: legislation)
+    expect(page.find("th", text: "Details")).to have_sibling("td", text: details)
+    expect(page.find("th", text: "Attachment", match: :prefer_exact)).to have_sibling("td", text: File.basename(file))
+    expect(page.find("th", text: "Attachment description")).to have_sibling("td", text: file_description)
+    expect(page.find("th", text: "Type of measure")).to have_sibling("td", text: CorrectiveAction.human_attribute_name("measure_type.#{measure_type}"))
+    expect(page.find("th", text: "Corrective action in place")).to have_sibling("td", text: CorrectiveAction.human_attribute_name("duration.#{duration}"))
```
edwardhorsford: Duration of action

psd-web/app/views/investigations/corrective_actions/confirmation.html.slim
```diff
@@ -25,6 +25,17 @@ h1.govuk-heading-l
           tr.govuk-table__row
             th.govuk-table__header scope="row" Business
             td.govuk-table__cell = @corrective_action.business.trading_name
+        tr.govuk-table__row
+          th.govuk-table__header scope="row" Type of measure
+          td.govuk-table__cell
+            = CorrectiveAction.human_attribute_name("measure_type.#{@corrective_action.measure_type}")
+        tr.govuk-table__row
+          th.govuk-table__header scope="row" Corrective action in place
```
edwardhorsford: Can this be `Duration of action` please?

psd-web/app/models/audit_activity/corrective_action/base.rb
```diff
@@ -22,7 +22,10 @@ def self.build_body(corrective_action)
     body += "Product: **#{self.sanitize_text corrective_action.product.name}**<br>" if corrective_action.product.present?
     body += "Legislation: **#{self.sanitize_text corrective_action.legislation}**<br>" if corrective_action.legislation.present?
     body += "Business responsible: **#{self.sanitize_text corrective_action.business.trading_name}**<br>" if corrective_action.business.present?
-    body += "Date decided: **#{corrective_action.date_decided.strftime('%d/%m/%Y')}**<br>" if corrective_action.date_decided.present?
+    body += "Date came into effect: **#{corrective_action.date_decided.strftime('%d/%m/%Y')}**<br>" if corrective_action.date_decided.present?
+    body += "Type of measure: **#{CorrectiveAction.human_attribute_name("measure_type.#{corrective_action.measure_type}")}**<br>"
+    body += "Action in place: **#{CorrectiveAction.human_attribute_name("duration.#{corrective_action.duration}")}**<br>"
```
edwardhorsford: Duration of action

nasirkhanbeis:
minor issue - displaying two version of error message for duration field 
image
Expected to display only one

psd-web/spec/models/corrective_action_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+  let(:date_decided) { Faker::Date.backward(days: 14) }
+  let(:legislation) { Rails.application.config.legislation_constants["legislation"].sample }
+  let(:measure_type) { CorrectiveAction::MEASURE_TYPES.sample }
+  let(:duration) { CorrectiveAction::DURATION_TYPES.sample }
+  let(:geographic_scope) { Rails.application.config.corrective_action_constants["geographic_scope"].sample }
+  let(:details) { Faker::Lorem.sentence }
+  let(:related_file) { "No" }
+
+  describe "#valid?" do
+    subject { corrective_action.valid? }
```
StupidCodeFactory: This is a bit of an anti-pattern. 
In an RSpec test, by convention, `subject` is almost always an instance of `described_class`. The only time it makes sense for `subject` not be an instance of `described_class` is in `request` specs and `controller` tests. 

The problem with re-assigning the `subject` is that when you open a test file you have to scroll up and back down again to try and out what is the subject which adds to the cognitive load of understanding the test.
Now most tests will read the same `expect(subject).to be true` and are not understandable by themself.

RSpec provides awesome syntactic sugar for predicate methods.
`expect(subject).to be_valid` 
This is far more readable because we understand within the body of the test what `subject` is and also what method is under test.

psd-web/db/migrate/20191223162619_add_attributes_to_corrective_actions.rb
```diff
@@ -0,0 +1,11 @@
+class AddAttributesToCorrectiveActions < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured do
+      change_table :corrective_actions, bulk: true do |t|
+        t.string :measure_type
```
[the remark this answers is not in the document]
slorek: I added a comment in Trello:

>>> The activity card content is generated when the activity occurs, so all existing corrective action activities are not going to be effected by content changes. I've added the new fields and changed the 'Date decided' to 'Date came into effect', but this change will only affect new actions added. I think this is fair since the semantics have changed.

psd-web/spec/models/corrective_action_spec.rb
slorek: Thanks, I didn't know about `be_valid`.

psd-web/db/migrate/20191223162619_add_attributes_to_corrective_actions.rb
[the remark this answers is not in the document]
slorek: It won't be possible to migrate the previous actions because this data is not known at all, unfortunately. It would have been nice to set `not null` on the columns.

slorek:
> I get a form resubmission error if I add an action and then return to page. I'm guessing this is an existing problem with the page though (not using post-redirect-get pattern).

Could you please describe in more detail how to reproduce this error?

psd-web/app/views/investigations/corrective_actions/confirmation.html.slim
slorek: Seems to be working for me.

psd-web/db/migrate/20191223162619_add_attributes_to_corrective_actions.rb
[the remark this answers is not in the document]
StupidCodeFactory: Damn, I had the feeling that might not have been possible.

slorek:
> minor issue - displaying two version of error message for duration field

Fixed.

nasirkhanbeis:
> Added some minor text changes.
> 
> I get a form resubmission error if I add an action and then return to page. I'm guessing this is an existing problem with the page though (not using post-redirect-get pattern).

Hi @edwardhorsford I think you mean browser back to corrective action confirm page ?I tested this and seems bit wider issue across other actions too , user can do browser back and re-submit the same action twice which I think isn't expected.I couldn't replicate resubmission error though. I suggest we can fix this properly in a separate card.

edwardhorsford:
> > I get a form resubmission error if I add an action and then return to page. I'm guessing this is an existing problem with the page though (not using post-redirect-get pattern).
> 
> Could you please describe in more detail how to reproduce this error?

I think the issue is that if you have form errors you're returned to the page after the post without being redirected. If you then progress to the confirmation page and attempt to go back, you'll get an error. I assume this might be the same for all cases where we return the user to the page after a form error.

psd-web/app/views/investigations/corrective_actions/confirmation.html.slim
edwardhorsford: My reading of it is that it shouldn't display if it wasn't answered - but it does - just blank.

edwardhorsford:
Screen Shot 2019-12-27 at 10 45 09


Example of details being shown despite not being answered - I assume that `.present?` was meant to stop it being shown?

PR #116, opened 2019-12-27
Sort search result by relevance

StupidCodeFactory:
## Description:

 Search improvements:
1. Fix products and businesses listing page which where only returning more 10
items per page.
2. Fix products and businesses default search sort order on the listing page,
now orderd by created_at descending order
3. Products, Businesses and Investigatio

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


psd-web/app/decorators/investigations_decorator.rb
```diff
@@ -1,2 +1,5 @@
 class InvestigationsDecorator < Draper::CollectionDecorator
+  def total_pages
```
slorek: Shall we just `delegate` this?

psd-web/spec/features/businesses_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+RSpec.feature "Business listing", :with_elasticsearch, :with_stubbed_mailer, :with_stubbed_keycloak_config do
+  let(:user)            { create :user, :activated }
+  let!(:business_one)   { create(:business, created_at: 1.day.ago) }
+  let!(:business_two)   { create(:business, created_at: 2.days.ago) }
+  let!(:business_three) { create(:business, created_at: 3.days.ago) }
+
+  before { create_list :business, 18, created_at: 4.days.ago }
+
+  scenario "lists products according to search relevance" do
+    Business.import(refresh: :wait_for)
```
slorek: I wonder if we need this, I have yet to encounter any issues with ES not indexing a record in time?

psd-web/spec/features/filter_investigations_spec.rb
```diff
@@ -17,6 +17,7 @@
   let!(:another_inactive_user) { create(:user, :inactive, organisation: user.organisation, teams: [team]) }
 
   before do
+    Investigation.import refresh: :wait_for
```
slorek: As previous comment.

psd-web/spec/features/investigations_spec.rb
```diff
@@ -0,0 +1,28 @@
+require "rails_helper"
```
slorek: Worth testing pagination here as well, like on the businesses spec?

psd-web/spec/features/products_spec.rb
```diff
@@ -0,0 +1,37 @@
+require "rails_helper"
+
+RSpec.feature "Products listing", :with_elasticsearch, :with_stubbed_mailer, :with_stubbed_keycloak_config do
```
slorek: Worth testing pagination here as well, like on the businesses spec?

psd-web/spec/features/products_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  let(:user)             { create :user, :activated }
+  let!(:iphone)          { create(:product_iphone, created_at: 1.day) }
+  let!(:iphone_3g)       { create(:product_iphone_3g, created_at: 2.days) }
+  let!(:washing_machine) { create(:product_washing_machine, created_at: 3.days) }
+
+  scenario "lists products according to search relevance" do
+    Product.import(refresh: :wait_for)
```
slorek: See previous comment.

psd-web/spec/features/products_spec.rb
```diff
@@ -0,0 +1,37 @@
+require "rails_helper"
+
+RSpec.feature "Products listing", :with_elasticsearch, :with_stubbed_mailer, :with_stubbed_keycloak_config do
+  let(:user)             { create :user, :activated }
+  let!(:iphone)          { create(:product_iphone, created_at: 1.day) }
+  let!(:iphone_3g)       { create(:product_iphone_3g, created_at: 2.days) }
+  let!(:washing_machine) { create(:product_washing_machine, created_at: 3.days) }
+
+  scenario "lists products according to search relevance" do
```
slorek: I wonder if we should split this into two scenarios, as we are actually testing the default sort order first, and then search relevance.

psd-web/spec/services/elasticsearch_query_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+
+  def perform_search
+    Investigation.full_search(subject)
+  end
+
+  # TODO: these specs are a port of the deprecated (and flaky) Minitest tests.
+  # Currently full_search returns pretty much every record so further work is
+  # needed to improve the relevance of the results. We should then add
+  # assertions that irrelevant records are *not* returned here.
+  describe "#build_query" do
```
slorek: I think we are actually testing `#full_search` on the model. `#build_query` is the argument.

psd-web/spec/services/elasticsearch_query_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+
+      context "searching by business trading name" do
+        let(:query) { business.trading_name }
+
+        it_behaves_like "finds the relevant investigation"
+      end
+
+      context "searching by business number" do
+        let(:query) { business.company_number }
+
```
slorek: Extra line

psd-web/spec/services/elasticsearch_query_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+      context "searching by complainant email address" do
+        let(:query) { complainant.email_address }
+
+        it_behaves_like "finds the relevant investigation"
+      end
+    end
+
+    context "when searcing on a business" do
+      let!(:business) { create(:business, investigations: [investigation]) }
```
slorek: I found I had to manually index the investigation when doing this in the previous implementation, I'm curious why this now works without.

psd-web/app/decorators/investigations_decorator.rb
StupidCodeFactory: I think I can actually get rid of it all together

psd-web/spec/features/businesses_spec.rb
StupidCodeFactory: will give it a go

psd-web/spec/features/investigations_spec.rb
StupidCodeFactory: yes was going to add it

psd-web/spec/features/products_spec.rb
StupidCodeFactory: I think it is not really worth. It will make the tests run slower.

psd-web/spec/services/elasticsearch_query_spec.rb
StupidCodeFactory: So this is debatable, sadly the `.full_search` is calling the `#build_query`. which is think should be the thing passed into the `.full_search` so that it not coupled to `ElasticsearchQuery` interface.

Ideally, this test should test that `Hash` structure returned by `#build_query` is correct, I, however, found ridiculously complicated to understand what it is building and thus prefered to call `.full_search` and check that the correct investigation record was returned.

psd-web/spec/services/elasticsearch_query_spec.rb
StupidCodeFactory: This works as I'm also reindexing the document https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/116/files#diff-87c64ae9bd8e89e80889abfcde7c440cR5

In the real world, this also won't work, we should really build service/command classes to perform all the necessary actions (1. validate record, 2. save the record, 3 re-index ,4, notify if needed)

slorek: Right, thanks. I found that re-indexing wasn't necessary generally, so we should probably only do so where needed, and as you say this will help us identify issues which mean we should include re-indexing in a service which handles the change.

psd-web/spec/features/filter_investigations_spec.rb
StupidCodeFactory: so this is actually needed, I think it is because I have removed the import at the bottom of the `app/models/investigation.rb`

edwardhorsford:
Have given it a play.

Some issues:

* When search term is used, the user is redirected to `/investigations/search` rather than /cases/search`
* When the search term is removed, the user is not redirected back to `/cases`
* When the search term is removed, the radio for 'sort by relevance' still exists.
* The cases link in the nav which is meant to restore your previous filters doesn't seem to restore the sort order when sort order is not relevancy.

edwardhorsford:
Checked again. My four issues above are now resolved.

Two further issues (already discussed with Yann):

* Visiting `/cases/search` without query params throws an error
* There should be an active state in the nav at the top on cases.

StupidCodeFactory:
> Checked again. My four issues above are now resolved.
> 
> Two further issues (already discussed with Yann):
> 
>     * Visiting `/cases/search` without query params throws an error
> 
>     * There should be an active state in the nav at the top on cases.

it's been fixed

PR #119, opened 2019-12-27
Allow antivirus checks and other background jobs to work inline in RSpec suite

slorek:
## Description
This ensures that background jobs (including antivirus checks) run inline during RSpec specs.

It introduces a new requirement for `ANTIVIRUS_URL` environment variable to be set, otherwise an exception will be thrown. I've updated the example .env files to reflect this. Please copy the `ANTIVIRUS_*` env vars from the example to your `.env.development` file.

It doesn't mean the API always has to be run in development mode as it will normally fail silently in the background job if it can't connect. Running it will allow you to add attachments to cases, however.

I've added the `antivirus` service to the `docker-compose.yml` file to allow Antivirus to be run locally:

```bash
$ docker-compose up antivirus
```

This change also addresses minor issues with email notification in the test environment and adds helpers to stub mail sending and antivirus scanning.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/.github/workflows/system_tests.yml
+++ b/.github/workflows/system_tests.yml
@@ -93,12 +93,9 @@ jobs:
         PGHOST: localhost
         PGUSER: postgres
         RAILS_ENV: test
-        ANTIVIRUS_URL: ${{ secrets.AntivirusUrl }}
-        ANTIVIRUS_USERNAME: ${{ secrets.AntivirusUsername }}
-        ANTIVIRUS_PASSWORD: ${{ secrets.AntivirusPassword }}
       run: |
         cd psd-web
-        bin/rails db:setup
+        bin/rails db:create db:schema:load
     - name: Run Tests
       env:
         ELASTICSEARCH_URL: localhost
--- a/.github/workflows/system_tests.yml
+++ b/.github/workflows/system_tests.yml
@@ -93,6 +93,9 @@ jobs:
         PGHOST: localhost
         PGUSER: postgres
         RAILS_ENV: test
+        ANTIVIRUS_URL: ${{ secrets.AntivirusUrl }}
+        ANTIVIRUS_USERNAME: ${{ secrets.AntivirusUsername }}
+        ANTIVIRUS_PASSWORD: ${{ secrets.AntivirusPassword }}
       run: |
         cd psd-web
         bin/rails db:setup
--- a/psd-web/app/analyzers/anti_virus_analyzer.rb
+++ b/psd-web/app/analyzers/anti_virus_analyzer.rb
@@ -5,7 +5,7 @@ class AntiVirusAnalyzer < ActiveStorage::Analyzer
 
   def metadata
     download_blob_to_tempfile do |file|
-      response = RestClient::Request.execute method: :post, url: ENV["ANTIVIRUS_URL"], user: ENV["ANTIVIRUS_USERNAME"], password: ENV["ANTIVIRUS_PASSWORD"], payload: { file: file }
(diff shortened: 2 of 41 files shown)


psd-web/app/analyzers/anti_virus_analyzer.rb
```diff
@@ -5,7 +5,7 @@ def self.accept?(_blob)
 
   def metadata
     download_blob_to_tempfile do |file|
-      response = RestClient::Request.execute method: :post, url: ENV["ANTIVIRUS_URL"], user: ENV["ANTIVIRUS_USERNAME"], password: ENV["ANTIVIRUS_PASSWORD"], payload: { file: file }
+      response = RestClient::Request.execute method: :post, url: ENV.fetch("ANTIVIRUS_URL", "http://antivirus"), user: ENV["ANTIVIRUS_USERNAME"], password: ENV["ANTIVIRUS_PASSWORD"], payload: { file: file }
```
StupidCodeFactory: can we add those to `config.application.rb` so that all configuration from ENV vars are all in the same place?
Than can be access as such: `Rails.applucation.config.antivirus_url`

psd-web/spec/rails_helper.rb
```diff
@@ -62,4 +62,6 @@
   config.filter_rails_from_backtrace!
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
+
+  ActiveJob::Base.queue_adapter = :inline
```
StupidCodeFactory: maybe it makes more sense to add this to `config/environments/test.rb`
`config.active_job.queue_adapter = :inline`

psd-web/spec/support/mailer.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+RSpec.shared_context "with stubbed mailer", shared_context: :metadata do
+  before { allow_any_instance_of(NotifyMailer).to receive(:mail) { true } }
```
StupidCodeFactory: I'm a bit worried about using `allow_any_instance_of` this is veru likely to cause problems further down the line.

psd-web/spec/rails_helper.rb
slorek: Good point.

psd-web/spec/support/mailer.rb
slorek: I agree, but I'm not sure there is a better way in this instance.

.github/workflows/system_tests.yml
```diff
@@ -95,7 +95,7 @@ jobs:
         RAILS_ENV: test
       run: |
         cd psd-web
-        bin/rails db:setup
+        bin/rails db:create db:schema:load
```
slorek: This avoids it trying to connect to the antivirus API inline when loading seeds, which aren't required in tests.

psd-web/test/test_helper.rb
```diff
@@ -14,6 +14,10 @@
 require "rails/test_help"
 require "rspec/mocks/standalone"
 
+# Added Webmock only to allow use of stub_request - Minitest suite is deprecated
+require "webmock/minitest"
+WebMock.allow_net_connect!
```
slorek: Standard behaviour is to disallow network connections, but the tests assume they can connect to Keycloak and Elasticsearch. I would have to stub them in almost every test, and since we are slowly migrating all tests to RSpec I decided to keep the environment working as currently.

PR #120, opened 2019-12-27
remove dodgy date_attribute

StupidCodeFactory:
## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


psd-web/app/helpers/application_helper.rb
```diff
@@ -15,4 +15,16 @@ def markdown(text)
     sanitized_input = sanitize(text, tags: %w(br))
     rc.render(sanitized_input).html_safe # rubocop:disable Rails/OutputSafety
   end
+
+  def itens_for_govuk_date_input(form, attribute)
```
edwardhorsford: typo

psd-web/app/helpers/application_helper.rb
```diff
@@ -15,4 +15,16 @@ def markdown(text)
...
+    dom_class = dom_class(form.object)
+    name = ->(i) { "#{dom_class}[#{attribute}](#{i}i))" }
+    id = ->(i) { "#{dom_class}_#{attribute}_#{i}i" }
+    value = ->(date_part) { form.object.public_send(attribute)&.public_send(date_part) }
+    [
+      { name: name.(3), value: value.(:day),   classes: "govuk-input--width-2", label: "Day",   id: id.(3) },
```
edwardhorsford: What are these three items? Do we need to set these three classes?

StupidCodeFactory: you don't really need to bother yourself with this. This is how rails build understand dates HTTP params. We have some really dodgy custom codes to do what rails already does, it's battle-tested.

edwardhorsford: The bit I was curious about was us explicitly (setting?) / hardcoding classes on each input - which feels like something that the component should be taking care of.

psd-web/app/views/investigations/emails/context.html.slim
```diff
@@ -28,6 +29,7 @@
               hint: { text: "Consumer contact details are protected by GDPR legislation" },
               items: [{ text: "No", value: "false" }, { text: "Yes", value: "true" }]
 
-      = render "form_components/govuk_date_input", form: form, key: :correspondence_date,
-              fieldset: { legend: { classes: "govuk-fieldset__legend--m", text: "Date sent" } }
+      = render "components/govuk_date_input", form: form, key: :correspondence_date, id: 'asdas',
```
edwardhorsford: What is this id?

psd-web/app/helpers/application_helper.rb
StupidCodeFactory: These are the 3 inputs for dates, GovUK component is not smart enough to figure out. 
This PR is a quick fix so that I can get going on the search ticket. 
We could add a ticket to make the design frontend gem a bit smarted. But it's out of scope for my story.

StupidCodeFactory: Ultimately, yes the component would need to be a bit smarter

psd-web/app/views/investigations/emails/context.html.slim
StupidCodeFactory: It is a draft, PR. so not cleaned up yet. will clean up once this PR is finished!

edwardhorsford: oops sorry my bad.

PR #121, opened 2019-12-30
Better field widths

edwardhorsford:
Adds a new set of classes for setting field widths and widths on autocompletes.

## Why do this?
The standard govuk classes for widths are % based - so they change in size depending on the screen size and columns in use. We want to size our inputs appropriately to the content they contain - but the widths are dramatically different depending on screen size - so we end up using widths much larger than needed in many cases.

At GDS a few years ago we discussed this and generally agreed to move to fixed width inputs - but the work never happened to make the change.

GOV.UK does have _some_ fixed width classes - but they apply even on mobile, where we want things full width. I've also found the range hard to remember / work with - I find working with fractions much easier (as well as making us more consistent). 

## Approach

I've done them as a series of modifiers that set a max width on the field - but only at tablet breakpoint and higher to maintain consistency with the GOV.UK style of making things full width on mobile.

Also includes a set of autocomplete specific width modifiers - as we can't directly set the input width of the autocomplete. The current autocomplete js doesn't copy classes from the source select so if we did apply a width class it wouldn't copy - instead we have to style based on a wrapper class.

## Applying the changes

I've only applied them to the product form - so that we can agree the approach and test in various browsers. I'd suggest if we're happy we then merge this and start using the classes as we go forward. The changes are minimal so they can be used alongside the existing widths.

## Before:
Desktop:
2019 12 30-11-16-35-Add product to case - Product safety database - GOV UK
Lower bound of desktop breakpoint:
2019 12 30-11-17-33-Add product to case - Product safety database - GOV UK
Mobile:
2019 12 30-11-19-48-Add product to case - Product safety database - GOV UK

## After:
Desktop:
2019 12 30-11-25-23-Add product to case - Product safety database - GOV UK

Lower bound of desktop breakpoint:
2019 12 30-11-25-16-Add product to case - Product safety database - GOV UK

Mobile:
2019 12 30-11-24-02-Add product to case - Product safety database - GOV UK

## Testing

It would be good to test this in a range of browsers. It's only ever setting max width so should be pretty safe.

## Questions

I debated over calling this `.app-!-max-width-two-thirds` vs `.app-!-input--max-width-two-thirds` or `.app-!-field--max-width-two-thirds` as they're mostly meant for inputs and _input-like_ things, and while they modify the tablet and desktop views, they do not apply to the mobile breakpoint.


nasirkhanbeis:
Tested in various browsers and mobile screens field widths looking good as expected.
the only persisting issue which I still see is white space in IE 11 browser in windows 7
image

psd-web/app/assets/stylesheets/helpers/field-widths.scss
```diff
@@ -0,0 +1,74 @@
...
+
+}
+
+// Custom width overrides so inputs can be fixed width regardless of resolution.
+@include govuk-media-query($from: tablet) {
+
+  // Widths chosen by eye to correspond to govuk widths at desktop 960px.
+
+  .app-\!-max-width-three-quarters {
+    max-width: 50ex !important;
```
frankieroberto: I must admit, I had to look up the `ex` unit to see how it differed from `em` or `px`. As I understand it, it relates more directly to the actual current font than `em` (ie it would change if you specified a monospace font, whereas `em` only relates to the `font-size`). Is that why you’re using it? If so, it might be worth a brief note for the sake of anyone else not familiar with the unit.

edwardhorsford:
@frankieroberto I've moved the autocomplete stuff as suggested.

We're using `ex` as this matches what govuk fixed width inputs use - so the field should scale with font size. On the autocomplete this means setting a font size on the container explicitly - as we don't know the size of the input itself.

PR #123, opened 2019-12-31
Fix back links on case attachment pages

slorek:
## Description
This is just a crude sticky plaster solution to fix the back links on the add/edit/remove attachment pages for a case, so that users don't see error messages. It should also resolve some exceptions we're seeing in Sentry due to the links routing to the wrong controller (which is part of the creation flow).

The changed page templates are shared for investigations/businesses/products, but there are big differences in how these three view template groups and routes work.

Businesses/products do not implement proper routes i.e. /products/:id/attachments, instead they implement a Javascript tab only, so the back links go to the overview page instead. It would be possible to add a further sticky plaster for investigations to go to `/attachments`, but I think if we want to go further we should just fix the crazy routing design.

The design really is crazy: the model association name for attachments is `documents`, but the model name for a document is resolved to `attachment` by the polymorphic route helpers. The route name is `documents`, so it is not resolved correctly, meaning it has to be overridden using [custom helpers](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/helpers/url_helper.rb).

Except the route name for *cases* is `attachments` on the `index` action only; member actions are called documents.  But the index action name on products/businesses is `documents`.

All of this means we have some crude URL helpers to maintain, which I've modified here.

There's plenty of other WTF in the attachments routing and behaviour, so to improve further we should consider rewriting this. The scope of this change is just to get the links working for users.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/spec/features/add_attachment_spec.rb
+++ b/psd-web/spec/features/add_attachment_spec.rb
@@ -69,13 +69,13 @@ RSpec.feature "Adding an attachment to a case", :with_stubbed_elasticsearch, :wi
   def expect_to_be_on_add_attachment_page
     expect(page).to have_current_path("/cases/#{investigation.pretty_id}/documents/new/upload")
     expect(page).to have_selector("h1", text: "Add attachment")
-    expect(page).to have_link("Back", href: "/cases/#{investigation.pretty_id}")
+    expect(page).to have_link("Back", href: investigation_path(investigation))
   end
 
   def expect_to_be_on_enter_attachment_details_page
     expect(page).to have_current_path("/cases/#{investigation.pretty_id}/documents/new/metadata")
     expect(page).to have_selector("h3", text: "Document details")
-    expect(page).to have_link("Back", href: "/cases/#{investigation.pretty_id}")
+    expect(page).to have_link("Back", href: investigation_path(investigation))
   end
 
   def expect_to_be_on_case_overview_page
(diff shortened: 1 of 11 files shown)


psd-web/spec/features/add_attachment_spec.rb
```diff
@@ -69,11 +69,13 @@
   def expect_to_be_on_add_attachment_page
     expect(page).to have_current_path("/cases/#{investigation.pretty_id}/documents/new/upload")
     expect(page).to have_selector("h1", text: "Add attachment")
+    expect(page).to have_link("Back", href: "/cases/#{investigation.pretty_id}")
```
StupidCodeFactory: Didn't notice this before, but I'd prefer to use the routes helper. `investigation_path(investigation)` and so on.

nasirkhanbeis:
@slorek can't test back link on edit/remove attachment pages as I see processing image label but not the actual image
image

edwardhorsford:
I created my own project and uploaded a file. When viewing the attachments tab, the edit / view / remove links are not present - I don't know if this is related to the image being stuck in processing or not.

Screenshot 2020-01-02 at 12 28 33

slorek:
Missing edit/remove links on review app were resolved by https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/125

PR #124, opened 2019-12-31
Fix Keycloak client test isolation

slorek:
## Description
The `KeycloakClient` singleton class was designed in such a way as state was persisted between tests, and it would also make a request to the OpenID configuration endpoint immediately upon instantiation, which wasn't always necessary.

This manifested in [errors](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/123/checks?check_run_id=369212250) depending on the spec suite run order if the Keycloak config wasn't stubbed correctly. Sometimes it would give a false positive if a previous test had already stubbed it. There were two model tests which were not stubbing the Keycloak config correctly.

It also meant that you had to stub the endpoint just to create a user with a factory, even if `KeycloakClient` was never called again.

This change ensures that the `KeycloakClient` state can be reset after each test, and lazily calls the configuration endpoint. This meant I could remove some unnecessary stubbing in some tests and it also yielded a small performance improvement of about 4 seconds on my machine.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/app/clients/keycloak_client.rb
+++ b/psd-web/app/clients/keycloak_client.rb
@@ -62,19 +62,24 @@ class KeycloakClient
   include Singleton
 
   def client
-    return @client if @client
-
-    @client = Keycloak::Client
-    @client.get_installation
-    @client
+    @client ||= begin
+      init_clients
+      @client
+    end
   end
 
   def admin
-    @admin ||= Keycloak::Admin
+    @admin ||= begin
+      init_clients
+      @admin
+    end
   end
 
   def internal
-    @internal ||= Keycloak::Internal
+    @internal ||= begin
+      init_clients
+      @internal
+    end
   end
 
   def reset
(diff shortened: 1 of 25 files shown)


psd-web/app/clients/keycloak_client.rb
```diff
@@ -61,18 +61,29 @@ def self.get_users(query_parameters = nil)
 class KeycloakClient
   include Singleton
 
-  def initialize
-    # The gem we're using has its api split across these three classes
+  def client
+    return @client if @client
```
StupidCodeFactory: following pattern is usually preferred:
```ruby
def client
  @client ||= begin
    c = Keycloak::Client
    c.get_installation
    c
  end
end
```

slorek:
There's a startup problem deploying this to PaaS: do not merge for now.

PR #129, opened 2020-01-03
Limit products shown on overview page

edwardhorsford:
We have a few examples of cases with many products. As they're shown on the right hand side of cases, they can make the page very long.

This PR adds a limit so if there's more than the limit, we don't show any further products and add a link to the products tab. I've tentatively set the limit at 6 products.

For the majority of cases, we should see no change.

The limit is pretty arbitrarily chosen - @LMallows may want to adjust. I think somewhere between 5-10 would be good.

At @frankieroberto's suggestion, it has a bit of logic to adjust the limit upwards so we don't end up with just 1 product not shown. So if there's 7 products we won't limit. If there's 8 it will limit to 6.

I hope that someone can help with adding tests for this.

Example (limit adjusted to 2 to ease screenshotting):
Screenshot 2020-01-03 at 17 22 25


psd-web/app/views/investigations/tabs/_overview.html.slim
```diff
@@ -1,5 +1,6 @@
 - page_heading = "Overview"
 - content_for :page_title, "#{page_heading} - #{@investigation.pretty_id}"
+- products_count = @investigation.products.count
```
StupidCodeFactory:
```suggestion
- products_count = @investigation.products.size
```
this will avoid an extra db query.

psd-web/app/views/investigations/tabs/_overview.html.slim
```diff
@@ -34,9 +35,14 @@
       = tag(:hr,
           class: "govuk-section-break govuk-section-break--m"\
           " govuk-section-break--visible app-!-display-mobile--only")
-      h2.govuk-heading-m = "Products (#{@investigation.products.count})"
+      h2.govuk-heading-m = "Products (#{products_count})"
+      - products_limit = 6
+      / Increase limit if we're within 1 product
+      - if ( products_count - products_limit == 1 )
+        - products_limit += 1
+      - products_remaining = products_count - products_limit
```
StupidCodeFactory: Let me re-write this logic in the presenter, templates should avoid at all cost logic.

edwardhorsford:
I should note that the order of the products isn't explicitly set. I _assume_ it's the order they were created, oldest first. I think it should probably be newest first, but we'd want to adjust elsewhere too - so we can leave as is for now.

psd-web/spec/decorators/investigation_decorator_spec.rb
```diff
@@ -297,4 +297,56 @@
...
+    context "with 7 images" do
+      let(:product_count) { 7 }
+
+      it "list all the products" do
+        products.each do |product|
+          expect(products_list).to have_link(product.name, href: product_path(product))
+        end
+      end
+
+      it "does not displays a link to see all the products" do
```
slorek: *display

psd-web/spec/decorators/investigation_decorator_spec.rb
```diff
@@ -297,4 +297,56 @@
...
+    context "with 6 images or less" do
+      let(:product_count) { 6 }
+
+      it "lists all the images" do
+        products.each do |product|
+          expect(products_list).to have_link(product.name, href: product_path(product))
+        end
+      end
+
+      it "does not displays a link to see all attached images" do
```
slorek: *display

psd-web/app/decorators/investigation_decorator.rb
```diff
@@ -2,6 +2,8 @@ class InvestigationDecorator < ApplicationDecorator
   delegate_all
   decorates_associations :documents_attachments
 
+  PRODUCT_IMAGE_DISPLAY_LIMIT = 6
```
slorek: Suggest renaming this as it doesn't seem to limit images only, but how many products are displayed.

nasirkhanbeis:
Tested this in review app, all good as expected. @StupidCodeFactory FYI

PR #131, opened 2020-01-06
Fix export

slorek:
## Description
Following the adding of the missing database indexes, we are still getting statement timeouts, albeit the query can be run successfully with one less association eager loaded.

I discovered that a lot of the data being pulled in the controller wasn't actually being rendered in the template; we're just displaying the counts of most associated records. This change makes the database queries much more efficient and should hopefully mean that the export works in production.

I've tested and compared the data output before/after the change is the same using the seed data.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/app/controllers/investigations_controller.rb
+++ b/psd-web/app/controllers/investigations_controller.rb
@@ -19,13 +19,14 @@ class InvestigationsController < ApplicationController
       format.xlsx do
         @answer = search_for_investigations
         @investigations = Investigation.eager_load(:complainant,
-                                                   :source,
-                                                   { products: :source },
-                                                   { activities: :source },
-                                                   { businesses: %i[locations source] },
-                                                   :corrective_actions,
-                                                   :correspondences,
-                                                   :tests).where(id: @answer.results.map(&:_id))
+                                                   :source).where(id: @answer.results.map(&:_id))
+
+        @activity_counts = Activity.group(:investigation_id).count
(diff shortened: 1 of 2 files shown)


psd-web/app/controllers/investigations_controller.rb
```diff
@@ -19,13 +19,14 @@ def index
...
-                                                   :correspondences,
-                                                   :tests).where(id: @answer.results.map(&:_id))
+                                                   :source).where(id: @answer.results.map(&:_id))
+
+        @activity_counts = Activity.group(:investigation_id).count
+        @business_counts = InvestigationBusiness.unscoped.group(:investigation_id).count
+        @product_counts = InvestigationProduct.unscoped.group(:investigation_id).count
+        @corrective_action_counts = CorrectiveAction.group(:investigation_id).count
+        @correspondence_counts = Correspondence.group(:investigation_id).count
+        @test_counts = Test.group(:investigation_id).count
```
StupidCodeFactory: seems like we'd want a counter cache on the relevant products, which would remove all of these queries.

StupidCodeFactory: By products I meant `associations` obviously

slorek: I'm wondering if this is worthwhile if we're only using it here, with the risk that the counts could go out of sync without us noticing.

PR #135, opened 2020-01-10
Fix pagination

StupidCodeFactory:
Fix missing pagination on Business and Product listing pages

https://trello.com/c/1WWJQVmN/266-can-only-view-10-products-on-the-product-page-and-10-businesses-on-the-business-page

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/app/helpers/businesses_helper.rb
+++ b/psd-web/app/helpers/businesses_helper.rb
@@ -8,17 +8,26 @@ module BusinessesHelper
   end
 
   def search_for_businesses(page_size)
-    Business.full_search(search_query)
-      .records
-      .paginate(page: params[:page], per_page: page_size)
+    BusinessDecorator.decorate_collection(
+      Business.full_search(search_query)
+        .page(params[:page])
+        .per_page(page_size)
+        .records
+    )
+  end
+
+  def sorting_params
+    return {} if params[:sort] == "relevance"
+
+    { created_at: :desc }
   end
 
   def sort_column
-    Business.column_names.include?(params[:sort]) ? params[:sort] : "legal_name"
+    Business.column_names.include?(params[:sort]) ? params[:sort] : "created_at"
   end
 
   def sort_direction
-    %w[asc desc].include?(params[:direction]) ? params[:direction] : "asc"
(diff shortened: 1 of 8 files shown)


edwardhorsford:
Laurence and I tried it out and get pagination when there's over 20 products - looks good.

edwardhorsford:
For clarity, I think this also includes two small feature changes:
* Sort products by date added, newest first.
* Sort businesses by date added, newest first.

psd-web/app/helpers/businesses_helper.rb
```diff
@@ -8,17 +8,26 @@ def defaults_on_primary_location(business)
...
+        .page(params[:page])
+        .per_page(page_size)
+        .records
+    )
+  end
+
+  def sorting_params
+    return {} if params[:sort] == "relevance"
+
+    { created_at: :desc }
```
frankieroberto: Might be slightly clearer as

```ruby
if params[:sort] == "relevance"
  {}
else
  { created_at: :desc }
end
```

psd-web/spec/features/busisnesses_spec.rb
```diff
@@ -0,0 +1,16 @@
+require "rails_helper"
+
+RSpec.describe "Busisnesses listing", :with_elasticsearch, :with_stubbed_mailer, :with_stubbed_keycloak_config do
+  let(:user) { create :user, :activated }
+
+  before { create_list :business, 21, created_at: 4.days.ago }
+
+  scenario "lists products according to search relevance" do
```
frankieroberto: This scenario seems to be testing that the pagination links appear, rather than that they appear in relevance order?

psd-web/app/helpers/businesses_helper.rb
slawosz: I think its default in Robocop, @frankieroberto

frankieroberto: @slawosz really? which rule? I'd be surprised if Rubocop prefers explicit return statements by default.

PR #136, opened 2020-01-10
Improve styling and accessibility for pagination

edwardhorsford:
Changes:

* Hide text 'previous' when you're on the first page
* Hide text 'next' when you're on the last page
* Improve spacing of the elements
* Prepend the links with "Page " using css so AT get clearer link text
* Put the navigation inside a nav element with aria-label (this markup matches what [govuk uses](https://www.gov.uk/search/all?keywords=support&order=relevance))

Before:
Screenshot 2020-01-10 at 18 01 18

After:
Screenshot 2020-01-10 at 17 59 00

I've tested the links in Voiceover on my mac. Previously when tabbing on a link you would get:
"Visited, link 2"

Now you get:
"Visited, link, Page 2, Pagination, navigation"

Longer term it would be good to render better link text and have more control over the markup - but in the short term this should improve things.


psd-web/app/assets/stylesheets/pagination.scss
```diff
@@ -1,7 +1,48 @@
...
+
+  // will_paginate shows unnecessary text 'previous' and 'next'
+  // even if you're on the first or last page.
+
+  // sass-lint:disable class-name-format
+  .previous_page.disabled,
+  .next_page.disabled {
+    display: none;
+  }
+  // sass-lint:enable class-name-format
```
frankieroberto: This sucks 😞 

In theory, it should be possible to alter the generated HTML by writing a [custom renderer](https://github.com/mislav/will_paginate/wiki/Link-renderer) but that doesn’t look super simple. To be honest, I’d be tempted to ditch the `will_paginate` gem and replace it with something simpler within the app...

edwardhorsford: Yep, I know. I think we could consider this a stopgap to make the links more accessible in the short term, with a longer term aim to add our own link renderer or replace `will_paginate` entirely.

edwardhorsford:
Suggest waiting until #141 gets merged - the release notes for that suggest it will fix several of the accessibility issues present in the default `will_paginate` view.

If that tests well, I can make this pr just about the styling changes.

edwardhorsford:
@frankieroberto could I get another review? I've removed the accessibility hacks as the updated `will_paginate` now has them built in.

PR #161, opened 2020-01-15
Remove redundant active_hash gem

slorek:
## Description
This gem is no longer used as of #56 .

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


slorek:
> you forgot to re-run bundle install there should be a diff for Gemfile.lock

Forgot to commit the file, good catch.

PR #190, opened 2020-01-20
Render out image rotation rather than using exif

edwardhorsford:
We've started noticing cases with images with the wrong rotation.

After some investigation it seems to be [related to this issue](https://stackoverflow.com/questions/42401203/chrome-image-exif-orienation-issue), where browsers (Chrome and Safari at least) don't respect the exif image orientation. This will tend to effect images taken with smartphones which are more likely to use this attribute.

As we're using ImageMagic to make previews, we can pass an option to fix this, effectively to ['render out' the rotation](https://imagemagick.org/script/command-line-options.php#auto-orient) in our thumbnails.

Before:
Screenshot 2020-01-20 at 15 22 45

After:
Screenshot 2020-01-20 at 15 23 17


edwardhorsford:
Sample image that should exhibit the issue:
IMG_6310

PR #191, opened 2020-01-20
Fix Sidekiq web not working with latest Rack versions

slorek:
## Description
This fixes the `/sidekiq` web interface not working with the latest `rack` versions.

2.1.1 produces `undefined method 'transform_keys' for #<ActionDispatch::Request::Session:0x7faef06f6660 not yet loaded>
` as per https://github.com/rack/rack/pull/1428

2.08 produces a `RuntimeError` as per https://github.com/rack/rack/issues/1432

Fixes appear to be tied to `rails` version bumps so for now we need to peg our release to 2.0.7.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


slorek:
Pegging `rack` to the merged https://github.com/rack/rack/pull/1428 until such time as a new release is issued.

PR #199, opened 2020-01-21
fix search

StupidCodeFactory:
not very happy with how this code looks. But not really possible to do otherwise as the refactoring the search/code/form would take days.
So this is really just another hack on top of a tone of hacks.
## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/app/controllers/searches_controller.rb
+++ b/psd-web/app/controllers/searches_controller.rb
@@ -2,7 +2,16 @@ class SearchesController < ApplicationController
   include InvestigationsHelper
 
   def show
-    set_search_params
+    params_to_save = params.dup
+    params_to_save.delete(:sort_by) if params[:sort_by] == SearchParams::RELEVANT
+
+    if params[:override_sort_by]
+      params[:sort_by] = params[:override_sort_by]
+    end
+
+    @search = SearchParams.new(query_params)
+    session[:previous_search_params] = params_to_save
+
     if @search.q.blank?
       redirect_to investigations_path(previous_search_params)
     else
--- a/psd-web/app/views/investigations/_filters.html.slim
+++ b/psd-web/app/views/investigations/_filters.html.slim
@@ -36,5 +36,4 @@
               fieldset: { legend: { text: "Sort by", classes: "govuk-fieldset__legend--s" } },
               items: search.sort_by_items(with_relevant_option: show_relevancy_radion_option)
 
-      = form.hidden_field :submitted_from, value: submitted_from
       = form.submit "Apply filters", name: nil, class: "govuk-button"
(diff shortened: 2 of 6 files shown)


psd-web/app/controllers/searches_controller.rb
```diff
@@ -2,7 +2,16 @@ class SearchesController < ApplicationController
   include InvestigationsHelper
 
   def show
-    set_search_params
+    params_to_save = params.dup
+    params_to_save.delete(:sort_by) if params[:sort_by] == SearchParams::RELEVANT
+
+    if params[:override_sort_by]
```
slorek: Can this bit not be added to the `set_search_params` method as it otherwise looks the same?

psd-web/app/views/investigations/index.html.slim
```diff
@@ -2,12 +2,13 @@
...
 .govuk-grid-row
-  = render 'investigations/filters', search: @search, show_relevancy_radion_option: false
+  = render 'investigations/filters', search: @search,\
+      show_relevancy_radion_option: false, submitted_from: "investigation_listing"
 
   .govuk-grid-column-three-quarters
     - if @investigations.any?
       = render partial: "investigations/table", collection: @investigations, as: :investigation,\
-         locals: { sorted_by: query_params[:sort_by] }
+         locals: { sorted_by: query_params[:sort_by]}
```
slorek: This space should probably be retained (surprised Rubocop didn't pick it up).

psd-web/app/views/investigations/_filters.html.slim
```diff
@@ -28,12 +28,13 @@
...
+
       = render "form_components/govuk_radios", form: form, key: :sort_by,
               classes: "govuk-radios--small",
               fieldset: { legend: { text: "Sort by", classes: "govuk-fieldset__legend--s" } },
               items: search.sort_by_items(with_relevant_option: show_relevancy_radion_option)
 
-      - unless show_relevancy_radion_option
-        = form.hidden_field :sort_by, value: SearchParams::RELEVANT
-
+      = form.hidden_field :submitted_from, value: submitted_from
```
slorek: This input seems redundant.

psd-web/app/controllers/searches_controller.rb
StupidCodeFactory: I'm not massively convinced I'd want to share the code between  the two controller. This really is destined to this controller.
This code is far from ideal but granted the state of the rest of the search related code randomly scattered across a few modules, I would want to add more **abstraction** to somethings that's already a bit mental to grasp.

PR #217, opened 2020-01-24
Devise - Omniauth - Keycloak

StupidCodeFactory:
This integrates with our existing Keycloak authentication service using Devise with Omniauth (and the OpenId protocol) instead of using a custom `KeycloakClient` class.

This is a stepping stone towards removing Keycloak by using Devise with the authentication credentials stored within the app.

**IMPORTANT:**
before merging we need to add this environment variable:
KEYCLOAK_REDIRECT_URI=https://$DOMAIN/users/auth/openid_connect/callback


psd-web/app/interactors/users/create_session.rb
```diff
@@ -0,0 +1,7 @@
+module Users
+  class CreateSession
+    include Interactor::Organizer
+
+    organize Load
```
StupidCodeFactory: I've created this organiser, so that we can will later add the sending of the 2FA interactor

psd-web/test/test_helper.rb
```diff
@@ -16,16 +16,17 @@
 
 # Added Webmock only to allow use of stub_request - Minitest suite is deprecated
 require "webmock/minitest"
+require_relative "test_helpers/devise"
+
+if ENV['CHROMEDRIVER_PATH']
+  Selenium::WebDriver::Chrome::Service.driver_path = ENV.fetch('CHROMEDRIVER_PATH')
```
StupidCodeFactory: maybe just add chromedriver to **your** path

frankieroberto: It is in my path, but for some reason that I've not figured out yet, I still need to explicitly tell Selenium where to find it. I can remove this now and just keep it stashed locally.

.github/workflows/minitest.yml
```diff
@@ -85,9 +91,10 @@ jobs:
...
-        KEYCLOAK_CLIENT_ID: ${{ secrets.KeycloakClientId }}
-        KEYCLOAK_CLIENT_SECRET: ${{ secrets.KeycloakClientSecret }}
-        KEYCLOAK_USER_PASSWORD: ${{ secrets.KeycloakUserPassword }}
       run: bundle exec rake test:without_system_tests
+
+    - uses: actions/upload-artifact@v1
+      if: failure()
+      with:
+        name: capybara screenshots
+        path: psd-web/screenshots
```
frankieroberto: I didn't think these tests would generate a screenshot (as they're not including the browser ones)?

StupidCodeFactory: they do generate screenshots though (basically also  html pages)

psd-web/app/controllers/sessions_controller.rb
```diff
@@ -1,41 +1,6 @@
...
-  end
-
-  def is_relative(url)
-    url =~ /^\/[^\/\\]/
+class SessionsController < Devise::SessionsController
+  def destroy
+    # ::KeycloakClient.instance.logout(refresh_token)
+    super
   end
 end
```
frankieroberto: Looks like we're not adding any behaviour to the parent class here – could we route to the `Devise::SessionsController` instead?

.github/workflows/minitest.yml
frankieroberto: ah, I'd assumed screenshots meant images of rendered pages, not HTML files...

psd-web/app/helpers/login_helper.rb
```diff
@@ -1,9 +1,5 @@
 module LoginHelper
-  def keycloak_login_url(request_path = nil)
-    KeycloakClient.instance.login_url(session_url_with_redirect(request_path))
-  end
-
-  def keycloak_registration_url(request_path = nil)
+  def keycloak_registration_urla(request_path = nil)
```
frankieroberto: is this a typo (adding the `a` to the method name)? Looks like this method isn't referenced anywhere – maybe it can be removed?

psd-web/app/controllers/sessions_controller.rb
StupidCodeFactory: we should log the user out to keep the previous behaviour. one of things left on my todo list

psd-web/app/helpers/login_helper.rb
StupidCodeFactory: ✂️

.github/workflows/minitest.yml
StupidCodeFactory: I didn't choose the path, it's the default. It's really not a massive, I could spend some time figuring out how to rename this folder, is it really worth it?

psd-web/app/models/user.rb
```diff
@@ -103,8 +104,9 @@ def team_names
     teams.map(&:name).join(", ")
   end
 
-  def assignee_short_name
-    if organisation.present? && organisation.id != User.current.organisation_id
+  # TODO: remove default value of User.current
+  def assignee_short_name(other_user: User.current)
```
StupidCodeFactory: @frankieroberto I'm going to clean this up and pass the current user in the template

frankieroberto: 👍

psd-web/spec/shared_examples/create_user.rb
```diff
@@ -0,0 +1,13 @@
+RSpec.shared_examples "creates a user for" do
```
slorek: "creates a user"

psd-web/spec/support/elasticsearch.rb
```diff
@@ -1,7 +1,21 @@
 RSpec.shared_context "with stubbed Elasticsearch", shared_context: :metadata do
   before do
     elasticsearch_url = ENV.fetch("ELASTICSEARCH_URL", "http://elasticsearch:9200")
-    stub_request(:any, /#{Regexp.quote(elasticsearch_url)}/).to_return(body: "Elasticsearch disabled in Rspec", status: 200)
+    empty_elasticsearch_response = {
```
slorek: This might not be intuitive...

psd-web/spec/support/keycloak.rb
```diff
@@ -7,6 +7,16 @@
...
+  around do |ex|
+    Singleton.__init__(KeycloakClient)
+    ex.run
+    Singleton.__init__(KeycloakClient)
+  end
+end
+
 RSpec.configure do |rspec|
   rspec.include_context "with stubbed Keycloak configuration", with_stubbed_keycloak_config: true
+  rspec.include_context "reset keycloak client", :reset_keycloak_client
```
slorek: This should probably just be default behaviour for every test.

psd-web/app/controllers/sessions_controller.rb
```diff
@@ -1,41 +1,6 @@
...
-  def auth_code
-    params.require(:code)
-  end
-
-  def is_relative(url)
-    url =~ /^\/[^\/\\]/
+class SessionsController < Devise::SessionsController
+  def destroy
+    # ::KeycloakClient.instance.logout(refresh_token)
+    super
```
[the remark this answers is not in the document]
StupidCodeFactory: @slorek  I was actually thinking to log the user out from keycloak but the effort doesn't seem worth it since we'll rip this off

psd-web/app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,30 @@
+class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
+  skip_before_action :has_accepted_declaration
+  skip_before_action :authorize_user
+
+  def openid_connect
+    if create_user.success?
+      sign_in_and_redirect(create_user.user)
+      set_flash_message(:notice, :success, kind: "Facebook")
```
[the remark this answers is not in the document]
StupidCodeFactory: the flash doesn't appear anyway i'll just ✂️ this.

psd-web/spec/support/elasticsearch.rb
StupidCodeFactory: That's an actual ES response. I had some really hard time debugging the "Elasticsearch disabled in Rspec" which would not appear anywhere has it's not JSON parseable and I i'd get some really funky behaviour.
I'd suggest we keep it that way it's `\/cases||\/products||\/businesses` expect to this shapre to work properly with the `ElasticSearch::ResultProxy`.

I'm open to suggestion though

psd-web/app/controllers/application_controller.rb
```diff
@@ -73,12 +81,20 @@ def previous_search_params
 
   def secondary_nav_items
     items = []
-    if User.current
+    if user_signed_in?
       items.push text: "Your account", href: KeycloakClient.instance.user_account_url
       items.push text: "Sign out", href: logout_session_path
     else
-      items.push text: "Sign in", href: keycloak_login_url
+      items.push text: "Sign in", href: user_openid_connect_omniauth_authorize_path
```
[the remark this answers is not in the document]
StupidCodeFactory: this will be killed when we have devise it don't really the point of wrapping this method into another `login_url` method which would not even appear when doing `rails routes` which would be confusing.

psd-web/app/controllers/application_controller.rb
```diff
@@ -73,12 +81,20 @@ def previous_search_params
...
     end
     items
   end
+
+  def after_sign_in_path_for(resource)
+    stored_location_for(resource) || investigations_path
```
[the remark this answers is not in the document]
StupidCodeFactory: ok I'll will have a look into that

psd-web/.env.development.example
```diff
@@ -1,7 +1,7 @@
 NOTIFY_API_KEY=
 NOTIFY_SMS_TEMPLATE_ID=
 
-DATABASE_URL=postgres://localhost/psd_dev?user=postgres
+DATABASE_URL=postgres://postgres@localhost:5432/
```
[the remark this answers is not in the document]
StupidCodeFactory: no it's on the database.yml

.github/workflows/minitest.yml
frankieroberto: No, let's leave this.

.github/workflows/system_tests.yml
```diff
@@ -3,13 +3,20 @@ name: System Tests
...
+  KEYCLOACK_REDIRECT_URI: http://localhost:3000/users/auth/openid_connect/callback
   DATABASE_URL: postgres://postgres@localhost:5432
   ELASTICSEARCH_URL: http://localhost:9200
   REDIS_URL: redis://localhost:6379
   REDIS_HOST: localhost
   RAILS_ENV: test
   PGHOST: localhost
   PGUSER: postgres
+  HTTP_HOST: localhost
+  HTTP_POST: 3000
```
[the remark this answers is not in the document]
frankieroberto: I wonder if we still need `HTTP_HOST` and `HTTP_POST` env vars, given that we've switched the url helpers to path helpers? Does something within the omniauth flow require a full url?

psd-web/test/test_helpers/devise.rb
```diff
@@ -0,0 +1,47 @@
...
+  module Devise
+    include TestHelpers::OmniAuthHelper
+
+    def sign_out
+      return if page.has_css?("a", text: "Sign in to your account")
+
+      click_on "Sign out", match: :first
+    end
+
+    def sign_in(user = users(:opss), roles: %i[psd_user opss_user])
```
frankieroberto: Can we remove the default user and roles here? Feels like the kind of thing that might confuse us later...

.github/workflows/system_tests.yml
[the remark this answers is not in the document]
StupidCodeFactory: @frankieroberto good question. I'll see if I can clean this up. My hunch is that due to something weird in our setup this is needed by Capybara. because its hosts and default url is www.example.com

psd-web/test/test_helpers/devise.rb
StupidCodeFactory: It's quite handy to not have to type it everywhere.
Minitests will go away and we won't be writing any anymore. I'd suggest we leave it as it is

psd-web/app/controllers/application_controller.rb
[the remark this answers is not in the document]
StupidCodeFactory: So looks like this test is not actually testing anything since it passes

psd-web/app/controllers/application_controller.rb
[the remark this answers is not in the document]
StupidCodeFactory: that's also the only login link reference

psd-web/app/controllers/application_controller.rb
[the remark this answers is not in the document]
slorek: it should definitely be `root_path`, see https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/101

psd-web/app/controllers/sessions_controller.rb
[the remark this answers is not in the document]
slorek: Right, I think we will need to ensure they are logged out if this goes to production.

psd-web/spec/support/elasticsearch.rb
slorek: I think we need some indication that it is a stubbed response and not just a real Elasticsearch producing no results. Perhaps a log message (that could be too noisy) or a fake result which is obviously a result of the stub?

psd-web/test/test_helpers/devise.rb
slorek: I don't mind this, but we have to be careful not to introduce any false positives in existing tests...

psd-web/app/controllers/application_controller.rb
```diff
@@ -10,36 +10,50 @@ class ApplicationController < ActionController::Base
   before_action :set_raven_context
   before_action :authorize_user
   before_action :has_accepted_declaration
+  before_action :has_not_view_introduction
```
slorek: Naming is different between these two. Prefer `has_viewed_introduction` (same as the `User#has_viewed_introduction`).

psd-web/app/interactors/users/create_session.rb
slawosz: I wonder how it will work with 2FA, as we are planning using lot of device stuff?

psd-web/spec/features/home_page_spec.rb
```diff
@@ -42,6 +42,7 @@ def expect_header_to_have_signed_in_links
 
       context "not previously accepted the declaration" do
         let(:has_accepted_declaration) { false }
+        let(:has_viewed_introduction) { false }
```
slorek: Has the factory for user changed? It should be `false` by default and used to be.

StupidCodeFactory: No the factory has not changed, it's set to `true` in the above context.

psd-web/spec/support/elasticsearch.rb
StupidCodeFactory: I'd have expected the `:with_stubbed_elasticsearch` metadata to be clear enough. 
A "fake" result would probably have some weird side effect, also I'm not sure what you mean by "fake" result. Agreed that the logging will be too noisy. 

I'll try to remove it all together but that might re-throw a 400 error. I spent an entire day trying to figure out what was going on.

psd-web/app/interactors/users/create_session.rb
StupidCodeFactory: That's a fair point, however we're going to extract more and more of the core logic into services classes orchestrated (buzz words... :D ) with interactors.

psd-web/spec/support/elasticsearch.rb
StupidCodeFactory:
```
     ActionView::Template::Error:
       undefined method `each_pair' for "Elasticsearch disabled in Rspec":String
       Did you mean?  each_char
     # /Users/yann/.gem/ruby/2.5.7/gems/hashie-3.6.0/lib/hashie/mash.rb:210:in `deep_update'
     # /Users/yann/.gem/ruby/2.5.7/gems/hashie-3.6.0/lib/hashie/mash.rb:124:in `initialize'
     # /Users/yann/.gem/ruby/2.5.7/gems/elasticsearch-model-6.0.0/lib/elasticsearch/model/response.rb:30:in `new'
     # /Users/yann/.gem/ruby/2.5.7/gems/elasticsearch-model-6.0.0/lib/elasticsearch/model/response.rb:30:in `response'
     # /Users/yann/.gem/ruby/2.5.7/gems/elasticsearch-model-6.0.0/lib/elasticsearch/model/response/records.rb:37:in `ids'
     # /Users/yann/.gem/ruby/2.5.7/gems/elasticsearch-model-6.0.0/lib/elasticsearch/model/adapters/active_record.rb:22:in `records'
     # /Users/yann/.gem/ruby/2.5.7/gems/elasticsearch-model-6.0.0/lib/elasticsearch/model/response/records.rb:13:in `each'
     # /Users/yann/.gem/ruby/2.5.7/gems/draper-3.1.0/lib/draper/collection_decorator.rb:43:in `map'
     # /Users/yann/.gem/ruby/2.5.7/gems/draper-3.1.0/lib/draper/collection_decorator.rb:43:in `decorated_collection'
     # /Users/yann/.gem/ruby/2.5.7/gems/draper-3.1.0/lib/draper/delegation.rb:10:in `any?'
     # ./app/views/investigations/index.html.slim:9:in `_app_views_investigations_index_html_slim___1924593218742029828_70276346109640'
     # /Users/yann/.gem/ruby/2.5.7/gems/sentry-raven-2.13.0/lib/raven/integrations/rails/controller_transaction.rb:7:in `block in included'
     # /Users/yann/.gem/ruby/2.5.7/gems/omniauth-1.9.0/lib/omniauth/strategy.rb:420:in `call_app!'
     # /Users/yann/.gem/ruby/2.5.7/gems/omniauth-1.9.0/lib/omniauth/strategy.rb:274:in `mock_call!'
     # /Users/yann/.gem/ruby/2.5.7/gems/omniauth-1.9.0/lib/omniauth/strategy.rb:186:in `call!'
     # /Users/yann/.gem/ruby/2.5.7/gems/omniauth-1.9.0/lib/omniauth/strategy.rb:169:in `call'
     # /Users/yann/.gem/ruby/2.5.7/gems/scout_apm-2.6.6/lib/scout_apm/middleware.rb:20:in `call'
     # /Users/yann/.gem/ruby/2.5.7/gems/warden-1.2.8/lib/warden/manager.rb:36:in `block in call'
     # /Users/yann/.gem/ruby/2.5.7/gems/warden-1.2.8/lib/warden/manager.rb:34:in `catch'
     # /Users/yann/.gem/ruby/2.5.7/gems/warden-1.2.8/lib/warden/manager.rb:34:in `call'
     # /Users/yann/.gem/ruby/2.5.7/gems/rack-2.2.1/lib/rack/tempfile_reaper.rb:15:in `call'
     # /Users/yann/.gem/ruby/2.5.7/gems/rack-2.2.1/lib/rack/etag.rb:27:in `call'
     # /Users/yann/.gem/ruby/2.5.7/gems/rack-2.2.1/lib/rack/conditional_get.rb:27:in `call'
     # /Users/yann/.gem/ruby/2.5.7/gems/rack-2.2.1/lib/rack/head.rb:12:in `call'
     # /Users/yann/.gem/ruby/2.5.7/gems/rack-2.2.1/lib/rack/session/abstract/id.rb:265:in `context'
     # /Users/yann/.gem/ruby/2.5.7/gems/rack-2.2.1/lib/rack/session/abstract/id.rb:259:in `call'
     # /Users/yann/.gem/ruby/2.5.7/gems/railties-5.2.3/lib/rails/rack/logger.rb:38:in `call_app'
     # /Users/yann/.gem/ruby/2.5.7/gems/railties-5.2.3/lib/rails/rack/logger.rb:26:in `block in call'
     # /Users/yann/.gem/ruby/2.5.7/gems/railties-5.2.3/lib/rails/rack/logger.rb:26:in `call'
     # /Users/yann/.gem/ruby/2.5.7/gems/request_store-1.5.0/lib/request_store/middleware.rb:19:in `call'
     # /Users/yann/.gem/ruby/2.5.7/gems/rack-2.2.1/lib/rack/method_override.rb:24:in `call'
     # /Users/yann/.gem/ruby/2.5.7/gems/rack-2.2.1/lib/rack/runtime.rb:22:in `call'
     # /Users/yann/.gem/ruby/2.5.7/gems/rack-2.2.1/lib/rack/sendfile.rb:110:in `call'
     # /Users/yann/.gem/ruby/2.5.7/gems/sentry-raven-2.13.0/lib/raven/integrations/rack.rb:51:in `call'
     # /Users/yann/.gem/ruby/2.5.7/gems/railties-5.2.3/lib/rails/engine.rb:524:in `call'
     # /Users/yann/.gem/ruby/2.5.7/gems/rack-2.2.1/lib/rack/urlmap.rb:74:in `block in call'
     # /Users/yann/.gem/ruby/2.5.7/gems/rack-2.2.1/lib/rack/urlmap.rb:58:in `each'
     # /Users/yann/.gem/ruby/2.5.7/gems/rack-2.2.1/lib/rack/urlmap.rb:58:in `call'
     # /Users/yann/.gem/ruby/2.5.7/gems/rack-test-1.1.0/lib/rack/mock_session.rb:29:in `request'
     # /Users/yann/.gem/ruby/2.5.7/gems/rack-test-1.1.0/lib/rack/test.rb:266:in `process_request'
     # /Users/yann/.gem/ruby/2.5.7/gems/rack-test-1.1.0/lib/rack/test.rb:129:in `custom_request'
     # /Users/yann/.gem/ruby/2.5.7/gems/rack-test-1.1.0/lib/rack/test.rb:58:in `get'
     # /Users/yann/.gem/ruby/2.5.7/gems/capybara-3.31.0/lib/capybara/rack_test/browser.rb:68:in `process'
     # /Users/yann/.gem/ruby/2.5.7/gems/capybara-3.31.0/lib/capybara/rack_test/browser.rb:52:in `block in process_and_follow_redirects'
     # /Users/yann/.gem/ruby/2.5.7/gems/capybara-3.31.0/lib/capybara/rack_test/browser.rb:47:in `times'
     # /Users/yann/.gem/ruby/2.5.7/gems/capybara-3.31.0/lib/capybara/rack_test/browser.rb:47:in `process_and_follow_redirects'
     # /Users/yann/.gem/ruby/2.5.7/gems/capybara-3.31.0/lib/capybara/rack_test/browser.rb:23:in `visit'
     # /Users/yann/.gem/ruby/2.5.7/gems/capybara-3.31.0/lib/capybara/rack_test/driver.rb:45:in `visit'
     # /Users/yann/.gem/ruby/2.5.7/gems/capybara-3.31.0/lib/capybara/session.rb:278:in `visit'
     # /Users/yann/.gem/ruby/2.5.7/gems/capybara-3.31.0/lib/capybara/dsl.rb:58:in `block (2 levels) in <module:DSL>'
     # ./spec/support/login_helpers.rb:21:in `sign_in'
     # ./spec/features/send_alert_spec.rb:10:in `block (2 levels) in <top (required)>'
     # ./spec/support/keycloak.rb:14:in `block (2 levels) in <top (required)>'
     # ./spec/support/database_cleaner.rb:9:in `block (3 levels) in <top (required)>'
     # /Users/yann/.gem/ruby/2.5.7/gems/database_cleaner-1.8.2/lib/database_cleaner/generic/base.rb:16:in `cleaning'
     # /Users/yann/.gem/ruby/2.5.7/gems/database_cleaner-1.8.2/lib/database_cleaner/configuration.rb:87:in `block (2 levels) in cleaning'
     # /Users/yann/.gem/ruby/2.5.7/gems/database_cleaner-1.8.2/lib/database_cleaner/configuration.rb:88:in `cleaning'
     # ./spec/support/database_cleaner.rb:8:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # NoMethodError:
     #   undefined method `each_pair' for "Elasticsearch disabled in Rspec":String
     #   Did you mean?  each_char
     #   /Users/yann/.gem/ruby/2.5.7/gems/hashie-3.6.0/lib/hashie/mash.rb:210:in `deep_update'
```

StupidCodeFactory: @slorek that's error i get ^^ It's not fun to debug.

slorek: `undefined method `each_pair' for "Elasticsearch disabled in Rspec":String` is quite obvious IMO or you could raise an exception, but I'm not too fussed as long as there's some easy way to see that it's stubbed. Otherwise it could be a big time sink for someone in future 😂

psd-web/spec/features/add_attachment_spec.rb
```diff
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 RSpec.feature "Adding an attachment to a case", :with_stubbed_elasticsearch, :with_stubbed_antivirus, :with_stubbed_mailer, :with_stubbed_keycloak_config do
-  let(:user) { create(:user, :activated) }
+  let(:user) { create(:user, :activated, as_viewed_introduction: true) }
```
slorek:
```suggestion
  let(:user) { create(:user, :activated, has_viewed_introduction: true) }
```

slorek:
@StupidCodeFactory please add the new env var to the int, staging and prod environments (and add to the manifest.yml and manifest.review.yml if they're new services. You might just modify the existing `psd-keycloak-env`).

StupidCodeFactory:
> Tested this in review app and looks good to me.
> Did manual regression testing around user roles, case assign and over all exploratory testing around roles and permissions.
> Re-tested redirection and flash message- all good now.
> Also did concurrency load test with 10 users connecting at same time (ramp-up 5 second which means 5th second all 10 different users connected) - all green
> NB: atm sign-out link in service doesn't sign-out completely it seems as when I try login with different user it still shows as logged in with another users(which jmeter also reported). Laurence is okay with this atm as users in prod would not have more than one account which is obvious so 0% impact
> Session timeout cannot be tested as we tried to set 1 min for testing purpose but didn't worked.
> Anyway not a blocker. so all good!

FYI: @frankieroberto tested the timeout and it worked.

PR #226, opened 2020-01-28
Bump govuk-design-system-rails to fix checkbox aria attribute

slorek:
## Description
This pulls the latest design system gem update to address a bug with the `aria-describedby` attribute as described by https://github.com/UKGovernmentBEIS/govuk-design-system-rails/pull/14

Also resolved a `rubocop` gemfile conflict.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/Gemfile
+++ b/psd-web/Gemfile
@@ -39,7 +39,7 @@ gem "strong_migrations", "0.6.0"
 gem "webpacker", "4.2.2"
 gem "wicked"
 
-gem "govuk-design-system-rails", git: "https://github.com/UKGovernmentBEIS/govuk-design-system-rails", tag: "0.6.0", require: "govuk_design_system"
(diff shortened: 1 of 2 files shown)


psd-web/Gemfile
```diff
@@ -67,7 +67,7 @@ group :development, :test do
   gem "selenium-webdriver"
   gem "simplecov"
   gem "simplecov-console"
-  gem "slim_lint"
+  gem "slim_lint", "0.19.0"
```
StupidCodeFactory: what was the problem with newer version?

slorek: The `rubocop-govuk` gem version 2.0.0 is pegged to an older version of `rubocop` than `slim-lint` 0.20.0 accepts and so bundler cannot resolve the dependencies. I'm not show how it worked...

PR #228, opened 2020-01-28
Pin Ruby buildpack to v1.8.4

slorek:
## Description
This pins the buildpack version to fix deployments since the new buildpack is failing.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


StupidCodeFactory:
😱 flaky test - I will have a look first thing tomorrow

PR #237, opened 2020-01-30
Bugfix: sets yesterday’s date when today is the first of the month

frankieroberto:
Previously when the day was the first of the month, clicking "Yesterday" set the date as the 0th of the month. This fix correctly sets it to the last day of the previous month, by using the [`setDate()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setDate) function on the date object, which if you pass `0`, sets the date object to the last day of the previous month (and previous year, if necessary).

See https://trello.com/c/hpFZ3R5K/170-yesterday-link-doesnt-work-on-1st-of-the-month

### DIFF
--- a/psd-web/app/assets/javascripts/date_input.js
+++ b/psd-web/app/assets/javascripts/date_input.js
@@ -2,27 +2,31 @@ import $ from 'jquery';
 
 function dateInput(idPrefix) {
   $(document).ready(() => {
-    const dateObj = new Date();
-    const currentDay = dateObj.getUTCDate();
-    const currentMonth = dateObj.getUTCMonth() + 1;
-    const currentYear = dateObj.getUTCFullYear();
+    const currentDate = new Date();
+    var dateYesterday = new Date(currentDate.getTime())
+
+    // This sets the date to the previous day, since setting
+    // the date to 0 sets the date to the last day of the
+    // previous month.
+    dateYesterday.setDate(dateYesterday.getDate() - 1)
+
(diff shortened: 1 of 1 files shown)


psd-web/app/assets/javascripts/date_input.js
```diff
@@ -2,27 +2,31 @@ import $ from 'jquery';
 
 function dateInput(idPrefix) {
   $(document).ready(() => {
-    const dateObj = new Date();
-    const currentDay = dateObj.getUTCDate();
-    const currentMonth = dateObj.getUTCMonth() + 1;
-    const currentYear = dateObj.getUTCFullYear();
+    const currentDate = new Date();
+    var dateYesterday = new Date(currentDate.getTime())
```
StupidCodeFactory: So according to this yesterday is the current day. Have I been leaving in the past all my life 😆 ??

frankieroberto: 😂

frankieroberto:
@StupidCodeFactory I think we should avoid creating new system tests – in cosmetics we managed to convert them all to feature specs using rack-test, which is way faster.

At some point I'd be happy to convert this to a proper component using progressive-enhancement, and at that point we could unit test the javascript using [jest](https://jestjs.io).

edwardhorsford:
I'm not sure I can test this, short of waiting 2 days for the first of the month to roll by.

PR #243, opened 2020-01-30
Ignore db folder for codeclimate analysis

frankieroberto:
This folder contains migrations, the schema, and seed data. The files are often necessarily verbose, and there is little value in refactoring old migrations.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,2 +1,6 @@
 exclude_patterns:
 - "psd-web/db/"
+- "psd-web/test/"
+- "psd-web/spec/"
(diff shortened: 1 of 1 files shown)


slorek:
I’m guessing the introduction of the file has overridden some defaults around ignoring pre-existing code.

frankieroberto:
@slorek I've added some more folders to ignore. It still seems to find 2 "new" issues in app/models though (which obviously aren't new).

slorek:
I've disabled the status check being required, hopefully the introduction of the config file will only affect this PR and future ones will not dig up old dirt.

PR #245, opened 2020-01-31
WIP: Fixing test runner

frankieroberto:
This ensures that the ruby version can't get silently upgraded, and also includes the ruby version within the bundler cache, to avoid an issue whereby the ruby version is upgraded but the gems from a previous ruby version are re-used.


slorek:
We should pin the Ruby patch version in all the workflows (others are failing as well) and use the same cache key across all of them (some of them aren't specific enough).

frankieroberto:
Closing in favour of #250.

PR #259, opened 2020-02-04
Team page tests

nasirkhanbeis:
Added further additional tests around invite team member functionality

## Description
validate error message when invite existing user
validate normal users should not be able to invite team members
## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


psd-web/spec/features/team_spec.rb
```diff
@@ -33,13 +33,53 @@ def resend_link_selector(email)
...
       end
+      #Need to cover this scenario later when we integrate new auth devise
+
+      # scenario "I should be able to invite a team member" do
+      #   click_link "Invite a team member"
+      #   fill_in "new_user_email_address", with: invite_email
+      #   click_button "Send invitation email"
+      #   save_and_open_page
+      #   expect(page).to have_css(".hmcts-banner__message", text: "Invite sent to #{invite_email}")
+      # end
```
StupidCodeFactory: ✂️

psd-web/spec/features/team_spec.rb
```diff
@@ -33,13 +33,53 @@ def resend_link_selector(email)
...
+        expect(page).to have_css(".govuk-error-summary__list", text: "You cannot invite this person to join your team because they are already a member of another team from a different organisation.")
+      end
+    end
+  end
+
+  context "as a normal user" do
+    let(:user) { create(:user, :activated, :psd_user, teams: [team]) }
+
+    scenario "does not display the link for any users" do
+      expect(page).not_to have_css("a[href*=\"#{resend_invitation_team_path(team)}\"]")
```
StupidCodeFactory:
```suggestion
      expect(page).not_to have_link("Your team", href: resend_invitation_team_path(team))
```

psd-web/spec/features/team_spec.rb
```diff
@@ -33,13 +33,53 @@ def resend_link_selector(email)
...
+    end
+
+    scenario "does not display the invite a team member link" do
+      expect(page).not_to have_css("govuk-button", text: "Invite a team member")
```
StupidCodeFactory:
```suggestion
      expect(page).not_to have_button("Invite a team member")
```

psd-web/spec/features/team_spec.rb
StupidCodeFactory: this is not a very useful test, can be a false positive can you add an assertion so that it makes sure you're on the expected page and it's also useful for whom is going to read this test to know what page we're supposed to be on

psd-web/spec/features/team_spec.rb
StupidCodeFactory: same as my previous comment add an assertion that document what page we're on and prevent a false positive

psd-web/spec/features/team_spec.rb
```diff
@@ -33,13 +33,53 @@ def resend_link_selector(email)
...
       end
     end
+  end
+
 
-    context "as a normal user" do
-      let(:user) { create(:user, :activated, :psd_user, teams: [team]) }
+  context "As a team admin user" do
+    let(:email_whitelist_enabled) { false }
+    context "when I invite a new user" do
```
StupidCodeFactory: you can remove this entire block it's not doing anything

psd-web/spec/features/team_spec.rb
```diff
@@ -34,12 +34,27 @@ def resend_link_selector(email)
...
     end
 
-    context "as a normal user" do
-      let(:user) { create(:user, :activated, :psd_user, teams: [team]) }
+    context "when I invite an existing user" do
+      let(:user) { create(:user, :activated, :team_admin, teams: [team]) }
 
-      scenario "does not display the link for any users" do
-        expect(page).not_to have_css("a[href*=\"#{resend_invitation_team_path(team)}\"]")
+      scenario ",I should see an error message" do
```
slorek: Remove `,`

psd-web/spec/features/team_spec.rb
```diff
@@ -34,12 +34,27 @@ def resend_link_selector(email)
       end
     end
 
-    context "as a normal user" do
-      let(:user) { create(:user, :activated, :psd_user, teams: [team]) }
+    context "when I invite an existing user" do
```
slorek: the `context` here is `as a team admin`. The `scenario` is `inviting an existing user`

psd-web/spec/features/team_spec.rb
```diff
@@ -34,12 +34,27 @@ def resend_link_selector(email)
...
+        click_button "Send invitation email"
+        expect(page).to have_css(".govuk-error-summary__list", text: "You cannot invite this person to join your team because they are already a member of another team from a different organisation.")
       end
     end
   end
+
+  context "as a normal user" do
+    let(:user) { create(:user, :activated, :psd_user, teams: [team]) }
+
+    scenario "does not display the link for any users" do
```
slorek: *does not display the resend invite link

psd-web/spec/features/team_spec.rb
slorek: I think the b utton text would be resend invite not your team.

psd-web/spec/features/team_spec.rb
```diff
@@ -34,12 +34,27 @@ def resend_link_selector(email)
...
-    context "as a normal user" do
-      let(:user) { create(:user, :activated, :psd_user, teams: [team]) }
+    context "when I invite an existing user" do
+      let(:user) { create(:user, :activated, :team_admin, teams: [team]) }
 
-      scenario "does not display the link for any users" do
-        expect(page).not_to have_css("a[href*=\"#{resend_invitation_team_path(team)}\"]")
+      scenario ",I should see an error message" do
+        click_link "Invite a team member"
+        fill_in "new_user_email_address", with: user.email.to_s
```
slorek: Check we are on the correct page after clicking the link before attempting to fill in the form.

psd-web/spec/features/team_spec.rb
nasirkhanbeis: Not sure if i'm flowing this, as team admin there's only 1 scenario and another context as "as existing user" and scenario " I should see error message"

psd-web/spec/features/team_spec.rb
nasirkhanbeis: done

psd-web/spec/features/team_spec.rb
slorek: `context` blocks are used to set the scene, so you are describing what should be true before you perform any actions. In this case, we can wrap all the scenarios that may need to be tested as a team admin. `context` blocks should not be used to describe an action.

Also all of the other test descriptions are written in the third person, so it would be good to keep a consistent style.

I would write it as:

```ruby
context "as a team admin" do
  let(:user) { create(:user, :activated, :team_admin, teams: [team]) }
 
  scenario "scenario is inviting an existing user shows an error" do
    ...
  end
end
```

nasirkhanbeis: inviting an existing user should show an error message - I have particularly used the word "Should" as referring to specs and as part of QA tests I recommend we use these terms "Verify", "Should" etc. appreciate for elaborating the use of context which is something very commonly used in tests  as Given/pre-condition/assumption in my tests too. My first few tests were just to get myself familiar with codebase and ruby specs.
Hopefully going forward I will be providing test skeleton with scenarios so you guys can just pass the test as part of feature development!

slorek: :+1: We don't use the word "should" in this type of test as it is a description of what is supposed to happen - see http://www.betterspecs.org/#should. For context in old RSpec it used to use `should` syntax instead of `expect` (see the link for an example). In the other scenarios in the same test you are describing it in the correct way, i.e. "does not display" instead of "should not display".

nasirkhanbeis: ah okay, got it. In my QA type tests I normally prefer this word as they refer back to a/c what is defined, anyway I agree what you mean.

psd-web/spec/features/team_spec.rb
```diff
@@ -34,12 +34,30 @@ def resend_link_selector(email)
...
-      let(:user) { create(:user, :activated, :psd_user, teams: [team]) }
+    context "as a team admin" do
+      let(:user) { create(:user, :activated, :team_admin, teams: [team]) }
 
-      scenario "does not display the link for any users" do
-        expect(page).not_to have_css("a[href*=\"#{resend_invitation_team_path(team)}\"]")
+      scenario "inviting an existing user shows an error message" do
+        click_link "Invite a team member"
+        expect(page).to have_selector("h1", text: "Invite a team member")
+        fill_in "new_user_email_address", with: user.email.to_s
```
StupidCodeFactory: remove the `to_s` email is arlready a string

psd-web/spec/features/team_spec.rb
```diff
@@ -34,12 +34,30 @@ def resend_link_selector(email)
...
-    context "as a normal user" do
-      let(:user) { create(:user, :activated, :psd_user, teams: [team]) }
+    context "as a team admin" do
+      let(:user) { create(:user, :activated, :team_admin, teams: [team]) }
 
-      scenario "does not display the link for any users" do
-        expect(page).not_to have_css("a[href*=\"#{resend_invitation_team_path(team)}\"]")
+      scenario "inviting an existing user shows an error message" do
+        click_link "Invite a team member"
+        expect(page).to have_selector("h1", text: "Invite a team member")
```
StupidCodeFactory:
```suggestion
        expect(page).to have_css("h1", text: "Invite a team member")
```
we use `have_css` (clearer than `have_selector`) everywhere let's keep consistency

psd-web/spec/features/team_spec.rb
```diff
@@ -34,12 +34,30 @@ def resend_link_selector(email)
...
+        expect(page).to have_css(".govuk-error-summary__list", text: "You cannot invite this person to join your team because they are already a member of another team from a different organisation.")
       end
     end
   end
+
+  context "as a normal user" do
+    let(:user) { create(:user, :activated, :psd_user, teams: [team]) }
+
+    scenario "does not display the resend invite link for any users" do
+      expect(page).to have_selector("h1", text: "test organisation")
```
StupidCodeFactory:
```suggestion
      expect(page).to have_css("h1", text: "test organisation")
```

psd-web/spec/features/team_spec.rb
```diff
@@ -34,12 +34,30 @@ def resend_link_selector(email)
...
+      expect(page).not_to have_link("Your team", href: resend_invitation_team_path(team))
+    end
+
+    scenario "does not display the invite a team member link" do
+      expect(page).to have_selector("h1", text: "test organisation")
```
StupidCodeFactory:
```suggestion
      expect(page).to have_css("h1", text: "test organisation")
```

PR #265, opened 2020-02-07
Cache user roles in application database

slorek:
## Description
This saves the user roles from Keycloak in the application database, introducing a new model `UserRole`, in order to help facilitate the eventual removal of Keycloak from the system.

Previously, roles were cached in Redis for 30 minutes. This change reduces the effective cache delay to up to 10 minutes.

Due to limitations of the Keycloak API it is not possible to get all user roles in one, or even a few, requests. You have to iterate over each user. Currently there are around 1200 users, so that's 1200 extra API requests. To reduce the risk of the whole job failing in the event of a failed request I schedule a new job for each user.

I have increased the cron delay to 10 minutes due to the additional API requests required.

Minimal changes were needed to the existing tests which validates the approach.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/app/models/user.rb
+++ b/psd-web/app/models/user.rb
@@ -4,6 +4,7 @@ class User < ApplicationRecord
   has_many :investigations, dependent: :nullify, as: :assignable
   has_many :activities, through: :investigations
   has_many :user_sources, dependent: :destroy
+  has_many :user_roles, dependent: :destroy
 
   has_and_belongs_to_many :teams
 
@@ -55,6 +56,8 @@ class User < ApplicationRecord
 
         record.update!(user.slice(:name, :email, :organisation))
         record.teams = user[:teams]
+
+        SyncKeycloakUserRolesJob.perform_later(record.id)
       rescue ActiveRecord::ActiveRecordError => e
         if Rails.env.production?
           Raven.capture_exception(e)
@@ -65,22 +68,23 @@ class User < ApplicationRecord
     end
   end
 
+  def load_roles_from_keycloak
+    roles = KeycloakClient.instance.get_user_roles(id).uniq
+
+    return if roles == user_roles.pluck(:name).map(&:to_sym)
+
+    transaction do
+      user_roles.delete_all
+      roles.each { |role| user_roles.create!(name: role) }
+    end
+  end
+
   def self.current
     RequestStore.store[:current_user]
   end
 
   def self.current=(user)
     RequestStore.store[:current_user] = user
-  end
-
-  def has_role?(role)
-    roles.include?(role)
-  end
-
-  def roles
-    @roles ||= Rails.cache.fetch("user_roles_#{id}", expires_in: 30.minutes) do
-      KeycloakClient.instance.get_user_roles(id)
-    end
   end
 
   def name
@@ -138,9 +142,25 @@ class User < ApplicationRecord
     update has_viewed_introduction: true
   end
 
+  # TODO: Remove these when switching over roles resolution to #user_roles
+  def has_role?(role)
+    roles.include?(role)
+  end
+
+  def roles
+    @roles ||= Rails.cache.fetch("user_roles_#{id}", expires_in: 30.minutes) do
+      KeycloakClient.instance.get_user_roles(id)
+    end
+  end
+
 private
 
   def current_user?
     User.current&.id == id
   end
+
+  # TODO: Enable this when switching over roles resolution to #user_roles
+  # def has_role?(role)
+  #   user_roles.exists?(name: role)
+  # end
 end
--- a/psd-web/app/models/user_role.rb
+++ b/psd-web/app/models/user_role.rb
@@ -1,0 +1,5 @@
+class UserRole < ApplicationRecord
+  belongs_to :user
+  validates_presence_of :name
+  validates :name, uniqueness: { scope: :user }
+end
--- a/psd-web/db/migrate/20200207112514_create_user_roles.rb
+++ b/psd-web/db/migrate/20200207112514_create_user_roles.rb
@@ -1,0 +1,11 @@
+class CreateUserRoles < ActiveRecord::Migration[5.2]
+  def change
+    create_table :user_roles do |t|
+      t.belongs_to :user, type: :uuid
+      t.string :name, null: false
+      t.timestamps
+    end
+
+    add_index :user_roles, %i[user_id name], unique: true
+  end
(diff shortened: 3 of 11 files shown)


slorek:
Support ticket raised to try to fix review app deployment. In the meantime the code is ready for review, but we'll need to do some manual testing as well when the review app is working.

psd-web/app/models/user.rb
```diff
@@ -151,4 +152,8 @@ def has_viewed_introduction!
   def current_user?
     User.current&.id == id
   end
+
+  def has_role?(role)
+    user_roles.where(name: role).count.positive?
```
frankieroberto: [exists?](https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-exists-3F) might be marginally faster: 

```suggestion
    user_roles.exists?(name: role)
```

psd-web/db/migrate/20200207112514_create_user_roles.rb
```diff
@@ -0,0 +1,9 @@
+class CreateUserRoles < ActiveRecord::Migration[5.2]
+  def change
+    create_table :user_roles do |t|
+      t.belongs_to :user, type: :uuid
+      t.string :name
```
frankieroberto: Might be worth preventing nulls in these columns (even though these should get caught by the validation):

```suggestion
      t.string :name, null: false
```

psd-web/app/models/user.rb
```diff
@@ -65,6 +67,15 @@ def self.load_from_keycloak(users = KeycloakClient.instance.all_users)
     end
   end
 
+  def load_roles_from_keycloak
+    roles = KeycloakClient.instance.get_user_roles(id)
+
+    transaction do
+      user_roles.delete_all
+      roles.each { |role| user_roles.create(name: role) }
```
frankieroberto: If this is run frequently, it might generate a lot of unnecessary database deletes and inserts (given that mostly the roles won't change)? However, given that we're likely ripping this out soon, it probably doesn't matter..

psd-web/app/models/user.rb
```diff
@@ -73,16 +84,6 @@ def self.current=(user)
     RequestStore.store[:current_user] = user
   end
 
-  def has_role?(role)
```
slawosz: I would suggest not removing old `has_role?` and `roles` methods. Instead, comment out new `has_role?`. This way, we can merge and deploy this one, test if there are some issues with sidekiq and roles synchronization, and if they are we can still use old, direct roles query with redis cache. After a while of this being run on production, we can verify everything is working fine and switch `has_role?` implementation.

psd-web/db/migrate/20200207112514_create_user_roles.rb
```diff
@@ -0,0 +1,9 @@
...
+      t.timestamps
```
frankieroberto: Could drop an index in here:

```ruby
add_index :user_roles, [:user_id, :name], unique: true
```

(the model validation should also cover this, but that _can_ fail due to race conditions between the select and the insert. Should also also make the select faster.)

psd-web/app/models/user_role.rb
```diff
@@ -0,0 +1,4 @@
+class UserRole < ApplicationRecord
```
StupidCodeFactory: why not just `Role` ? this suggests this is a join model between User and Role, I immediately looked for the Role model.
I think this could actually be modelled as a has_many :through.
We're going to have tons of dup'd roles. probably all meaning the same. I'm not quite sure this is the correct way to model this.

slorek: This was in anticipation of possibly adding a `Role` model later which define a bit more about what those roles do, i.e. the permissions they grant.

slorek: Effectively this is just an intermediary step to isolate the app from Keycloak, not the final design.

psd-web/app/models/user.rb
slawosz: Thats actually good catch, for fast refactor you can do:
```
return if roles == user_roles.pluck(:name)
```
So it wont recreate roles.

psd-web/db/migrate/20200207112514_create_user_roles.rb
slawosz: Index on both columns won't make things faster, providing we have small amout of roles. Regarding validation, I would add unique constraint to database.

psd-web/db/migrate/20200207112514_create_user_roles.rb
```diff
@@ -0,0 +1,9 @@
+class CreateUserRoles < ActiveRecord::Migration[5.2]
+  def change
+    create_table :user_roles do |t|
+      t.belongs_to :user, type: :uuid
```
StupidCodeFactory: maybe also add and index

psd-web/app/models/user.rb
slorek: Yes, it will.

psd-web/app/models/user.rb
slorek: Could be a good idea.

psd-web/db/migrate/20200207112514_create_user_roles.rb
frankieroberto: @slawosz well no, at our scale the difference is likely marginal. 😄 Adding the index above does also add the database uniqueness constraint though.

psd-web/app/models/user.rb
StupidCodeFactory: also maybe add a `user_roles.create!` I can't remember but I don't think a simple `#create` will trigger a db rollback if anything funny happens

slorek: Unfortunately there can be duplicate roles so this would break it.

frankieroberto: How about:
`roles.uniq.each { |role| user_roles.create!(name: role) }`

psd-web/db/migrate/20200207112514_create_user_roles.rb
slorek: Indexes are added to `belongs_to` automatically.

slorek:
Per @slawosz suggestion I have commented out the new #has_role? functionality to allow us to test the background job before switching over. I've also addressed all the other feedback.

PR #277, opened 2020-02-12
Add index to account_activated column on users table to improve query performance

slorek:
## Description
While browsing Scout I noticed a few slow database queries (average 69ms, or 19% of total page load) reported. Here is an example: https://scoutapm.com/apps/156303/endpoints/Q29udHJvbGxlci9pbnZlc3RpZ2F0aW9ucy9pbmRleA==#QWN0aXZlUmVjb3JkL1VzZXIvZmluZA==

Looking at the query I noticed a missing index on the `account_activated` column, so I think adding this index should improve the speed of these queries.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


psd-web/db/migrate/20200212120801_add_index_on_account_activated_to_users.rb
```diff
@@ -0,0 +1,7 @@
+class AddIndexOnAccountActivatedToUsers < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def change
+    add_index :users, :account_activated, algorithm: :concurrently
```
frankieroberto: If we’re _always_ querying for active users, then we could use a partial index containing only active users?

slorek: Not always, many times we need to retrieve inactive users as well.

slorek:
Will merge after #217

psd-web/db/migrate/20200212120801_add_index_on_account_activated_to_users.rb
frankieroberto: Good point. In theory we could add 2 partial indexes (for active and inactive users), but that might be overkill.

PR #283, opened 2020-02-13
fix omniauth

StupidCodeFactory:
## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


psd-web/app/views/homepage/show.html.slim
```diff
@@ -26,7 +26,7 @@
             = link_to "Sign in to your account",
                 user_openid_connect_omniauth_authorize_path,
                 class: "govuk-button govuk-button--start app-button--inverted",
-                role: "button", draggable: false
+                role: "button", draggable: false, method: :post
```
frankieroberto: I think we should probably convert this to an actual form with a button instead? `link_to path, method: :post` uses javascript to dynamically generate the form, so this wouldn't work if javascript was turned off / unavailable / broken.

StupidCodeFactory: ```OmniAuth.config.allowed_request_methods = [:post, :get]```

frankieroberto: https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284 suggests that:

> using GET requests for /auth/:provider is not recommended and you should try to stick to POST requests if at all possible

If we switch to a form doing a POST, then we should be able to remove `OmniAuth.config.allowed_request_methods = [:post, :get]` and stick with the default from `omniauth-rails_csrf_protection` to only allow POST requests?

PR #284, opened 2020-02-13
Remove Keycloak

frankieroberto:
Remove Keycloak 🗡

PR #286, opened 2020-02-14
Stop using keycloak for team invites

frankieroberto:
This changes the invitation system to remove the need for Keycloak, instead adding the users and the invitation to the database directly, and sending the invitation via GOV.UK Notify.

It makes as few changes to the existing code as possible, saving any refactoring for the future.

The invitations are attached directly the the `users` table, via new `invitation_token` and `invited_at` columns.

The sending of the invitations happens asynchronously via a new `SendUserInvitationJob`, which calls a new method on the existing `NotifyMailer` class (which uses GOV.UK Notify).

The invitations themselves contain a link to a new route which is not yet implemented, as that will be a separate task within this epic.

Review app: https://psd-pr-286.london.cloudapps.digital


psd-web/app/jobs/send_user_invitation_job.rb
```diff
@@ -0,0 +1,16 @@
...
+  GOV_UK_NOTIFY_TEMPLATE_ID = "22b3799c-aa3d-43e8-899d-3f30307a488f".freeze
+
+  def perform(user_id, user_inviting_id)
+    user = User.find(user_id)
+    user_inviting = User.find(user_inviting_id)
+
+    invitation_url = create_account_user_url(user.id, invitation: user.invitation_token, host: ENV.fetch("PSD_HOST"))
+
+    NotifyMailer.invitation_email(user.email, invitation_url, user_inviting.name).deliver_now
+    user.update(has_been_sent_welcome_email: true, invited_at: DateTime.now)
```
StupidCodeFactory:
```suggestion
    user.update!(has_been_sent_welcome_email: true, invited_at: Time.current)
```

psd-web/config/routes.rb
```diff
@@ -25,6 +25,12 @@
     end
   end
 
+  resources :users, only: [] do
+    member do
+      get :create_account
+    end
+  end
```
StupidCodeFactory: how is that used anywhere? also can you avoid mixing up `resources` and `actions` this really should be expressed this way:
```ruby
resources :users, only: [] do
  resource :account, only: :create
end
```
avoiding using member/get/post/patch/delete/collection is a great way to ensure you nailed what the problem domain is and helps reasoning about single responsibility principle.

psd-web/app/jobs/send_user_invitation_job.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class SendUserInvitationJob < ApplicationJob
+  include Rails.application.routes.url_helpers
+  GOV_UK_NOTIFY_TEMPLATE_ID = "22b3799c-aa3d-43e8-899d-3f30307a488f".freeze
+
+  def perform(user_id, user_inviting_id)
+    user = User.find(user_id)
+    user_inviting = User.find(user_inviting_id)
+
+    invitation_url = create_account_user_url(user.id, invitation: user.invitation_token, host: ENV.fetch("PSD_HOST"))
```
StupidCodeFactory: can we add all the environment variable added to config/application.rb
```
confid.psd.host = ENV.fetch("PSD_HOST", "some_default_value_or_not_or_better_setting_it_in_its_enviroment_file")

File: psd-web/app/jobs/send_user_invitation_job.rb [hunk 4]
```diff
@@ -0,0 +1,16 @@
...
+
+    NotifyMailer.invitation_email(user.email, invitation_url, user_inviting.name).deliver_now
```
StupidCodeFactory: 1. How do we now the API call has been successful? 
2. should we rescue something?
3. should we set the flags only once we are certain this worked? 

Possibly there is nothing to do here, however I'ld like to have confirmation that `NotifyMailer#invitation_email` raises an error. This can be demonstrated in a spec raising the expected exception.
```ruby
allow(NotifyMailer).to receive(:invitation_email).and_raise(WhateverNotifyError)
travel_to nowish do
  expect { subject.perform )
    .to_not change(user, : has_been_sent_welcome_email).from(false).to(true)
    .and change(user, invited_at: nowish).from(nil).to(nowish)
end
```

psd-web/app/jobs/send_user_invitation_job.rb
StupidCodeFactory: Inject `invitation_url` this way this class doesn't need to about about how to construct a url.
In real life this will be inject in the controller.

StupidCodeFactory: And you won't need to add the url helpers, this is a bit of a code smell.

psd-web/app/controllers/teams_controller.rb
```diff
@@ -39,7 +39,7 @@ def resend_invitation
 private
 
   def resend_invitation_to_user(email_address)
-    User.resend_invite email_address, @team, root_url
+    User.resend_invite email_address, @team, root_url, current_user
```
StupidCodeFactory:
```suggestion
    User.resend_invite email_address, @team, root_url, current_user, create_account_user_url(current_user, invitation: current_user.invitation_token, host: ENV.fetch("PSD_HOST"))
    
```

StupidCodeFactory: read explanation below

StupidCodeFactory:
@frankieroberto I think there is potentially a bug. You probably want to ensure the log in query doesn't load invited users.
It's one of the reason I think this is a bad design and and invite model should have been created. 
Now we have tangle up together User which can already be on two states, activated true/false.
Now were add another state which is invited which itself has three state, invited present or not and is the invitation expired or not. 
The possible combination of a user state has just blown up. 

I think we should consider creating an Invite model, or there are lot more policy tests to write and you need to write a policy so that user who are in invited states can not access whatever.

StupidCodeFactory:
@frankieroberto @scruti I can also how this can easily blow up in complexity when add in permissions/roles.

StupidCodeFactory:
I'm also not sure why you're not using devise's API to send notification

https://www.rubydoc.info/github/plataformatec/devise/master/Devise%2FModels%2FAuthenticatable:send_devise_notification

psd-web/app/jobs/send_user_invitation_job.rb
frankieroberto: This doesn't do the API call itself, it just calls the `NotifyMailer`, which in turn calls the Notify API.  I don't think we need to rescue anything - if the NotifyMailer throws an exception then Sidekiq will mark the job as failed and will retry it later (which seems like the right behaviour).

I'm not sure the spec is that helpful – isn't it just asserting the standard ruby behaviour that any exceptions stop any further code from being run?

StupidCodeFactory: not really no - I'm just asking whether we have we done diligence of checking what happens when the API call fails.
Can you for sure tell me what NotifyMailer does when it fails sending an email?

frankieroberto:
@StupidCodeFactory the existing behaviour already adds invited users into the `users` table straight away, and uses an `account_activated` boolean to track which users have set up their password, etc – so we didn't want to change this in the short term.

I think there’s also some advantages for having users and invited users in the same table - eg it means we can add a database constraint to make sure that the e-mail address are unique across both users and invited users, which wouldn't be possible if they were separate tables.

psd-web/app/jobs/send_user_invitation_job.rb
StupidCodeFactory: Reason I'm asking is because `NotifyMailer` does not raise an on error exception. 
The doc is pretty clear about it https://github.com/alphagov/notifications-ruby-client/blob/300a4bf410cc7a45a60561aa48c94a8dfd8815d0/DOCUMENTATION.md#error-codes
The is a bit more error handling to do

psd-web/app/jobs/send_user_invitation_job.rb
frankieroberto: Good point. We wrote this code before we realised that NotifyMailer existed (had assumed we'd call the Notify API directly from the job).

I'll move the call to the url helper to the `NotifyMailer` for now. That makes it consistent with the existing code for the `NotifyMailer.investigation_created` method.

psd-web/app/models/user.rb
```diff
@@ -17,21 +17,24 @@ def self.activated
...
     team.add_user(user)
 
-    KeycloakClient.instance.send_required_actions_welcome_email keycloak_user[:id], redirect_url
+    SendUserInvitationJob.perform_later(user.id, inviting_user.id)
   end
 
-  def self.resend_invite(email_address, _team, redirect_url)
-    user_id = KeycloakClient.instance.get_user(email_address)[:id]
-    KeycloakClient.instance.send_required_actions_welcome_email user_id, redirect_url
+  # TODO: remove `_team` and `_redirect_url` arguments, as these are no
```
slawosz: So how about removing them now?

psd-web/app/jobs/send_user_invitation_job.rb
frankieroberto: @StupidCodeFactory I _think_ that documentation is a bit misleading, as the `Notifications::Client` code does seem to raise exceptions.

I've verified this by running

```ruby
client.send_email(email_address: "test@test.com", template_id: SecureRandom.uuid)
```

in a console, which throws:

```ruby
Notifications::Client::BadRequestError (BadRequestError: Template not found)
```

Tracing through the client code, I think this happens within : [`perform_request!`](https://github.com/alphagov/notifications-ruby-client/blob/1613606acf2ff6dc66547a9a3a11766ba0d506c2/lib/notifications/client/speaker.rb#L143) ... and there are [specs for it](https://github.com/alphagov/notifications-ruby-client/blob/1613606acf2ff6dc66547a9a3a11766ba0d506c2/spec/notifications/client/request_errors_spec.rb).

psd-web/app/models/user.rb
frankieroberto: @slawosz good point. Our original intention was to avoid any changes to the method API so that we didn't have to touch the controller code. However given we ended up needing to add an argument, it’s not much extra effort to remove the unused ones. I've done that now. 👍

psd-web/app/controllers/teams_controller.rb
frankieroberto: 👍  I've moved the url helper call to the mailer.

psd-web/app/models/user.rb
```diff
@@ -17,21 +17,24 @@ def self.activated
...
-  def self.create_and_send_invite(email_address, team, redirect_url)
-    KeycloakClient.instance.create_user(email_address)
-
-    keycloak_user = KeycloakClient.instance.get_user(email_address)
-
-    # Create the user in the local cache database so that we don't have to wait until the next sync
-    user = create(id: keycloak_user[:id], email: email_address, organisation: team.organisation)
+  # TODO: remove `_redirect_url` argument, as this is no longer used.
+  def self.create_and_send_invite(email_address, team, _redirect_url, inviting_user)
+    user = create(
```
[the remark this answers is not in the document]
slorek: Agreed, we should add a validation on `email` (and use the whitelist for this) but it is actually already validated in the controller by the time it gets here (see https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/e541e843049da4ee2dd6d7ce565c589a9af4cdb6/psd-web/app/controllers/teams_controller.rb#L49) so in the spirit of not adding extra refactoring I'm happy to let this go for now. We should probably use `create!` here as well.

psd-web/app/models/user.rb
```diff
@@ -17,21 +17,21 @@ def self.activated
...
 
-    KeycloakClient.instance.send_required_actions_welcome_email keycloak_user[:id], redirect_url
+    SendUserInvitationJob.perform_later(user.id, inviting_user.id)
   end
 
-  def self.resend_invite(email_address, _team, redirect_url)
-    user_id = KeycloakClient.instance.get_user(email_address)[:id]
-    KeycloakClient.instance.send_required_actions_welcome_email user_id, redirect_url
+  def self.resend_invite(email_address, inviting_user)
+    user = User.find_by!(email: email_address)
```
slorek: Not a biggie but we might want to check the user belongs to the same team as `inviting_user` here by adding an extra condition.

psd-web/app/controllers/teams_controller.rb
```diff
@@ -17,7 +17,7 @@ def invite_to
     if existing_user
       invite_existing_user_if_able existing_user
     elsif whitelisted_user
```
slorek: I don't think we're doing any checking if the `current_user` has the `team_admin` role. An existing bug but one we should fix (i.e. create a new ticket if not part of this PR).

psd-web/app/models/team.rb
```diff
@@ -10,10 +10,7 @@ class Team < ApplicationRecord
   validates :path, presence: true
 
   def add_user(user)
-    # Update the local cached team membership so the change appears immediately
```
slorek: Do we think this method is still adding value? We can probably remove it I think.

psd-web/app/models/user.rb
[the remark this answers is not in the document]
frankieroberto: @StupidCodeFactory the validation errors are currently added by the [`TeamsController`](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/controllers/teams_controller.rb) – not ideal, but we didn't want to refactor this until later.

psd-web/config/routes.rb
slorek: It appears to be used in the mailer to generate the URL.

slorek: There is no `account` resource, they are just completing their profile. Perhaps there is a better name for the route.

psd-web/spec/jobs/send_user_invitation_job_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "rails_helper"
+
+RSpec.describe SendUserInvitationJob do
+  subject(:perform) { described_class.new.perform(user_id, user_inviting.id) }
```
slorek: Previously @StupidCodeFactory expressed issues with using `subject` in this way, preferring `let`. Let's just check everyone is using it consistently so the tests are consistent.

psd-web/config/routes.rb
frankieroberto: This was a bit of a placeholder until we come to implement the create account story, and adding a method to the existing controller was quicker.

Let's come back to this when implementing the account creation story. It might make sense to create a new controller & nested resource at that point.

psd-web/spec/models/user_spec.rb
```diff
@@ -10,6 +10,67 @@
     end
   end
 
+  describe ".create_and_send_invite" do
+    context "with valid email and team" do
+      let(:email) { "testuser@southampton.gov.uk" }
+      let(:team) { create(:team) }
+      let(:inviting_user) { create(:user) }
+
+      before do
```
slorek: Generally single-line blocks should be expressed with curly braces instead of `do...end`. I'm not sure why Rubocop is not picking this up.

psd-web/app/models/user.rb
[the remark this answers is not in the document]
StupidCodeFactory: 99% of tech debt is due to I'll refactor this later :)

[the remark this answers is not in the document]
StupidCodeFactory: I don't have numbers to back this up. 

However we definitely want to use `create!`

psd-web/app/models/team.rb
frankieroberto: No, it's definitely not... 😂   We didn't want to delete it until we’d checked where else it was being called from. But I've just checked and it's only called from one other place, so it won't be hard to remove (without falling too far into the refactoring hole).

psd-web/app/models/user.rb
StupidCodeFactory:
```suggestion
    user = inviting_user.teams.find_by!(email: email_address)
```

psd-web/spec/jobs/send_user_invitation_job_spec.rb
frankieroberto: I'm fairly agnostic on this, but @scruti was in favour of using subjects. Let's agree an approach.

psd-web/spec/models/user_spec.rb
frankieroberto: Our specs are full of single line blocks using do...end (eg `it` blocks containing a single `expect` line).

Personally I find

```ruby
      before do
        allow(SendUserInvitationJob).to receive(:perform_later)
      end
```

to be slightly more scannable than
```ruby
```

given the nested braces & parentheses.

Looks like Rubocop [has some settings for this](https://rubocop.readthedocs.io/en/latest/cops_style/#styleblockdelimiters) though, so if we agree a particularly style we should be able to get Rubocop to auto-correct it.

psd-web/app/controllers/teams_controller.rb
frankieroberto: I _think_ this is handled by the [`TeamPolicy`](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/policies/team_policy.rb#L6-L8) but I haven't yet gotten my head around how the policy classes and Pundit works. I think the policy might be being invoked when [`set_team`](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/controllers/teams_controller.rb#L60) is called...

psd-web/app/models/user.rb
frankieroberto: Good point! The existing code doesn't do this though, so can we raise this as a separate ticket?

slorek: It’s probably a very simple fix so we should be able to do it now.

frankieroberto: Well one complication is that users can currently belong to multiple teams – although in practice this hasn't happened, and we might remove this ability in future.

So the simplest fix would be something like

```ruby
user = inviting_user.teams.first.users.find_by!(email: email_address)
```

but in theory in should be something more like

```ruby
user = User.find_by!(email: email_address)
raise ActiveRecord::RecordNotFound unless user.teams & inviting_user.teams
```
(ie checking the intersection of the two users’ teams)

psd-web/spec/models/user_spec.rb
StupidCodeFactory: I favour the one liners in `{}` but not fussed at all on that.

psd-web/app/models/user.rb
StupidCodeFactory: this would always raise

StupidCodeFactory:
```suggestion
    user = User.where(teams: inviting_user.teams).find_by!(email: email_address)
```
this is probably what you'd want.

psd-web/spec/jobs/send_user_invitation_job_spec.rb
StupidCodeFactory: so yes, `subject` is always and instance of the described class. Otherwise when you open a spec you never now what subject is unless you jump up and down the file, loosing context of what you're trying to do. It's crucial we must trust that `subject` is always and instance of `described_class`.
and within the body of the test: 
`subject.perform` so when we read the body of the test we instantly know what is being tested.

and also certainly never do something like:

```ruby
subject { instance.boolean_method }
it { is_expected.to eq(true) }
```
which totally defeats the purpose of spec being documentation, and whichever file you open `it` blocks are all the same. Adds a tremendous cognitive load to understanding the tests.

psd-web/spec/jobs/send_user_invitation_job_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    it "sends an email via the NotifyMailer" do
+      expect(NotifyMailer).to receive(:invitation_email).with(user, user_inviting).and_return(message_delivery_instance)
+      expect(message_delivery_instance).to receive(:deliver_now)
+      perform
+    end
+
+    it "adds the time that the user was invited" do
+      perform
+      expect(user.reload.invited_at).to be_within(2.seconds).of(Time.zone.now)
```
StupidCodeFactory: so no let's do this properly:

```ruby
let(:nowish) { Time.current }

it "adds the time that the user was invited" do
  travel_to nowish do  
    perform
    expect(user.reload.invited_at).to eq(nowish)
  end
end
```

psd-web/spec/jobs/send_user_invitation_job_spec.rb
scruti: **TLDR:** Happy to change this specific case but not so happy about "subject must always be an instance of the described class" taken as a rule.

**Long version:**
As much as I am in favour of reducing code indirection as much as possible and have explicit tests to avoid the situation you described, I don't see how "`subject` is always an instance of the described class" is a rule to follow. 
If we're testing an specific method of a class public interface, the subject under test is conceptually that method, not an instance of the described class.
I also favour to use explicit subjects (naming them) in order to avoid the lack of explicitness situation, and named subjects are not precisely meant to just represent class instances. 
In this case, I see that calling `perform` makes quite clear that we're telling the job to `perform`.  
 
Given that (if I recall correctly) `subject` and `let` are basically the same method but named differently, I have also seen this argument used to discourage the use of `let` altogether. 

Like almost anything in coding I think the right approach is based on pragmatism and trade-off analysis, and there is not a golden rule for every situation.

Said this, given this specific case, I don't mind at all just calling `described_class.new.perform(user_id, user_inviting.id)` or creating a `subject(:job) { described_class.new)` and call it from each spec as `job.perform(user_id, user_inviting.id)`, and given that this is a job class that basically performs one single thing and exposes a single method I can even visualise the job itself being the subject under test.  
I just don't agree with the "subject must be an instance of the class" taken as a rule when using rspec.

psd-web/app/mailers/notify_mailer.rb
```diff
@@ -10,10 +10,13 @@ class NotifyMailer < GovukNotifyRails::Mailer
     }.freeze
 
 
-  def invitation_email(email, invitation_url, inviting_team_member_name)
+  def invitation_email(user, inviting_user)
     set_template(TEMPLATES[:invitation])
+
+    invitation_url = create_account_user_url(user.id, invitation: user.invitation_token)
```
scruti: :clap:

psd-web/app/models/team.rb
```diff
@@ -9,10 +9,6 @@ class Team < ApplicationRecord
   validates :name, presence: true
   validates :path, presence: true
 
-  def add_user(user)
```
scruti: :clap:

psd-web/spec/jobs/send_user_invitation_job_spec.rb
scruti: I always forget Rails added that helper <3

psd-web/spec/models/user_spec.rb
scruti: I'm one the same boat as Yann. 
I tend to write them (if short lines) as one liners but do not mind much about it.

scruti: Also... should we add `rubocop-rspec` to the project? Many of this style issues/differences would be just normalised by using it.

psd-web/spec/jobs/send_user_invitation_job_spec.rb
StupidCodeFactory: yeah so what seems pragmatic to me, when I have 250 specs to fix the that I can open the spec at the failure i.e `spec/services/create_user_spec.rb:253` and be sure that subject is an instance of `described_class` and i can see within the body the `it` what is under test. 
Otherwise you have to scroll and down the file scanning nested context, it is just madness.
Don't worry you will get use to this form of pragmatism, and I'm sure even come to like and spread the good word.

psd-web/app/jobs/send_user_invitation_job.rb
```diff
@@ -0,0 +1,14 @@
+require "notifications/client"
+
+class SendUserInvitationJob < ApplicationJob
+  include Rails.application.routes.url_helpers
```
StupidCodeFactory: ✂️

psd-web/app/models/user.rb
```diff
@@ -17,21 +17,21 @@ def self.activated
     where(account_activated: true)
   end
 
-  def self.create_and_send_invite(email_address, team, redirect_url)
-    KeycloakClient.instance.create_user(email_address)
+  def self.create_and_send_invite(email_address, team, inviting_user)
+    user = create(
```
StupidCodeFactory: you want to kown if this fails - if user is invalid what happens?
```suggestion
    user = create!(
```

psd-web/spec/jobs/send_user_invitation_job_spec.rb
scruti: Attempted to use rails Time freezing helpers but ended reverting to the `be_within` original approach.

Was getting errors in the form of:
```ruby
expected: 2020-02-18 17:12:47.433394984 +0000
got: 2020-02-18 17:12:47.000000000 +0000
```
Seems they where caused because [ruby time objects stores greater precision than database datetime objects](https://stackoverflow.com/questions/20403063/trouble-comparing-time-with-rspec).
Given that, when parsing back the database value and comparing it against the ruby object... they're different.

Seems the `be_within` is a good way to go in this case (and probably why Frankie originally used it) as other proposed solutions (like casting the values) can have failures depending how close to the end of the second is `Time.now` when invoked. Reduced the allowed time disparity to 1 sec though.

psd-web/app/controllers/teams_controller.rb
scruti: Did a bit of investigation here:
Effectively, as @frankieroberto mentioned, [`TeamPolicy`](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/policies/team_policy.rb#L6-L8) is handling this. 
The policy is being applied to the team with [`set_team`](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/controllers/teams_controller.rb#L60) call. 
Once the policy gets applied, Pundit maps the helper methods [`show?`](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/policies/team_policy.rb#L2) and [`invite_to?`](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/policies/team_policy.rb#L6) to the corresponding actions in the controller: [`show`](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/controllers/teams_controller.rb#L9) and [`invite_to`](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/controllers/teams_controller.rb#L12). So depending on the controller action name the corresponding policy check is being injected/applied before actually running the action.

psd-web/spec/jobs/send_user_invitation_job_spec.rb
scruti: to be honest I've just spent some time catching up on [discussions on BetterSpecs](https://github.com/lelylan/betterspecs/issues/7) and some [blog posts](https://semaphoreci.com/community/tutorials/rspec-subject-helpers-hooks-and-exception-handling) about the matter and I've changed my mind about how I've been using `subject` until now.

Seems like an step in the same direction as the "stop this over-nesting contexts and redefining `lets` crazyness" approach that I recently embraced after having to deal with too complex test setups. 

Thanks for bringing this up! :+1:

psd-web/spec/models/user_spec.rb
frankieroberto: @scruti yep, happy to add `rubocop-rspec`, but let's do that in a separate PR against master? We’re already using this in the cosmetics project.

frankieroberto: Marking this as resolved for now – as we can decide (and enforce) a rule for this separately.

scruti:
@StupidCodeFactory @slorek change requests addressed.

frankieroberto:
@StupidCodeFactory @slorek I've merged this into the feature branch, as we've addressed all the outstanding comments, so that we can start work on the next, dependent feature.

Feel free to add any further comments to the parent PR though: #284.

PR #289, opened 2020-02-14
Authenticate user using Keycloak DB

slawosz:
Allows authentication using existing Keycloak DB. Exposes one method:
```
KeycloakCredential.authenticate(email, plain_password)
```

returns `true` when password matches, `false` when password is wrong.
Raises `ActiveRecord::RecordNotFound` when email does not exists.


psd-web/config/routes.rb
```diff
@@ -12,6 +12,9 @@
 
 # rubocop:disable Metrics/BlockLength
 Rails.application.routes.draw do
+  get 'verify_keycloak/index'
```
slawosz: Entries in `routes.rb` will be removed before merging.

psd-web/app/views/verify_keycloak/index.html.erb
```diff
@@ -0,0 +1,34 @@
+
```
slawosz: This will be deleted before merging

psd-web/app/controllers/verify_keycloak_controller.rb
```diff
@@ -0,0 +1,36 @@
+class VerifyKeycloakController < ActionController::Base
```
slawosz: This will be deleted before merging

psd-web/config/initializers/sidekiq.rb
```diff
@@ -38,11 +38,26 @@ def schedule_keycloak_sync_job
   end
 end
 
+def schedule_keycloak_passwords_sync_job
+  keycloak_sync_job = Sidekiq::Cron::Job.new(
+    name: "#{ENV['SIDEKIQ_QUEUE'] || 'psd'}: Sync users and teams with Keycloak",
+    cron: "*/1 * * * *",
```
slawosz: I will change this to every 5 minutes after testing (just before merging).

psd-web/app/models/keycloak_connector.rb
```diff
@@ -0,0 +1,21 @@
+class KeycloakConnector < ApplicationRecord
+  KEY_LEN = 64
```
scruti: This doesn't seem to be used anywhere... :scissors: ?

psd-web/config/database.yml
```diff
@@ -16,3 +16,12 @@ production:
   connect_timeout: 2
   variables:
     statement_timeout: <%= ENV["STATEMENT_TIMEOUT"] || "500ms" %>
+
+keycloak:
```
scruti: I am guessing this was created as independent database in order to be able to make easier the complete remove of Keycloak in the future right?

slawosz: This is database used currently by keycloak.

psd-web/db/migrate/20200213114958_create_keycloak_credentials.rb
```diff
@@ -0,0 +1,14 @@
+class CreateKeycloakCredentials < ActiveRecord::Migration[5.2]
+  def change
+    create_table :keycloak_credentials do |t|
```
scruti: should we add some enforced `null: false` fields or it doesn't matter?

psd-web/spec/factories/keycloak_credentials.rb
```diff
@@ -0,0 +1,10 @@
+FactoryBot.define do
+  factory :keycloak_credential do
+    salt { "\xC6\xAF\xE7\x91y\x91Fn\xE2\v\xF4\xE3^A\x98g" }
+    # passwordpasswordpasswordpassword
```
scruti: cutcutcutcutcut :D

psd-web/db/migrate/20200213114958_create_keycloak_credentials.rb
slawosz: I don't think we need. Data in this table are copy of some fields from keycloak DB.

psd-web/spec/models/keycloak_credential_spec.rb
```diff
@@ -0,0 +1,34 @@
+require "rails_helper"
+
+RSpec.describe KeycloakCredential, type: :model do
+  context "authentication" do
```
scruti: shouldn't be this a `describe ".authenticate" do` instead?

psd-web/spec/models/keycloak_credential_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    before do
+      create(:keycloak_credential)
+    end
+
+    let(:password) do
+      "passwordpasswordpasswordpassword"
+    end
+    let(:email)       { "test@example.org" }
+    let(:wrong_email) { "wrong@example.org" }
```
scruti: This seems unused. :scissors: ?

psd-web/spec/models/keycloak_credential_spec.rb
slawosz: I guess its a matter of preference.

psd-web/spec/models/keycloak_credential_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    context "when successful" do
+      let(:email)       { "test@example.org" }
+
+      it "returns true" do
+        expect(KeycloakCredential.authenticate(email, password)).to be_truthy
+      end
+    end
+
+    context "when unsuccessful" do
```
scruti: Just a suggestion (not a requirement): Given the context is not setting any shared values/states between the tests it wraps... I'm all in to remove the context nesting and get a flat as possible test list.

psd-web/spec/models/keycloak_credential_spec.rb
scruti: I'm all in for [BetterSpecs](http://www.betterspecs.org) guidelines as standards. Is not used as standard here?

psd-web/config/database.yml
scruti: Thanks for the clarification!

slorek:
Looks good, I would only question whether we should instead have a relationship between `User` and `KeycloakCredential` rather than just an `email` column. This would save us importing credentials for users which don't exist on PSD (e.g. Cosmetics users or deleted ones) and also make the interface to lookup credentials easier (e.g. `user.keycloak_credentials`).

slawosz:
@slorek I am still not sure what approach is the best here. By the best I mean easiest to implement/error prone. This pull request is kind of MVP and it does not introduces any decisions - yet. I am cooperating with @StupidCodeFactory as he is developing password authentication.

psd-web/spec/models/keycloak_credential_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+RSpec.describe KeycloakCredential, type: :model do
+  describe ".authenticate" do
+    before do
+      create(:keycloak_credential)
+    end
+
+    let(:password) do
+      "passwordpasswordpasswordpassword"
+    end
+    let(:email)       { "test@example.org" }
```
scruti: another annoying pure style comment: :scissors: empty spaces.

(going to open a proposal somewhere to start using `rubocop-rspec` and have a bot being annoying and getting the "oh F. you" reaction instead of a peer :laughing: )

infrastructure/env/get-env-from-vcap.sh
```diff
@@ -12,3 +12,5 @@ echo $VCAP_SERVICES | ./env/jq -r '
                 | "\(.key)=\(.value)"
         )
         | @tsv'
+
+echo "KEYCLOAK_DATABASE_URI=$(echo $VCAP_SERVICES | ./env/jq -r '.postgres[] | select(.name == "keycloak-database") .credentials .uri')"
```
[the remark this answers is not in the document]
slawosz: No, we need it to access keycloak db.

psd-web/spec/factories/keycloak_credentials.rb
slawosz: Why? Its worth to know what is original password.

PR #273, opened 2020-02-11
Devise db authenticatable

StupidCodeFactory:
**CHANGES:** 

Add ability to login using email/password
[trello](https://trello.com/c/PLwIClnd/329-5-ability-to-log-in-without-2fa)
## Checklist:

- [ ] Automated checks are passing locally.
- [ ] CHANGELOG updated if change is worth telling users about.
### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only
- [ ] Tested with one screen reader
- [ ] Zoom page to 400% - content still visible
- [ ] Disable css - does content make sense and appear in a logical order?
- [ ] Passes automated checker (automated in build or manual)

PR #294, opened 2020-02-19
Creating an account from an invitation

scruti:
Provides the implementation for the user account creation after following the link in the invitation email.

If the user is already signed in as a different user, we show this view, asking if they would prefer to sign out before continuing or not:

Screenshot 2020-02-24 at 12 37 40

If the invitation has expired (after 14 days), we display this view (content tbc):

Screenshot 2020-02-24 at 12 37 48


psd-web/app/models/user.rb
```diff
@@ -11,6 +14,21 @@ class User < ApplicationRecord
 
   validates :id, presence: true, uuid: true
 
+
+  with_options on: :registration_completion do |registration_completion|
+    registration_completion.validates :mobile_number, presence: true
+    registration_completion.validate :validate_mobile_number
+    registration_completion.validates :name, presence: true
+    registration_completion.validates :password, presence: true
```
StupidCodeFactory: This is reduntant, if you validate for an input to have a length > 0 then it is essentially validating for the presence.

psd-web/spec/features/creating_an_account_from_an_invitation_spec.rb
```diff
@@ -0,0 +1,65 @@
+require "rails_helper"
+require "support/feature_helpers"
+
+RSpec.feature "Creating an account from an invitation", :with_stubbed_elasticsearch, :with_stubbed_antivirus, :with_stubbed_mailer, :with_stubbed_keycloak_config do
+  let(:invited_user) { create(:user, :invited) }
+  let(:existing_user) { create(:user) }
+
+  scenario "Creating an account from an invitation" do
+    visit "/users/#{invited_user.id}/complete-registration?invitation=#{invited_user.invitation_token}"
```
StupidCodeFactory: maybe the route helper would be a bit nicer`sign_out_before_accepting_invitation_path(invited_user, invitation: invited_user.invitation_token)`

psd-web/spec/models/user_spec.rb
```diff
@@ -1,6 +1,71 @@
 require "rails_helper"
 
 RSpec.describe User do
+  describe "validations" do
+    context "on registration completion" do
+      it "validates the presence of the mobile number" do
+        user = build(:user, mobile_number: "")
+        expect(user.valid?(:registration_completion)).to eq false
```
StupidCodeFactory: RSpec provides a nice helper for this:
```suggestion
        expect(user).to be_valid(:registration_completion)
```

psd-web/spec/models/user_spec.rb
```diff
@@ -233,4 +302,26 @@
       end
     end
   end
+
+  describe "#invitation_expired?" do
+    it "returns false when the user has not been invited" do
+      user = build_stubbed(:user, invited_at: nil)
+      expect(user.invitation_expired?).to be_falsey
```
StupidCodeFactory: why falsey? predicate method should always be returning a boolean.

psd-web/app/views/users/complete_registration.html.erb
```diff
@@ -0,0 +1,38 @@
+<% title = "Create an account" %>
```
frankieroberto: We should prefix this with "Error: " if there are validation errors present.

psd-web/spec/requests/user_completes_registration_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+      before do
+        patch user_path(user.id),
+              params: {
+                        invitation: user.invitation_token,
+                        user: {
+                          name: "Foo Bar",
+                          password: "foobarnoteasyatall1234!",
+                          mobile_number: "07235671232"
+                        }
+                      }
```
StupidCodeFactory: the indentation is a bit off

psd-web/spec/requests/user_completes_registration_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+        user.reload
+      end
+
+      it "sets the activated flag on the user" do
+        expect(user.account_activated?).to be true
```
StupidCodeFactory:
```suggestion
        expect(user).to be_account_activated
```

psd-web/spec/models/user_spec.rb
scruti: Good catch, fixing the method to stick to the convention.

psd-web/app/models/user.rb
scruti: I think this may seem redundant from dev perspective but be an important distinction from UX perspective. Checking it with Ed.

psd-web/app/controllers/users_controller.rb
```diff
@@ -0,0 +1,46 @@
+class UsersController < ApplicationController
+  skip_before_action :authenticate_user!
+  skip_before_action :has_accepted_declaration
+  skip_before_action :has_viewed_introduction
+
+  def complete_registration
+    @user = User.find(params[:id])
+
+    return render :signed_in_as_another_user if user_signed_in? && !signed_in_as?(@user)
+    return redirect_to(root_path) if signed_in_as?(@user)
```
slorek: We might want to check if some content is required in this case.

slorek: We might want to check if some content is required in this case.  /cc @edwardhorsford @1AmyMcNichol

psd-web/app/controllers/users_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    return render(:expired_token) if @user.invitation_expired?
+    return redirect_to("/sign-in") if @user.account_activated?
```
slorek: Again we should check if we want some content for this scenario. I'm told that some users bookmark this page or use the e-mail link even after registering. /cc @edwardhorsford @1AmyMcNichol

psd-web/app/controllers/users_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    return (render "errors/not_found", status: :not_found) if @user.invitation_token != params[:invitation]
```
slorek: We should probably (perhaps as a separate ticket) consider brute force protection for the token, i.e. resetting the token after X attempts.

psd-web/app/controllers/users_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def sign_out_before_accepting_invitation
+    sign_out
+    redirect_to complete_registration_user_path(params[:id], invitation: params[:invitation])
+  end
+
+  def update
+    @user = User.find(params[:id])
+    return render("errors/forbidden", status: :forbidden) unless params[:invitation] == @user.invitation_token
```
slorek: Again we should probably (perhaps as a separate ticket) consider brute force protection for the token.

psd-web/app/controllers/users_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    @user.assign_attributes(new_user_attributes.merge(account_activated: true))
+
+    if @user.save(context: :registration_completion)
+      sign_in :user, @user
+      redirect_to root_path
+    else
```
slorek: We could improve this by removing the `else` e.g. 

```ruby
if @user.save(context: :registration_completion)
  sign_in :user, @user
  return redirect_to root_path
end

render :complete_registration
```

psd-web/app/models/user.rb
```diff
@@ -24,6 +42,11 @@ def self.create_and_send_invite!(email_address, team, inviting_user)
       organisation: team.organisation,
       invitation_token: SecureRandom.hex(15)
     )
+
+    # TODO: remove this once we’ve updated the application to no
+    # longer depend upon this role.
+    user.user_roles.create!(name: "psd_user")
```
slorek: I think we need to consider when to add the `opss_user` role as well.

psd-web/app/models/user.rb
```diff
@@ -11,6 +14,21 @@ class User < ApplicationRecord
 
   validates :id, presence: true, uuid: true
 
+
+  with_options on: :registration_completion do |registration_completion|
+    registration_completion.validates :mobile_number, presence: true
+    registration_completion.validate :validate_mobile_number
```
slorek:
```suggestion
    registration_completion.validate :validate_mobile_number, unless: -> { mobile_number.blank? }
```

psd-web/app/models/user.rb
```diff
@@ -146,8 +169,31 @@ def has_viewed_introduction!
...
+    File.foreach(COMMON_PASSWORDS_FILE_PATH, chomp: true) do |common_password|
+      if common_password == password
+        errors.add(:password, I18n.t(:too_common, scope: %i[activerecord errors models user attributes password]))
+        break
+      end
+    end
+  end
+
+  def validate_mobile_number
+    return if mobile_number.blank?
```
slorek: ✂️

psd-web/app/views/users/complete_registration.html.erb
```diff
@@ -0,0 +1,38 @@
...
+<% content_for :page_title, title %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+
+    <%= govukErrorSummary(titleText: "There is a problem", errorList: @user.errors.messages.map { |key, value|
```
slorek: @StupidCodeFactory created a helper for this: https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/273/files#diff-3a9651b5997ab974e9d52961cd84ed6bR8 - we should use a consistent approach (I don't have a preference for which one).

psd-web/app/views/users/complete_registration.html.erb
```diff
@@ -0,0 +1,38 @@
...
+        value: @user.name
+        %>
+
+      <%= render "form_components/govuk_input", key: :mobile_number, form: form,
+        classes: "app-!-max-width-two-thirds",
+        label: { text: "Mobile number" },
+        value: @user.mobile_number
+        %>
+
+      <%= render "form_components/govuk_input", key: :password, form: form,
```
slorek: Again @StupidCodeFactory created helpers: https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/273/files#diff-4e53a7fd9506bf807e36609e587b29ffR16 - let's align the approach between these PRs.

psd-web/app/models/user.rb
frankieroberto: Yeah, they're different error messages, so having these two different validations is deliberate. We also make sure you only ever see the most relevant message.

psd-web/spec/features/creating_an_account_from_an_invitation_spec.rb
frankieroberto: @StupidCodeFactory @slorek we previously agreed a compromise of using the full paths in feature specs, but the helpers in request specs...  (rationale being that feature specs should be unaware of the implementation, and only test based upon the exposed user-interface).

psd-web/app/controllers/users_controller.rb
frankieroberto: @edwardhorsford suggests this is fine - they're using the invitation link just to get to the service, so redirecting them to the homepage is fine.

psd-web/app/controllers/users_controller.rb
frankieroberto: @edwardhorsford suggests that no content is required here – although we could redirect to homepage instead (which has a sign in button on it)?

psd-web/app/controllers/users_controller.rb
frankieroberto: Agreed, but separate ticket? Risk feels fairly low in the short-term.

psd-web/app/models/user.rb
frankieroberto: Yep, we forgot about that... 🤦‍♂

psd-web/app/models/user.rb
frankieroberto: Nice! We had `registration_completion.validate :validate_mobile_number, allow_blank: true` but that doesn't work for custom validations. `unless` should though.

psd-web/app/controllers/users_controller.rb
scruti: Not so sure on how avoiding the `else` in this case would result in an improvement.

I'm all in for early returns when they cut the natural happy path of the method and especially when they help to avoid nested conditions.
But in this case, `render :complete_registration` gets reached when we try to update the user but the validations fail. So we would have an `update` method that has a last line that only gets e where when the `update` fails. 

May be a question of style preferences but for me that is more confusing that 
```ruby
if @user.save
  render :happy_times
else
  render :unhappy_times
end
```

If we want to use the early return approach I would go for inverting the order:

```ruby
return render :complete_registration unless @user.save(context: :registration_completion)

sign_in :user, @user
return redirect_to root_path
```

In this way we have an early return when we weren't able to save the user, leaving the last lines for the default/expected behaviour.

I still prefer the simple `if/else` in this case though :sweat_smile:

edwardhorsford:
Some suggested content for the expired invitation:

```markdown

# This invitation has expired

Invitations are valid for [days valid] days. Ask the person who invited you to resend the invite or [email OPSS] for support.

```

psd-web/app/controllers/users_controller.rb
slorek: It's definitely subjective, but one we should be consistent with. Good point and I agree with the "escape early" principle, but I'd avoid mutations within appended conditions due to cognitive complexity. So maybe:

```ruby
unless @user.save(context: :registration_completion)
  return render :complete_registration
end

sign_in :user, @user
return redirect_to root_path
```

psd-web/app/views/users/complete_registration.html.erb
frankieroberto: Done in 6ab235d8ecc7549b996d910dcb572a789a11b0e6

psd-web/app/controllers/users_controller.rb
frankieroberto: Have to say, I think I prefer the `if/else` version too – early returns are good for exceptional cases, but rendering a page with validation errors feels like more of an equal branch with the the happy path of successfully saving...

psd-web/db/migrate/20200220142847_add_mobile_number.rb
```diff
@@ -0,0 +1,5 @@
+class AddMobileNumber < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :mobile_number, :text
```
slawosz: Why text? String would be enough.

psd-web/spec/models/user_spec.rb
```diff
@@ -1,6 +1,72 @@
 require "rails_helper"
 
 RSpec.describe User do
+  describe "validations" do
```
slawosz: As the `users_spec.rb` grows, how about splitting it to many files, depending for example on top level `describe` block.

psd-web/db/migrate/20200220142847_add_mobile_number.rb
frankieroberto: In Postgres, there is no difference between `varchar` (which is what `:string` maps to)  and `text` in terms of performance, so it makes no difference... 🤷‍♂  (see [reference](https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_varchar.28n.29_by_default))

Rails _used_ to map `:string` to `varchar(255)`, and I once got bitten by creating a `url` column as a `:string`, and then getting an error when a looong URL was entered several months later. Converting the `varchar(255)` column to a `text` column required a full table lock, which meant 10 mins or so of downtime, and so since then I've always used `:text`...  Looks like [this got changed in Rails](https://github.com/rails/rails/commit/f4226c3ab6651f6871e02f3c6754c29ab155b938) a while ago though.

psd-web/app/controllers/users_controller.rb
scruti: Added the [ticket in our backlog](https://trello.com/c/V5sTVLCh/384-provide-brute-force-protection-for-user-registration-over-invitation-token).

psd-web/app/controllers/users_controller.rb
scruti: Added the [ticket in our backlog](https://trello.com/c/V5sTVLCh/384-provide-brute-force-protection-for-user-registration-over-invitation-token).

nasirkhanbeis:
@frankieroberto @scruti  can we also implement/fix below scenario: 
Scenario: account setup complete -invite link is used
Given user completed the invite setup
When user click on invite link after account setup
Then user should be taken to the homepage
currently it is redirecting to sign-in page
eg, used invite link
https://psd-pr-294.london.cloudapps.digital/users/2cff1a97-15fa-497c-a79c-1f4dd237694c/complete-registration?invitation=a1a8736436098188ea318f0cd081bc

psd-web/spec/models/user_spec.rb
scruti: When unit testing, I usually prefer to keep it together matching classes with test files. And, if the size becomes a problem... maybe is a cry for help from the class having too many responsibilities? (and a good opportunity for logic extraction from the class itself).

nasirkhanbeis:
weird edge case 
when I enter password as TestPassword123 I see weird error as below
image

Expected to show error message as "Choose a less frequently used password"

nasirkhanbeis:
expecting minimum validation around phone number field to accept  numbers starting with +447 or 07xxxx rather not to accept any vague format 
image

psd-web/app/controllers/users_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  skip_before_action :has_accepted_declaration
+  skip_before_action :has_viewed_introduction
+
+  def complete_registration
+    @user = User.find(params[:id])
+
+    return render :signed_in_as_another_user if user_signed_in? && !signed_in_as?(@user)
+    # Some users will bookmark the invitation URL received on the email and will
+    # become their link to "PSD". Hence redirecting them to the root page.
+    return redirect_to(root_path) if signed_in_as?(@user) || @user.account_activated?
```
scruti: Now that I put together both conditions that will result in the visitor being redirected to the root path... I wonder if there is any value in keeping `signed_in_as?(@user)` here.
I assume that the visitor being  "signed in as the user" implies the user "account being activated".

So... could we just do `return redirect_to(root_path) if @user.account_activated?` given that would catch the "user being already signed in" condition?

scruti:
Addressed multiple comments.
Pending:
- Check using `< DeviseController` for registration. Raised by @StupidCodeFactory on [this Slack Thread](https://safetyandstandards.slack.com/archives/CN6DQV6GL/p1582646661000600)

psd-web/spec/models/user_spec.rb
scruti: @slawosz ended extracting the custom validations (and their tests) outside the User model to their own Validator classes.

psd-web/app/controllers/users_controller.rb
nasirkhanbeis: can we also add not signed-in condition? so they can be redirected to unauthenticated home page ?

nasirkhanbeis:
@scruti 
Also noticed, user couldn't login with given username/pwd after they completed the account setup, I guess this will work after we merge with sign-in code changes? or storing new pwd in different db?

psd-web/app/controllers/users_controller.rb
scruti: @nasirkhanbeis users not being signed in is the expected behaviour when reaching the registration page.

nasirkhanbeis: yeap works fine now!

LMallows:
When a user has finished setting up their account and logs in for the first time they are presented with a declaration screen. On this page when a user clicks the terms and conditions, privacy policy or footer links it is redirecting them back to the declaration page. They should instead be taken to these pages in anew tab. 

In addition, when a user has agreed the declaration if they are a TS user they are presented with 4 intro slides. When they are on these pages they should be able to go any of the links in the footer.  On this PR they are being redirected to the slide instead of correctly taken to the page in question. 

This is not the current behaviour in production when it goes to whatever the link is correctly.

LMallows:
When a user has finished setting up their account and logs in for the first time they are presented with a declaration screen. On this page when a user clicks the terms and conditions, privacy policy or footer links it is redirecting them back to the declaration page. They should instead be taken to these pages in anew tab. 

In addition, when a user has agreed the declaration if they are a TS user they are presented with 4 intro slides. When they are on these pages they should be able to go any of the links in the footer.  On this PR they are being redirected to the slide instead of correctly taken to the page in question. 

This is not the current behaviour in production when it goes to whatever the link is correctly.

nasirkhanbeis:
@LMallows @frankieroberto I suspect this will be fixed by itself when we merge into feature branch as I see it's working fine in feature branch. 
I tested in this review app which has feature branch code https://psd-pr-306.london.cloudapps.digital/
try logging with user testuser16@test.com pwd: loadtest

nasirkhanbeis:
@LMallows @frankieroberto I suspect this will be fixed by itself when we merge into feature branch as I see it's working fine in feature branch. 
I tested in this review app which has feature branch code https://psd-pr-306.london.cloudapps.digital/
try logging with user testuser16@test.com pwd: loadtest

scruti:
Will definitely do!

PR #296, opened 2020-02-20
redirect to root path on old bookmarked url

StupidCodeFactory:
## Description
Some user have bookmarks containing old redirect uri that do not exists anymore.
let's redirect those users to the home page

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/config/routes.rb
+++ b/psd-web/config/routes.rb
@@ -179,6 +179,6 @@ Rails.application.routes.draw do
   unauthenticated do
     root to: "homepage#show"
   end
-  match "/sessions/signin", to: redirect("/"), via: [:get, :post]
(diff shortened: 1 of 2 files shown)


psd-web/config/routes.rb
```diff
@@ -179,5 +179,6 @@
   unauthenticated do
     root to: "homepage#show"
   end
+  match "/sessions/signin", to: redirect("/"), via: %i[get post]
```
slorek:
```suggestion
  # Handle old post-login redirect URIs from previous implementation which are still bookmarked
  match "/sessions/signin", to: redirect("/"), via: %i[get post]
```

PR #236, opened 2020-01-30
WIP: New sign in / account creation templates

frankieroberto:
This add some new templates needed by the switch from an external login system to one within the application itself.

Not to be merged - this is just for early previewing of the design and content.

### Reset password flow
https://psd-pr-236.london.cloudapps.digital/sign-in
https://psd-pr-236.london.cloudapps.digital/reset-password
https://psd-pr-236.london.cloudapps.digital/check-your-email
https://psd-pr-236.london.cloudapps.digital/two-factor
https://psd-pr-236.london.cloudapps.digital/new-password

Errors: 
https://psd-pr-236.london.cloudapps.digital/link-expired
https://psd-pr-236.london.cloudapps.digital/text-not-received

TODO: 
* [x] error messages
* [ ] error screen for 2FA code expired (with button to resend)?

### Sign in flow
https://psd-pr-236.london.cloudapps.digital/sign-in
https://psd-pr-236.london.cloudapps.digital/two-factor

Errors:
https://psd-pr-236.london.cloudapps.digital/text-not-received

TODOs:
* [x] error messages

### Registration flow
https://psd-pr-236.london.cloudapps.digital/create-account
https://psd-pr-236.london.cloudapps.digital/two-factor

Errors:
* https://psd-pr-236.london.cloudapps.digital/invite-expired
* https://psd-pr-236.london.cloudapps.digital/text-not-received-account-creation (variant on other template?)

TODOs:
* [x] Invite expired page
* [x] Code not received page allows users to enter a new mobile phone number somehow...
* [x] error messages

PR #298, opened 2020-02-20
Devise reset password flow

StupidCodeFactory:
This adds the functionality to allow users to reset their password, by entering their e-mail address and then receiving a link via e-mail which is valid for a single use, for a short amount of time.

The feature _doesn’t_ yet require 2-factor-authentication, as that will be added to the feature branch via  later Pull Request.

It also doesn’t yet guard against the edge-case of users requesting a password reset when they haven't yet created their initial account from the invitation e-mail. This is specified a future task here: https://trello.com/c/7CHn7yzy/391-reset-password-for-people-who-have-received-an-invite-but-have-not-completed-account-set-up

This feature relies heavily upon the default devise `PasswordsController`, although with some customisations implemented via a custom controller which inherits, and calls `super`.

One key difference is that we are validating the reset token _before_ rendering the new password form, rather than after it is submitted. This is to provide more immediately feedback to the user.  There may be some security considerations for this, although they might be tackled by the future task to guard against brute force attacks: https://trello.com/c/KPrSkHF4/383-5-prevent-brute-forcing-of-passwords-at-sign-in

The new password form should be using the same validations as when creating a password from the original invitation – ie checking the password length and that it’s not one of the most commonly-used passwords.

https://trello.com/c/njmHsb3W/336-5-reset-password-flow

### Reset password page

Screen Shot 2020-02-26 at 17 23 39

### Check your email page

Screenshot 2020-03-02 at 16 21 41

### Enter new password page

Screenshot 2020-03-02 at 16 22 05


frankieroberto:
@StupidCodeFactory could you change the base branch for this PR to be `keycloak-replacement`? I think `devise-db-authenticatable` has already been squash-merged into the feature branch.

edwardhorsford:
Pages looking good.

Some comments from me:

* Please can the reset page please be loaded on `/reset-password` as per [this line](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/236/files#diff-3a3a48bab79b405125d4441554348ef2R19).
* Please can the check your email page be loaded on `/check-your-email` as per [this line](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/236/files#diff-3a3a48bab79b405125d4441554348ef2R19).

psd-web/app/controllers/users/passwords_controller.rb
```diff
@@ -0,0 +1,31 @@
+module Users
+  class PasswordsController < Devise::PasswordsController
```
slawosz: How about `ResetPasswordController`?

psd-web/app/forms/concerns/email_form_validation.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  included do
+    attribute :email
+
+    validates :email,
+              email: {
+                message: I18n.t(:wrong_email_or_password, scope: "sign_in_form.email"),
+                if: ->(sign_in_form) { sign_in_form.email.present? }
+              }
+    validates_presence_of :email, message: "Enter your email address"
```
slorek:
```suggestion
    validates_presence_of :email, message: I18n.t(:enter_email_address)
```

psd-web/config/initializers/notify_mailer.rb
```diff
@@ -1 +1 @@
-ActionMailer::Base.add_delivery_method :govuk_notify, GovukNotifyRails::Delivery, api_key: ENV["NOTIFY_API_KEY"]
+ActionMailer::Base.add_delivery_method :govuk_notify, GovukNotifyRails::Delivery, api_key: ENV["NOTIFY_API_KEY"] || "#{SecureRandom.uuid}.#{SecureRandom.uuid}"
```
slorek: Please change: This could be quite misleading and I don't see how it's useful. You can use `with_stubbed_mailer` if you need to test email delivery.

frankieroberto:
@StupidCodeFactory could you remove the flash message from the redirect to the root path if you  try to visit the password reset page whilst already signed in?

ie avoiding this:

Screenshot 2020-02-27 at 10 33 25

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+        email_address: user.email,
+        template_id: NotifyMailer::TEMPLATES[:reset_password_instruction],
+        reference: "Password reset",
+        personalisation: {
+          name: user.name,
+          edit_user_password_url_token: edit_user_password_url_token
+        }
+      }
+
+      stub_request(:post, "https://api.notifications.service.gov.uk/v2/notifications/email")
```
slorek: Suggestion: I think we should just use the `with_stubbed_mailer` shared context, since this line is stubbing low level functionality within a third party gem which is subject to change and already tested.

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
+require "rails_helper"
+
+RSpec.describe "Password reset management", :with_test_queue_adpater do
```
slorek: Suggestion: Just use the inline test adapter (default). I don't think we actually need to switch to test adapter here as we're not doing any assertions against the queue.

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      .to receive(:generate)
+            .with(User, :reset_password_token).and_return(reset_token)
+    _raw, enc = reset_token
+    user.update!(reset_password_token: enc)
+
+    visit new_user_session_path
+
+    click_link "Forgot your password?"
+
+    perform_enqueued_jobs do
```
slorek: ✂️ 
```suggestion
```

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+          click_on "Continue"
+
+          expect(page).to have_css("h2#error-summary-title", text: "There is a problem")
+          expect(page).to have_link("Password cannot be blank", href: "#password")
+        end
+      end
+    end
+  end
+
+  context "with and invalid token" do
```
slorek:
```suggestion
  context "with an expired token" do
```

slorek: Please change: this test is for an expired token, we should test against an invalid one as well.

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      click_on "Sign in to your account"
+
+      fill_in "Email address", with: user.email
+      fill_in "Password", with: "a_new_password"
+      click_on "Continue"
+
+      expect(page).to have_css("h1", text: "Declaration")
+    end
+
+    context "when the password does not fit the critiera" do
```
slorek:
```suggestion
    context "when the password does not fit the criteria" do
```

psd-web/spec/models/user_spec.rb
```diff
@@ -233,4 +233,19 @@
...
+    subject { create(:user) }
+
+    it "enqueues a job posting email to notify", :with_test_queue_adpater do
+      expect { subject.send_reset_password_instructions }
+        .to enqueue_job
+              .on_queue("psd")
+              .at(:no_wait)
+              .with do |user, token|
+        expect(user).to be user
+        expect(token).to be token
```
StupidCodeFactory: oops let me this doesn't test anything. Let me ✂️ this
.

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      fill_in "Email address", with: "not_an_email"
+      click_on "Send email"
+
+      expect(page).to have_css("h2#error-summary-title", text: "There is a problem")
+      expect(page).to have_link("Enter your email address in the correct format, like name@example.com", href: "#email")
+    end
+  end
+
+  context "with a valid token" do
+    it "does not allow you to reset you pasword" do
```
slorek:
```suggestion
    it "allows you to reset your password" do
```

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      stub_request(:post, "https://api.notifications.service.gov.uk/v2/notifications/email")
+        .with(body: body.to_json).to_return(status: 200, body: {}.to_json, headers: {})
+
+      fill_in "Email address", with: user.email
+      click_on "Send email"
+    end
+  end
+
+  context "when entering an invalid email" do
+    it "does not allow you to reset you pasword" do
```
slorek:
```suggestion
    it "does not allow you to reset your password" do
```

psd-web/spec/features/password_reset_spec.rb
slorek:
```suggestion
RSpec.describe "Resetting your password", :with_test_queue_adpater do
```

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      visit new_user_session_path
```
slorek: Please change: Use full path. Then verify we are on the expected page (check the `h1`).

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+
+      click_link "Forgot your password?"
+      fill_in "Email address", with: "not_an_email"
```
slorek: Suggestion: Check we are on the correct page before attempting to fill in the form.

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
+require "rails_helper"
+
+RSpec.describe "Password reset management", :with_test_queue_adpater do
+  let(:user)                         { create(:user) }
+  let!(:reset_token)                 { Devise.token_generator.generate(User, :reset_password_token) }
+  let(:edit_user_password_url_token) { edit_user_password_url(reset_password_token: reset_token.first) }
+
+
+  def send_reset_password
```
slorek:
```suggestion
  def request_password_reset
```
Seems to describe more accurately what is happening.

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+    allow(Devise.token_generator)
+      .to receive(:generate)
+            .with(User, :reset_password_token).and_return(reset_token)
+    _raw, enc = reset_token
+    user.update!(reset_password_token: enc)
+
+    visit new_user_session_path
```
slorek: Please change: Use full path. then check we are on the expected page.

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+        personalisation: {
+          name: user.name,
+          edit_user_password_url_token: edit_user_password_url_token
+        }
+      }
+
+      stub_request(:post, "https://api.notifications.service.gov.uk/v2/notifications/email")
+        .with(body: body.to_json).to_return(status: 200, body: {}.to_json, headers: {})
+
+      fill_in "Email address", with: user.email
```
slorek: Suggestion: Check we are on the expected page before attempting to fill in the form.

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      expect(page).to have_link("Enter your email address in the correct format, like name@example.com", href: "#email")
+    end
+  end
+
+  context "with a valid token" do
+    it "does not allow you to reset you pasword" do
+      send_reset_password
+      expect(page).to have_css("p.govuk-body", text: "Click the link in the email to reset your password.")
+
+      visit edit_user_password_url_token
```
slorek: Please change: Use full path and then check we are on the expected page.

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      fill_in "Password", with: "a_new_password"
+      click_on "Continue"
+
+      expect(page).to have_css("h1", text: "Declaration")
+    end
+
+    context "when the password does not fit the critiera" do
+      context "when the password is too short" do
+        let(:password) { "as" }
+        it "does not allow you to reset you pasword" do
```
slorek:
```suggestion
        it "does not allow you to reset your password" do
```

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+          send_reset_password
+
+          visit edit_user_password_url_token
```
slorek: Please change: Use full path.

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+
+          expect(page).to have_css("h2#error-summary-title", text: "There is a problem")
+          expect(page).to have_link("Password is too short (minimum is 8 characters)", href: "#password")
+        end
+      end
+
+      context "when the password is too short" do
+        let(:password) { "" }
+
+        it "does not allow you to reset you pasword" do
```
slorek:
```suggestion
        it "does not allow you to reset your password" do
```

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+          send_reset_password
+
+          visit edit_user_password_url_token
```
slorek: Please change: Use full paths.

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+
+          expect(page).to have_css("h2#error-summary-title", text: "There is a problem")
+          expect(page).to have_link("Password cannot be blank", href: "#password")
+        end
+      end
+    end
+  end
+
+  context "with and invalid token" do
+    it "does not allow you to reset you pasword" do
```
slorek:
```suggestion
    it "does not allow you to reset your password" do
```

psd-web/app/forms/sign_in_form.rb
```diff
@@ -1,15 +1,8 @@
 class SignInForm
   include ActiveModel::Model
   include ActiveModel::Attributes
+  include EmailFormValidation
```
scruti: Suggestion: Do we need this extraction?
A form class with barely 15 loc related to two attributes and their validations seemed simple enough. 
Maybe is being reused somewhere else but I didn't find it. 
I wonder if the indirection penalty is worth the abstraction in this case. But, as the title says... just a suggestion.

psd-web/app/mailers/notify_mailer.rb
```diff
@@ -1,14 +1,25 @@
...
+      reset_password_instruction: "cea1bb37-1d1c-4965-8999-6008d707b981",
+      investigation_updated: "10a5c3a6-9cc7-4edb-9536-37605e2c15ba",
+      investigation_created: "6da8e1d5-eb4d-4f9a-9c3c-948ef57d6136",
+      alert: "47fb7df9-2370-4307-9f86-69455597cdc1",
+      user_added_to_team: "e3b2bbf5-3002-49fb-adb5-ad18e483c7e4",
+      welcome: "035876e3-5b97-4b4c-9bd5-c504b5158a85",
+      invitation: "7b80a680-f8b3-4032-982d-2a3a662b611a"
     }.freeze
 
+  def reset_password_instruction(user, token)
```
scruti: Nitpick: Is the reset password an individual instruction or "instructions?"

psd-web/app/models/user.rb
```diff
@@ -153,6 +153,10 @@ def has_viewed_introduction!
 
 private
 
+  def send_reset_password_instructions_notification(token)
+    SendResetPasswordInstructions.perform_later(self, token)
+  end
```
scruti: :question:  What is calling this method? Devise under the hood?

psd-web/app/forms/sign_in_form.rb
scruti: Forget about my previous comment: Just saw  it reused in 2 different forms. :+1:  on it.

psd-web/app/controllers/users/passwords_controller.rb
StupidCodeFactory: Because it’s a copy of the devise controller. It’s also the resource that’s being edited (`super`). Best to keep with how it’s done in Devise, it’s less confusing.

psd-web/config/initializers/notify_mailer.rb
slawosz: Actually, thats very good point. However, lets do this:
```
ActionMailer::Base.add_delivery_method :govuk_notify, GovukNotifyRails::Delivery, api_key: ENV.fetch("NOTIFY_API_KEY")
```
This way, if this var is not set, it will fail on startup. Please add this to `env.development.example` and `.env.test.example`

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+
+      expect(page).to have_css("h2#error-summary-title", text: "There is a problem")
+      expect(page).to have_link("Enter your email address in the correct format, like name@example.com", href: "#email")
+    end
+  end
+
+  context "with a valid token" do
+    it "does not allow you to reset you pasword" do
+      send_reset_password
+      expect(page).to have_css("p.govuk-body", text: "Click the link in the email to reset your password.")
```
frankieroberto: I think we could probably remove the assertion about the css here:

```suggestion
      expect(page).to have_content("Click the link in the email to reset your password.")
```

psd-web/spec/features/password_reset_spec.rb
scruti: Suggestion: Not introduced in this PR but could we fix the typo in `:with_test_queue_adpater` for `:with_test_queue_adapter` before it keeps expanding around the codebase :sweat_smile: ?

StupidCodeFactory: So I chose to do this to be very clear in this tests that were expecting a jobs to be asynchronously processed and what is being passed to the job in the spirit of full stack test and keep it descriptive.

psd-web/spec/features/password_reset_spec.rb
StupidCodeFactory: This check the data flows with integrity and that ultimately `NotifyMailer` does call out to `GovukNotifyMailer`'s api correctly.
I really don't think the tradeoff versus smaller test is better.

psd-web/config/initializers/notify_mailer.rb
StupidCodeFactory: @slawosz actually I don't think we can do that. I would have loved to, but I don't think we have access to `NOTIFY_API_KEY` at server build time. So this would throw an exception.

psd-web/app/models/user.rb
```diff
@@ -153,6 +153,10 @@ def has_viewed_introduction!
 
 private
 
+  def send_reset_password_instructions_notification(token)
```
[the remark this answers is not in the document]
StupidCodeFactory: @frankieroberto happy to resolve your comment?

psd-web/app/models/user.rb
StupidCodeFactory: @scruti happy to resolve your comment?

psd-web/spec/features/password_reset_spec.rb
StupidCodeFactory: We've used `have_css` everywhere so far, it also a documents directly how this piece of text is displayed. I find is easier to reason about.

psd-web/spec/features/password_reset_spec.rb
StupidCodeFactory: @slorek is this satisfactory enough?
https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/298/files#diff-6e57542a5ecd3023d04f99204f94bb47R6

psd-web/app/models/user.rb
scruti: Oh sure was just curiosity. Thanks for clarifying it on the code.

slorek:
This error is incorrect (entering unknown email address). It should not confirm whether the account exists and display the  'check your email' message anyway.
Screenshot 2020-03-02 at 09 31 34

slorek:
Going to the reset page with an invalid token also shows the form before the 'expired' message. Example: https://psd-pr-298.london.cloudapps.digital/password/edit?reset_password_token=test

slorek:
Entering 'Password123' is accepted but should be rejected as it is in the common passwords list file.

slorek:
A user who has been invited but not yet set up their account is able to receive the forgot email link and reset their password. Submitting the new password form triggers an infinite redirect loop.

psd-web/spec/features/password_reset_spec.rb
frankieroberto: @slorek this does contain a full path, but it's set up in the `let` at the top of the page. Perhaps this isn't obvious though – maybe it'd be better to defactor this and just duplicate the full path in place?

.github/workflows/rspec.yml
```diff
@@ -16,7 +16,7 @@ env:
   ANTIVIRUS_USERNAME: av
   ANTIVIRUS_PASSWORD: password
   PSD_HOST: example.com
-
+  NOTIFY_API_KEY: 6eb625ef-7df2-45d7-ab42-1bc528823f55.6d0fc2cf-633f-42ba-97eb-63c96d254aca
```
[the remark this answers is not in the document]
slorek: I think we should remove this. it will cause confusion. We should be able to get the tests to pass without supplying a fake key.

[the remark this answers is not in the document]
frankieroberto: Done in 1d2220ec09386093fc59ec9d45c36e32e93e6c70 – using the stubbed mailer instead (which is what we currently do in other feature specs).

psd-web/app/controllers/users/passwords_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+    skip_before_action :assert_reset_token_passed, only: :edit
```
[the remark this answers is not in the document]
slorek: I think we should keep `super` so that we're not reimplementing functionality which might change.

psd-web/app/jobs/send_reset_password_instructions.rb
```diff
@@ -0,0 +1,5 @@
+class SendResetPasswordInstructions < ApplicationJob
+  def perform(user, token)
+    NotifyMailer.reset_password_instructions(user, token).deliver_now
+  end
+end
```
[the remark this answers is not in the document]
slorek: Agreed, we can probably just call `deliver_later` there.

slorek:
Review app deploy was successful several mins before timeout occurred. Manual testing was performed there and confirmed all identified issues are resolved.

PR #308, opened 2020-02-25
Remove keycloak syncing

slawosz:
Removes creating `User`, `Team` and `Organisation` from keycloak. 
Only import code from Keycloak left is in `KeycloakConnector`. This code have to be run once, when service is stopped for maintenance to update passwords and other data from keycloak.

To be merged as last PR in the branch, after stopping keycloak. Reason for it is to be able to notice any problems with current keycloak teams and be able to have data from keycloak on new review app dbs


slorek:
I think you'll also need to remove the related specs for the removed code, as well as the Keycloak stubbing in Rspec and the user/team/org setup methods in Minitest.

slorek:
I have searched the code and I think we need to refactor the tests in `psd-web/test/controllers/teams_controller_test.rb` as they are using/stubbing the Keycloak clients. We might choose to do this separately. Apart from that it should be good to go.

PR #311, opened 2020-02-26
2fa flow

StupidCodeFactory:
## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


psd-web/app/services/govuk_notify.rb
```diff
@@ -0,0 +1,29 @@
...
+  TEMPLATES = {
+    send_otp_code: "33517ff-2a88-4f6e-b855-c550268ce08a"
+  }
+
+  attr_reader :pool
+
+  def initialize
+    self.pool = ConnectionPool.new(size: ENV.fetch("RAILS_MAX_THREADS", 5), timeout: 1) do
+      Notifications::Client.new(Rails.configuration.notify_api_key)
+    end
```
StupidCodeFactory: Connection pooling to avoid blocking the current thread on IO and allow other threads to run, sadly `notification-ruby` is not implemented in such a way that this pool will also keep the connection open to the remote server (this greatly reduce the dreaded EOF or ERNOENT when the connction drops in the middle of the request).
I'll probably open a PR to notification-ruby to allow reusing open connection.

scruti:
Extracted a branch to continue with this work without the dependency on the "devise-reset-password-flow" and re-pointing the 2FA work against `keycloak-replacement` branch. 
Current PR: https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/333

PR #314, opened 2020-02-27
Bugfix: don't check whether user has viewed introduction on help pages

frankieroberto:
Users should be able to view the "terms and conditions" and "privacy notice" pages at any point, regardless of whether they have accepted the declaration or not, are signed in or not, or have viewed the introduction slides or not.

This is particularly important as the declaration page links to the terms and conditions and privacy notice pages, which users are agreeing to by accepting the declaration. However currently non-OPSS users are unable to view these at this pont, as the pages are redirecting to the introduction slides, which then redirect back to the declaration page.


slorek:
Consider merging these tests with the existing ones in https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/spec/features/help_pages_spec.rb

LMallows:
Working correctly on both declaration and intro slides. :)

PR #315, opened 2020-02-27
Fix Readme broken links

scruti:
While reading our Readme I noticed several links were broken, so decided to fix them.

Also recorded in the Readme the decision to use ERB instead of Slim.


scruti:
Ammend & Push with no changes to re-trigger Github checks as Github was down and many checks didn't run and were blocking the PR merge.

PR #319, opened 2020-02-28
Check whether account has been setup, not activated

frankieroberto:
The "activated" flag should only be set after the user has accepted the declaration, so we need a different check within the account setup flow to check that the user hasn't already set up a password and phone number.

### DIFF
--- a/psd-web/app/models/user.rb
+++ b/psd-web/app/models/user.rb
@@ -151,6 +151,10 @@ class User < ApplicationRecord
     has_role? :team_admin
   end
 
+  def has_setup_account?
+    encrypted_password.presence && name.presence && mobile_number.presence?
+  end
+
   def self.get_assignees(except: [])
     user_ids_to_exclude = Array(except).collect(&:id)
     self.activated.where.not(id: user_ids_to_exclude).eager_load(:organisation, :teams)
--- a/psd-web/app/models/user.rb
+++ b/psd-web/app/models/user.rb
@@ -152,7 +152,7 @@ class User < ApplicationRecord
   end
 
   def has_completed_registration?
-    encrypted_password.presence && name.presence && mobile_number.presence?
+    encrypted_password.present? && name.present? && mobile_number.present?
   end
 
   def self.get_assignees(except: [])
--- a/psd-web/app/models/user.rb
+++ b/psd-web/app/models/user.rb
@@ -151,7 +151,7 @@ class User < ApplicationRecord
     has_role? :team_admin
   end
 
-  def has_setup_account?
+  def has_completed_registration?
     encrypted_password.presence && name.presence && mobile_number.presence?
   end
 
--- a/psd-web/spec/factories/users.rb
+++ b/psd-web/spec/factories/users.rb
@@ -10,6 +10,7 @@ FactoryBot.define do
     has_viewed_introduction { false }
     account_activated { false }
     hash_iterations { 27_500 }
+    mobile_number { "07700 900 982" }
 
     transient do
       roles { [:psd_user] }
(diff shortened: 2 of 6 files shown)


psd-web/app/models/user.rb
```diff
@@ -151,6 +151,10 @@ def is_team_admin?
     has_role? :team_admin
   end
 
+  def has_setup_account?
```
scruti:
```suggestion
  def has_completed_registration?
```
Just to align the language through the service and avoid confusions like the `account_activated?` one.

I would also rename the `account_activated` flag as is clearly ambiguous and prone to confusions (as happened to us).

frankieroberto: I think this is better, although "registration" is also arguably ambiguous (does registration include accepting the declaration or not?).

Maybe to be super explicit we could call it: `has_login_credentials?` or even `has_created_password_and_two_factor_authentication`?

scruti: I like the `has_login_credentials?` one.

slorek: I like `has_completed_registration` since we have decided to separate the declaration from registration.

frankieroberto: ok let's go with `has_completed_registration?`

psd-web/spec/requests/user_completes_registration_spec.rb
```diff
@@ -35,30 +35,30 @@
       end
     end
 
-    context "when the user already created an account with the invitation token" do
-      let(:user) { create(:user, :invited, :activated) }
+    context "when the user has already completed their registration" do
```
scruti: :clap:

psd-web/spec/requests/user_completes_registration_spec.rb
```diff
@@ -35,30 +35,30 @@
...
+      let(:user) { create(:user, invitation_token: "abc123", invited_at: Time.zone.now, account_activated: false) }
 
-      it "shows a message alerting about the account being already setup" do
+      it "redirects to the homepage" do
```
scruti: Nice catch! :clap:

psd-web/spec/models/user_spec.rb
```diff
@@ -328,4 +328,22 @@
       expect(user.invitation_expired?).to be true
     end
   end
+
+  describe "#has_completed_registration?" do
+    context "when the password, name and mobile number are missing" do
```
scruti: Suggestion: May be super verbose, but as unit tests are cheap... should we cover all the possible cases adding tests for when
- "password is missing" 
- "name is missing"
- "mobile number is missing"
and the other fields are filled the `has_completed_registration?` will return `false`?

psd-web/spec/factories/users.rb
```diff
@@ -38,9 +39,14 @@
     end
 
     trait :invited do
+      skip_password_validation { true }
       invitation_token { SecureRandom.hex(15) }
       invited_at { Time.zone.now }
       account_activated { false }
+      password { nil }
```
scruti: Question:question: 
Don't invited users have an empty password `""` instead of `nil`? Or is just for `password_encrypted` field?

psd-web/spec/models/user_spec.rb
frankieroberto: @scruti could do? Seemed a bit unnecessary to me, as that's not a scenario we'd ever currently expect. With 3 fields there's at least 8 possible combinations of values (counting only truthy/falsy ones)...

psd-web/spec/factories/users.rb
frankieroberto: The `encrypted_password` field can't be nil, but the `password=` setter seems to accept a nil value... 🤷‍♂

psd-web/spec/models/user_spec.rb
scruti: well, being literally one liner with  a couple `&&` is not like is going to have unexpected behaviours. :+1:

PR #321, opened 2020-02-28
Add rubocop-rspec linting

slorek:
## Description
This adds the [rubocop-rspec gem](https://github.com/rubocop-hq/rubocop-rspec) to the test suite to enforce spec styles. This is in order to avoid lots of code style comments in peer review.

I decided to increase the RSpec/NestedGroups Max setting to 5 as the default of 3 resulted in > 50 failures, and would require significant refactoring which I'm not sure would be a positive change. I did reduce one file down from 6 to 5. We can revisit this later if there is disagreement.

I have also added a special configuration for feature specs to reflect the way we are using them with long scenarios and multiple expectations.

There's a single instance of `allow_any_instance_of` which I've ignored for now as I don't think there's a straightforward way to avoid this, as it was [discussed in the original PR](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/119/files#diff-84b21e7ae9f54a4bc2a15aba433a14e4R4).

I would encourage you to review each commit individually as I have tried to address each cop independently.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


frankieroberto:
@slorek getting there! 🙌

psd-web/spec/features/filter_investigations_spec.rb
```diff
@@ -22,22 +22,22 @@
...
+  it "selecting filters only shows other active users in the assigned to and created by filters" do
     expect(page).to have_css("#assigned_to_someone_else_id option[value=\"#{another_active_user.id}\"]")
     expect(page).not_to have_css("#assigned_to_someone_else_id option[value=\"#{another_inactive_user.id}\"]")
 
     expect(page).to have_css("#created_by_someone_else_id option[value=\"#{another_active_user.id}\"]")
     expect(page).not_to have_css("#created_by_someone_else_id option[value=\"#{another_inactive_user.id}\"]")
   end
 
-  scenario "no filters applied shows all cases" do
+  it "no filters applied shows all cases" do
```
frankieroberto: I personally quite like the `feature` and `scenario` aliases for feature specs. But I can why other might not. If we do switch to `it` blocks, it’d be nice to re-word them so that they still read as a sentence - eg  `it “shows all cases when no filters are applied”` here.

We could do that as a future PR though.

psd-web/spec/features/.rubocop.yml
```diff
@@ -0,0 +1,7 @@
+inherit_from: ../../.rubocop.yml
+
+RSpec/MultipleExpectations:
+   Max: 20
+
+RSpec/ExampleLength:
```
frankieroberto: Should we scope this exception to feature specs only?

psd-web/Gemfile
```diff
@@ -65,6 +65,7 @@ group :development, :test do
   gem "rubocop"
   gem "rubocop-govuk", "~> 2.0.0"
   gem "rubocop-performance"
+  gem "rubocop-rspec", require: false
```
frankieroberto: I think it’d be good to start specifying versions in the Gemfile using the pessimistic operator (`~>`) - apparently called the “twiddle-wakka”:

```suggestion
  gem "rubocop-rspec", "~> 1.37", require: false
```

psd-web/spec/features/filter_investigations_spec.rb
slorek: I do as well but this is the default cop settings. I also sometimes struggled with switching between `scenario/it` as it requires a quite different approach to test structure and semantics. I agree we should reword the descriptions. I hadn’t actually noticed autocorrect had done this until after opening the PR. I’ll try to do this on Monday.

psd-web/spec/features/.rubocop.yml
slorek: The location of the file limits the scope of this config.

frankieroberto: Doh, I didn’t spot that!

psd-web/spec/support/not_talk_to_db_matcher.rb
```diff
@@ -1,9 +1,13 @@
 RSpec::Matchers.define :not_talk_to_db do |_expected|
   match do |block_to_test|
-    %w(exec_delete exec_insert exec_query exec_update).each do |meth|
-      expect(ActiveRecord::Base.connection).not_to receive(meth)
-    end
+    stub_methods = %w(exec_delete exec_insert exec_query exec_update)
+
+    connection_double = instance_double("ActiveRecord::ConnectionAdapters::AbstractAdapter", Hash[stub_methods.zip])
```
scruti: :clap:

PR #328, opened 2020-03-02
Set PR_NUMBER from github event context

frankieroberto:
This should be more reliable than parsing the `GITHUB_REF` environment variable, as that variable doesn’t contain the PR number when re-running a failed workflow.

See https://github.community/t5/GitHub-Actions/GITHUB-REF-is-inconsistent/m-p/48129

### DIFF
--- a/.github/workflows/review-apps.yml
+++ b/.github/workflows/review-apps.yml
@@ -10,18 +10,17 @@ jobs:
     steps:
     - uses: actions/checkout@v1
 
+    - name: Get GitHub Pull Request number
+      run: |
+        PR_NUMBER=${{ github.event.number }}
+        echo "::set-env name=PR_NUMBER::$PR_NUMBER"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/review-apps.yml
```diff
@@ -10,18 +10,17 @@ jobs:
     steps:
     - uses: actions/checkout@v1
 
+    - name: Get GitHub Pull Request number
+      run: |
+        PR_NUMBER=${{ github.event.number }}
```
scruti: This is waaaaay better. Do you know if this env value is available when a PR is merged? (as we also need/parse the PR number when merging/closing PRs).

frankieroberto: Yeah, hopefully the event context for closing and merging a PR _also_ includes the PR. Not sure how we can verify this other than by testing it though.

PR #331, opened 2020-03-02
Copy mobile phone and other data from keycloak

slawosz:
https://trello.com/c/WNCc6QD0/392-import-mobile-numbers-and-other-stuff-from-keycloak

Adds several keycloak fields to user table. Most important is mobile number.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,6 +6,8 @@ services:
       context: ./docker/db
     volumes:
       - postgres-volume:/var/lib/postgresql/data
+    env_file:
+      - ./docker/env.keycloak
     ports:
       - 5432:5432
 
@@ -17,6 +19,18 @@ services:
       - PORT=3006
     ports:
       - "3006:3006"
+
+  keycloak:
+    image: beisopss/keycloak:master
+    env_file:
+      - ./docker/env.keycloak
+    environment:
+      - NOTIFY_API_KEY=
+    depends_on:
+      - db
+    ports:
+      - "8080:8080"
+      - "9990:9990"
 
   elasticsearch:
     image: docker.elastic.co/elasticsearch/elasticsearch-oss:6.2.3
@@ -49,6 +63,7 @@ services:
       - elasticsearch
       - redis
       - psd-worker
+      - keycloak
     env_file:
       - ./docker/env.psd
 
--- a/docker/env.keycloak
+++ b/docker/env.keycloak
@@ -1,0 +1,7 @@
+# Note: These are DEVELOPMENT values only - not used in production
+
+# Keycloak database credentials
+KEYCLOAK_DATABASE=keycloak
+KEYCLOAK_DATABASE_HOSTNAME=db
+KEYCLOAK_DATABASE_USERNAME=keycloak
+KEYCLOAK_DATABASE_PASSWORD=password
--- a/psd-web/app/models/keycloak_connector.rb
+++ b/psd-web/app/models/keycloak_connector.rb
@@ -1,3 +1,4 @@
+# Trigger
 class KeycloakConnector < ApplicationRecord
   CREDENTIALS_QUERY = <<~SQL.freeze
     SELECT c.salt, c.value, c.hash_iterations, u.email, c.type FROM credential c INNER JOIN user_entity u ON c.user_id = u.id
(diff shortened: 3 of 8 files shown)


docker-compose.yml
```diff
@@ -18,6 +20,18 @@ services:
     ports:
       - "3006:3006"
 
+  keycloak:
```
slawosz: We still need keycloak locally

psd-web/config/initializers/sidekiq.rb
```diff
@@ -41,7 +41,7 @@ def schedule_keycloak_sync_job
 def schedule_keycloak_password_job
   keycloak_sync_job = Sidekiq::Cron::Job.new(
     name: "#{ENV['SIDEKIQ_QUEUE'] || 'psd'}: Sync user passwords with Keycloak",
-    cron: "*/5 * * * *",
+    cron: "*/1 * * * *",
```
slawosz: For testing on review app, will be changed later.

psd-web/db/schema.rb
```diff
@@ -174,17 +174,6 @@
...
-    t.datetime "created_at", null: false
-    t.string "credential_type"
-    t.string "email"
-    t.string "encrypted_password"
-    t.integer "hash_iterations"
-    t.binary "salt"
-    t.datetime "updated_at", null: false
-    t.index ["email"], name: "index_keycloak_credentials_on_email"
-  end
-
```
frankieroberto: Why have these been removed from the schema?

psd-web/app/models/keycloak_connector.rb
```diff
@@ -1,20 +1,50 @@
...
+
+    ua.value as mobile_number,
+
+    ue.id, ua.user_id, c.user_id, ua.name
+  SQL
+
   CREDENTIALS_QUERY = <<~SQL.freeze
-    SELECT c.salt, c.value, c.hash_iterations, u.email, c.type FROM credential c INNER JOIN user_entity u ON c.user_id = u.id
+    SELECT
+    #{FIELDS}
```
frankieroberto: I think I'd be tempted to include the list of fields here, one per line, rather than interpolate the string. Otherwise, maybe the var should be `COLUMNS` instead of `FIELDS`?

docker/env.keycloak
```diff
@@ -0,0 +1,7 @@
+# Note: These are DEVELOPMENT values only - not used in production
+
+# Keycloak database credentials
+KEYCLOAK_DATABASE=keycloak
+KEYCLOAK_DATABASE_HOSTNAME=db
+KEYCLOAK_DATABASE_USERNAME=keycloak
+KEYCLOAK_DATABASE_PASSWORD=password
```
frankieroberto: Do we need these in the branch?

slawosz: Yes. We need it locally.

psd-web/db/schema.rb
slawosz: Probably because of earlier merging issues.

psd-web/app/models/keycloak_connector.rb
slawosz: Both ways work I guess. `COLUMNS` is better name.

psd-web/db/schema.rb
frankieroberto: Yep. 😩  Not your fault. I've fixed this separately in #362 but it can be fixed in your branch too if that gets merged first.

docker/env.keycloak
slawosz: Yes, we still need keycloak locally.

psd-web/db/schema.rb
slawosz: Hard to say, merging issue probably. We don't create this table anyway.

psd-web/app/models/keycloak_connector.rb
```diff
@@ -1,21 +1,53 @@
...
       next unless user
 
-      user.password_salt = salt
-      user.encrypted_password = encrypted_password
-      user.hash_iterations = iterations
-      user.credential_type = credential_type
-      user.save!
+      user.keycloak_first_name = row[1]
+      user.keycloak_last_name = row[2]
+      user.keycloak_username = row[3]
```
slorek: Why do we need to store these three attributes? The email and name are already synced.

slorek:
Suggest renaming the `SyncKeycloakPasswordsJob` class as it is now syncing extra data as well.

PR #333, opened 2020-03-02
Require 2FA on Sign In

scruti:
Introduce two factor authentication on Sign in.

 ### TO DO:
- [x] Fix couple broken tests
- [x] Check if the branch accomplish acceptance criteria and, in that case, take screenshots, improve PR description and ask for a formal review.
- [x] Pass QA check.
- [x] address previous PR comments.

## How to Review
- Review App link: https://psd-pr-333.london.cloudapps.digital/
- Add your phone in Gov UK Notify team.
- Set up your phone for the user@example.com in Rails console for the `psd-pr-333` app
- Try to login as user@example.com

## Flow example
Screenshot from 2020-03-03 16-25-00

Screenshot from 2020-03-03 16-25-20

### Fixed to be 5 digits. Don't mind the 6 digits here
Screenshot from 2020-03-03 17-06-55

Screenshot from 2020-03-03 17-07-08

## Errors
Screenshot from 2020-03-13 09-27-27

Screenshot from 2020-03-13 09-28-53

Screenshot from 2020-03-13 09-29-43

### When introducing the wrong password or being locked for an hour due too many attempts:
Screenshot from 2020-03-13 10-47-39

### When the introduced code expired
Screenshot from 2020-03-13 16-54-31

### When introducing letters as part of the code
Screenshot from 2020-03-13 17-21-39


psd-web/app/models/user.rb
```diff
@@ -30,6 +33,9 @@ class User < ApplicationRecord
 
   attribute :skip_password_validation, :boolean, default: false
 
+  # TODO: remove when registration is merged
+  attribute :mobile_number
```
scruti: To do: Need to remove this. Registration has been merged

psd-web/app/controllers/users/two_factor_authentication_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+      redirect_to after_two_factor_success_path_for(resource)
+    end
+
+    def after_two_factor_fail_for(resource)
+      resource.second_factor_attempts_count += 1
+      resource.save
+
+      if resource.max_login_attempts?
+        sign_out(resource)
+        resource.errors.add(:direct_otp, find_message(:max_login_attempts_reached))
```
frankieroberto: I can’t see where `find_message` is defined? Should we add the error directly from `I18n.t` instead?

scruti:
Thanks for the comments @frankieroberto , going to re-apply the WIP tag while I address your comments.

psd-web/spec/models/user_spec.rb
```diff
@@ -417,4 +417,24 @@
       end
     end
   end
+
+  describe "devise gems echosystem" do
```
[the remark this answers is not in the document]
frankieroberto: Oh, I just realised that this is probably a typo for "ecosystem"? I previously assumed that "echosystem" was some term I'd not heard of... 😂 

I'd suggest either something like `describe "devise-overriden methods"` or otherwise just dropping this and un-nesting the `#send_two_factor_authentication_code`, possibly with a comment to note that the method overrides one from the super class?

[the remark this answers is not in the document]
StupidCodeFactory: Nesting is useful, it provide `context` (i'll actually change this to `context`) and avoids re-write ten times the thing `when the code is ok, then is does this` `when the code is ok, and this is turned on, then is does this`.

.github/workflows/rspec.yml
```diff
@@ -16,6 +16,9 @@ env:
   ANTIVIRUS_USERNAME: av
   ANTIVIRUS_PASSWORD: password
   PSD_HOST: example.com
+  OTP_SECRET_ENCRYPTION_KEY: 904c9fd9a034b993b891606758476d736e8d58a058528a2dbef9f479ba32c0b7f2e192a90b18b71137e4e966c6b1aa530139fe8629214ac2d7f33d4139febcec
+  TWO_FACTOR_AUTHENTICATION_ENABLED: false
```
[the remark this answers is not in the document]
scruti: Will revisit this comment tomorrow :P

psd-web/app/models/user.rb
```diff
@@ -181,9 +185,37 @@ def invitation_expired?
...
+  def send_two_factor_authentication_code(code)
+    SendTwoFactorAuthenticationJob.perform_later(self, code)
+  end
+
+  def need_two_factor_authentication?(_request)
+    Rails.configuration.two_factor_authentication_enabled
+  end
+
+private
+
```
scruti: Given the amount of methods (+ a constant) related to 2FA introduced in User model... I'm tempted to extract this to a `TwoFactorAuthentication` concern. Thoughts?

psd-web/app/views/devise/two_factor_authentication/show.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    <p class="govuk-body">We have sent you a security code by text message.</p>
+    <%= form_tag([resource_name, :two_factor_authentication], :method => :put) do %>
+      <%= render "components/govuk_input", id: "direct_otp", name: "user[direct_otp]", type: "text", classes: "govuk-input--width-5", label: { text: "Enter security code" }, attributes: { pattern: "[0-9]*", inputmode: "numeric" } %>
+
+      <%= govukButton(text: "Continue") %>
+    <% end %>
+
+    <div class="govuk-body">
+      <%= link_to "Not received a text message?", resend_code_user_two_factor_authentication_path, class: "govuk-link" %>
+    </div>
```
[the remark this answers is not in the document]
scruti: Will follow up on this tomorrow!

psd-web/spec/factories/documents.rb
```diff
@@ -1,17 +1,38 @@
...
+                         description: evaluator.document_description,
+                         updated: Time.now.iso8601
+                        })
+
+      model.documents.attach(file)
+    end
+  end
+
+  trait :with_antivirus_checked_document do
+    with_document_info
```
scruti: If you land here and ask yourself why are we touching Document/Attachment related code for a 2FA ticket... you're right, it does not make sense.

... but we had a flaky test on merge against base branch and had to solve it :P

psd-web/spec/jobs/send_two_factor_authentication_job_spec.rb
```diff
@@ -0,0 +1,14 @@
+require "rails_helper"
+
+RSpec.describe SendTwoFactorAuthenticationJob do
+  let(:user) { create(:user) }
+  let(:code) { 123 }
+
+  it "send the otp code" do
+    allow(SendSMS).to receive(:otp_code)
+    described_class.perform_now(user, code)
```
StupidCodeFactory: NIT: add some space around this line to make it clearer that's where the magic happens

psd-web/spec/models/user_spec.rb
```diff
@@ -417,4 +419,81 @@
...
+          expect(user.direct_otp).to eq(code_param)
+        end
+      end
+      # rubocop:enable RSpec/MultipleExpectations, RSpec/ExampleLength
+    end
+
+    describe "#two_factor_lock_expired?" do
+      it "defaults to true when there was no two factor lock for the user" do
+        user = build_stubbed(:user, second_factor_attempts_locked_at: nil)
+        expect(user.two_factor_lock_expired?).to eq true
```
StupidCodeFactory: consider using rspec syntactic sugar, it reads much nicer:
```suggestion
        expect(user).to be_two_factor_lock_expired
```

psd-web/spec/models/user_spec.rb
```diff
@@ -417,4 +419,81 @@
...
+      end
+
+      it "returns true when the lock time expired for the user" do
+        expired_lock_time = Time.current - (User::TWO_FACTOR_LOCK_TIME + 10)
```
StupidCodeFactory: 10 hours? 10 seconds? 
```suggestion
        expired_lock_time = (User::TWO_FACTOR_LOCK_TIME + 10.seconds).ago
```

psd-web/spec/requests/two_factor_authentication_spec.rb
```diff
@@ -0,0 +1,234 @@
...
+        it "removes the 2fa lock timestamp from the user" do
+          expect {
+            submit_2fa
+            user.reload
+          }.to change(user, :second_factor_attempts_locked_at).to(nil)
+        end
+      end
+    end
+  end
+end
```
StupidCodeFactory: Opinion: This is a really long and difficult to follow spec file. This is a clear sign the controller is doing way too much stuff and could have benefited with an interstitial _service_ class like `VerifyOPT.call(user: resource, params: params)`
This would have halved the controller and this file, making it easier to reason about, test and maintain. 

All the tests with `follow_redirect!` could be extracted to feature tests, that would help clean up this file a bit.

psd-web/spec/services/send_sms_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    let(:code) { 123 }
+    let(:expected_payload) do
+      { phone_number: phone_number, template_id: described_class::TEMPLATES[:otp_code], personalisation: { code: code } }
+    end
+
+    it "sends the otp code" do
+      stub_request(:post, "https://api.notifications.service.gov.uk/v2/notifications/sms")
+        .with(body: expected_payload.to_json).and_return(body: {}.to_json)
+
+      described_class.otp_code(mobile_number: phone_number, code: code)
```
StupidCodeFactory: this test isn't testing anything anymore... just ✂️

psd-web/app/controllers/declaration_controller.rb
```diff
@@ -1,6 +1,7 @@
 class DeclarationController < ApplicationController
   skip_before_action :has_accepted_declaration
   skip_before_action :has_viewed_introduction
+
```
StupidCodeFactory: ✂️

psd-web/spec/models/user_spec.rb
scruti: Much clearer :clap:

.github/workflows/rspec.yml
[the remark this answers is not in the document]
slorek: I agree it would be preferable not to have to set these env vars to run the tests. They could also cause confusion. I get the point about wanting the application to fail on start up in production mode if not set, but this behaviour could be limited to production mode. They should at the very least be added to `.env.test.example` so that developers know what settings they need to manage - I would also recommend documenting them in the `README.md`.

psd-web/.env.test.example
```diff
@@ -15,3 +15,5 @@ ELASTICSEARCH_URL=http://localhost:9200
 ANTIVIRUS_URL=http://localhost:3006/safe
 ANTIVIRUS_USERNAME=av
 ANTIVIRUS_PASSWORD=password
+
+TWO_FACTOR_AUTHENTICATION_ENABLED=false
```
slorek: Seems like we need to add `OTP_SECRET_ENCRYPTION_KEY` here as well.

.github/workflows/system_tests.yml
```diff
@@ -15,6 +15,9 @@ env:
   RAILS_ENV: test
   PGHOST: localhost
   PGUSER: postgres
+  NOTIFY_API_KEY: eef30646-d6cc-4708-bb7a-46d63caf752a.653d8bdf-03d1-4734-81a9-9cc687a5e1b3
+  OTP_SECRET_ENCRYPTION_KEY: 904c9fd9a034b993b891606758476d736e8d58a058528a2dbef9f479ba32c0b7f2e192a90b18b71137e4e966c6b1aa530139fe8629214ac2d7f33d4139febcec
+  TWO_FACTOR_AUTHENTICATION_ENABLED: false
```
slawosz: Please change: move all ENVS to .env.test.example and copy it over in github tests.

psd-web/.env.test.example
slawosz: Please change: add missing variables.

psd-web/app/models/user.rb
slawosz: Good idea.

psd-web/app/models/user.rb
```diff
@@ -181,9 +185,37 @@ def invitation_expired?
...
+    (second_factor_attempts_locked_at + TWO_FACTOR_LOCK_TIME) < Time.current
+  end
+
+  def lock_two_factor!
+    self.update_column(:second_factor_attempts_locked_at, Time.current)
+  end
+
+  def unlock_two_factor!
+    self.update_columns(second_factor_attempts_locked_at: nil)
```
slorek:
```suggestion
    self.update_column(second_factor_attempts_locked_at: nil)
```
Minor edit for consistency with previous method.

psd-web/spec/models/user_spec.rb
scruti: Ended taking a `travel_to` approach to make more obvious that the test is happening before/after the locked time expires. (Or at least I found it more readable).

psd-web/app/models/user.rb
slorek: I prefer to keep it all in one place, it makes it a lot easier to debug.

psd-web/app/models/user.rb
```diff
@@ -181,9 +185,37 @@ def invitation_expired?
...
+  end
 
   # BEGIN: place devise overriden method calls bellow
+  def send_two_factor_authentication_code(code)
+    SendTwoFactorAuthenticationJob.perform_later(self, code)
```
slorek: We would want to check that Devise limits the number of codes that can be sent to a user in a given space of time.

psd-web/config/initializers/devise.rb
```diff
@@ -317,6 +317,19 @@
   # When set to false, does not sign a user in automatically after their password is
   # changed. Defaults to true, so a user is signed in automatically after changing a password.
   # config.sign_in_after_change_password = true
+
+  # Devise two_factor_authentication gem
+  if ENV["OTP_SECRET_ENCRYPTION_KEY"]
```
slorek: Suggest checking `TWO_FACTOR_AUTHENTICATION_ENABLED` instead.

psd-web/spec/models/user_spec.rb
```diff
@@ -417,4 +419,81 @@
...
+          .to enqueue_job(SendTwoFactorAuthenticationJob)
+                .at(:no_wait)
+                .on_queue("psd")
+                .with do |user_param, code_param|
+          expect(user_param).to be user
+          expect(user.direct_otp).to eq(code_param)
+        end
+      end
+      # rubocop:enable RSpec/MultipleExpectations, RSpec/ExampleLength
+    end
```
slorek:
```suggestion
      it "enqueues a SendTwoFactorAuthenticationJob", :with_test_queue_adapter do
        expect { user.send_new_otp }
          .to enqueue_job(SendTwoFactorAuthenticationJob)
                .at(:no_wait)
                .on_queue("psd")
                .with(user, code_param)
      end
    end
```

A couple of issues here:
- Reset the queue adapter after the example. otherwise all subsequent tests will be using this different `test` adapter. Ideally create a shared example (`with_test_queue_adapter`) which has before/after hooks.
- You should be able to use the `with` style above to avoid having to disable the `RSpec/MultipleExpectations` cop.

psd-web/app/services/send_sms.rb
```diff
@@ -0,0 +1,29 @@
+class SendSMS
+  include Singleton
+
+  TEMPLATES = {
+    otp_code: "091c8861-8532-4907-abc0-f89632c34f09"
```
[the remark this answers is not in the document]
slorek: Yes it's just a template ID. We are doing so elsewhere.

psd-web/spec/requests/two_factor_authentication_spec.rb
scruti: I agree on the code smell. As I was increasing the logic to cover some edge cases and achieve the acceptance criteria the controller and tests were growing and growing on complexity and brittleness. 

Originally I was thinking about extracting a `TwoFactorForm` to deal with the code format validations and its tests but thought could be an overkill given its a form with a single field. 

Extracting a service seems a good idea, hope I manage to get something out without Devise dependencies.

slorek:
I am not able to test the review app currently as it is throwing a 500 error.

psd-web/.env.test.example
StupidCodeFactory: I'm actually not sure about this, this PR does not encrypt the OTP code. Here is the documentation fro encrypting the code https://github.com/Houdini/two_factor_authentication/tree/1d6c97809c9c5547a2809b1a96440c7e05174af5#adding-the-totp-encryption-option-to-an-existing-app

This key is only used then, otherwise it is useless and we can probably remove it if we plan not to encrypt our OTPs which we probably should.

psd-web/app/models/user.rb
StupidCodeFactory: @slorek `update_column` takes two arguments it would be  `self.update_column(second_factor_attempts_locked_at,  nil)`

psd-web/spec/services/send_sms_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+      { phone_number: phone_number, template_id: described_class::TEMPLATES[:otp_code], personalisation: { code: code } }
+    end
+
+    it "sends the otp code" do
+      notification = stub_request(:post, "https://api.notifications.service.gov.uk/v2/notifications/sms")
+                      .with(body: expected_payload.to_json).and_return(body: {}.to_json)
+
+      described_class.otp_code(mobile_number: phone_number, code: code)
+
+      expect(notification).to have_been_requested
```
StupidCodeFactory: `have_been_requested.with(expected_payload)` otherwise remove the `let(:expected_payload) {}`

StupidCodeFactory: but this test has 0 value if you don't test what the payload looks like and I'd be in favour of removing this test if we're not testing anything valuable.

psd-web/spec/requests/two_factor_authentication_spec.rb
StupidCodeFactory: I'm on board with the TwoFactorForm idea, even if it's one field. I like small and easy to reason about classes that follow the SRP.

psd-web/spec/services/send_sms_spec.rb
scruti: The  `notification` stub definition only stubs the call with the expected body,
so calling `expect(notification).to have_been_requested` asserts over the `expected_payload` being part of the call.

psd-web/.env.test.example
scruti: Removed OTP_SECRET_ENCRYPTION_KEY as is used for encrypted TOTP codes and we are doing Direct OTP.

scruti: Removed the other env var (`OTP_SECRET_ENCRYPTION_KEY`)

scruti:
[(3) Requirement to use 2FA when signing in and password reset](https://trello.com/c/82J6GsM3/344-3-requirement-to-use-2fa-when-signing-in-and-password-reset)

scruti:
[(3) Requirement to use 2FA when signing in and password reset](https://trello.com/c/82J6GsM3/344-3-requirement-to-use-2fa-when-signing-in-and-password-reset)

slorek:
2FA code is not currently being sent. Error messages seem to work fine. I think on the code entry page the header nav links should be hidden (see screenshot highlighted red).

Screenshot 2020-03-11 at 11 50 33

slorek:
Issue: when attempting to sign in and getting to the 2FA code entry page, it is impossible to abandon the sign in attempt and go back to the home page or sign in page. If I want to use a different login or reset my password I am not able to.

psd-web/spec/services/send_sms_spec.rb
StupidCodeFactory: ah my bad I thought i had seen`.with(body: {}.to_json)` ...

edwardhorsford:
As discussed in person, ideally the header would have no nav links and have a 'Sign out' link.

scruti:
Navigation links removed and "Sign Out" link added to the header.
Screenshot from 2020-03-11 15-40-07

psd-web/spec/models/user_spec.rb
scruti: Seems using `with(user, user.direct_otp)` does not refresh the `user.direct_otp` before defining the expectation.
So gets the expectation with, eg: "12345", then the `user.send_new_otp` refreshes the code for user, eg: "54321", but the expectation was already defined. So when checks the assertion compares "12345" (expected) with "54321" (received) and fails.

psd-web/spec/models/user_spec.rb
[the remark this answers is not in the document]
scruti: I am happy to keep it as `describe` as is `describ`ing what are we testing, instead of setting up an specific `context` for following tests.
Not all the methods under this block are overriding Devise methods, so preferred a more generic "Devise gem related methods" description.

psd-web/app/models/user.rb
scruti: Tried to implement this by guard clause this method with a check against the last `direct_otp_sent_at` to don't queue the job if the timestamp was less than 10 seconds ago.
What I discover through the test, is that by the time we reach this point, `direct_otp_sent_at` [has been override by a new timestamp](https://github.com/Houdini/two_factor_authentication/blob/82e5aff7b344ac2c5f4543a3887db3e6bfcd0008/lib/two_factor_authentication/models/two_factor_authenticatable.rb#L101-L108). So we always have something close to the current time when reaching here and that defeats my attempt to limit triggering jobs at this point.

We could add a rate limit middleware like Rack-Attack for the endpoint or a throttler for Sidekiq, but guess that adding those extra layers should be discussed on project-level.

psd-web/app/controllers/users/two_factor_authentication_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    def secondary_nav_items
+      [text: "Sign out", href: destroy_user_session_path]
+    end
+
+  private
+
+    def two_factor_authentication_form
+      @two_factor_authentication_form ||= TwoFactorAuthenticationForm.new(otp_code: otp_code_param)
+    end
```
frankieroberto: Suggestion: I wonder if the controller would be slightly easier to follow if this private method was just a local variable assignment at the start of the `update` method? ie:

```ruby
def update
  two_factor_authentication_form = TwoFactorAuthenticationForm.new(otp_code: otp_code_param)

  if two_factor_authentication_form.invalid?
    resource.errors.merge!(two_factor_authentication_form.errors)
    return render :show
  end
 
  ...
```

psd-web/app/controllers/users/two_factor_authentication_controller.rb
```diff
@@ -0,0 +1,73 @@
+module Users
+  class TwoFactorAuthenticationController < Devise::TwoFactorAuthenticationController
+    skip_before_action :has_accepted_declaration
+    skip_before_action :has_viewed_introduction
+
+    def update
```
frankieroberto: Would it be worth adding a comment to explain how the `resource` variable is set, and what it contains? Presumably this comes from the superclass controller somehow?

psd-web/app/models/user.rb
```diff
@@ -181,9 +185,54 @@ def invitation_expired?
...
+
+  def two_factor_locked?
+    second_factor_attempts_locked_at && !two_factor_lock_expired?
+  end
+
+  def fail_two_factor_authentication!
+    return if max_login_attempts?
+
+    self.second_factor_attempts_count += 1
+    self.save
```
frankieroberto: Could use the Rails [`increment!` method](https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-increment) here:

```suggestion
    self.increment!(:second_factor_attempts_count, 1)
```

The main difference is that validations and any callbacks will be skipped.

psd-web/app/models/user.rb
```diff
@@ -181,9 +185,54 @@ def invitation_expired?
...
+  def lock_two_factor!
+    self.update_column(:second_factor_attempts_locked_at, Time.current)
+  end
+
+  def unlock_two_factor!
+    self.update_column(:second_factor_attempts_locked_at, nil)
+  end
+
+  def two_factor_lock_expired?
+    return true if !second_factor_attempts_locked_at
```
frankieroberto: Suggestion for readability:

```suggestion
    return true if second_factor_attempts_locked_at.nil?
```

psd-web/app/services/send_sms.rb
```diff
@@ -0,0 +1,29 @@
...
+  }.freeze
+
+  attr_reader :pool
+
+  def initialize
+    self.pool = ConnectionPool.new(size: ENV.fetch("RAILS_MAX_THREADS", 5), timeout: 1) do
+      Notifications::Client.new(Rails.configuration.notify_api_key)
+    end
+  end
```
frankieroberto: I'm not convinced that we need connection pooling for this client, as it just does a new http request for each API call and doesn’t seem to maintain any long-lived connections. The client also seems to initialize pretty quickly.

If we do keep the connection pool, I'm not sure that it makes sense to default the connection pool size to the `RAILS_MAX_THREADS` size, as this is always called asynchronously from the queue, rather from within one of the Puma threads.  Sidekiq is currently hardcoded to with a concurrency value of 10 (in `config/sidekiq.yml`).

psd-web/config/locales/en.yml
```diff
@@ -157,6 +157,12 @@ en:
           temporary: "Temporarily"
           unknown: "Unknown"
   enter_email_address: "Enter your email address"
+  otp_code:
+    blank: "Enter the security code"
+    too_short: "You haven't entered enough numbers"
+    too_long: "You've entered too many numbers"
+    incorrect: "Incorrect security code"
+    expired: "You took too long to login. Login process starting from beginning."
```
frankieroberto: Think this line needs a content review... 😉

psd-web/db/migrate/20200309104244_two_factor_authentication_add_to_users.rb
```diff
@@ -0,0 +1,18 @@
...
+    safety_assured do
+      change_table :users, bulk: true do
+        add_column :users, :second_factor_attempts_count, :integer, default: 0
+        add_column :users, :second_factor_attempts_locked_at, :datetime
+        add_column :users, :encrypted_otp_secret_key, :string
+        add_column :users, :encrypted_otp_secret_key_iv, :string
+        add_column :users, :encrypted_otp_secret_key_salt, :string
+        add_column :users, :direct_otp, :string
+        add_column :users, :direct_otp_sent_at, :datetime
+        add_column :users, :totp_timestamp, :timestamp
```
frankieroberto: Do we need this column? Seems like we're not using TOTP?

psd-web/config/locales/en.yml
scruti: Yup! Got the wrong version of the acceptance text. Thanks for raising it!

psd-web/app/models/user.rb
scruti: Nice!

psd-web/db/migrate/20200309104244_two_factor_authentication_add_to_users.rb
scruti: Seems a default created by the gem [automatic initial setup](https://github.com/Houdini/two_factor_authentication#automatic-initial-setup). If we plan to add TOTP I guess is ok to have it here instead of having to remember to manually add this db field later.

frankieroberto: I’d suggest that it’s _probably_ be better to add the column later, when it’s actually needed. But it's not a big deal.

psd-web/app/controllers/users/two_factor_authentication_controller.rb
scruti: sure

edwardhorsford:
Suggestion:
* If code entered is not numeric, reply with error `The code must be 5 numbers`

.github/workflows/rspec.yml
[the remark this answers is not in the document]
scruti: API keys removed :white_check_mark:

.github/workflows/system_tests.yml
scruti: Done! (for the Notify API Key)

psd-web/app/controllers/users/two_factor_authentication_controller.rb
scruti: Done!

psd-web/spec/requests/two_factor_authentication_spec.rb
scruti: Implemented the form (that ended being very helpful) + moved bunch of logic to the user model. I think tests became way easier to follow and controller logic does the minimum to route (or got way better on this aspect at least).

psd-web/app/forms/two_factor_authentication_form.rb
```diff
@@ -0,0 +1,22 @@
+class TwoFactorAuthenticationForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  # From rails validations guide: https://guides.rubyonrails.org/active_record_validations.html#numericality
+  NUMERICALITY_REGEX = /\A[+-]?\d+\z/.freeze
```
frankieroberto: I think we only need to check for digits, not `+` or `-` prefixes, as we’re not _really_ checking whether it’s a logical number or not.

```suggestion
  SECURITY_CODE_REGEX = /\A\d+\z/.freeze
```

psd-web/app/forms/two_factor_authentication_form.rb
```diff
@@ -0,0 +1,22 @@
...
+            numericality: { only_integer: true, message: I18n.t(".otp_code.numericality") },
+            allow_blank: true
+  validates :otp_code,
+            length: { maximum: Devise.direct_otp_length, too_long: I18n.t(".otp_code.too_long") },
+            allow_blank: true,
+            if: -> { NUMERICALITY_REGEX === otp_code }
+  validates :otp_code,
+            length: { minimum: Devise.direct_otp_length, too_short: I18n.t(".otp_code.too_short") },
+            allow_blank: true,
+            if: -> { NUMERICALITY_REGEX === otp_code }
```
frankieroberto: Could you combine these like this?

```suggestion
  validates :otp_code,
            length: { 
              maximum: Devise.direct_otp_length, 
              too_long: I18n.t(".otp_code.too_long"),
              minimum: Devise.direct_otp_length,
              too_short: I18n.t(".otp_code.too_short")
            },
            allow_blank: true,
            if: -> { NUMERICALITY_REGEX === otp_code }
```

Also, if you remove the `too_short: I18n.t(".otp_code.too_short")` lines, could you use the default scope within the `en.yml` file?

psd-web/app/controllers/users/two_factor_authentication_controller.rb
```diff
@@ -0,0 +1,81 @@
+module Users
+  class TwoFactorAuthenticationController < Devise::TwoFactorAuthenticationController
+    skip_before_action :has_accepted_declaration,
+                       :has_viewed_introduction
+
+    def show
+      # Supress unwanted Devise alerts.
+      # eg: when you try to navigate back to sign-in it alerts about you already being signed in.
+      # We only want to show submission errors.
+      flash.delete(:alert)
```
StupidCodeFactory: What we ended up doing is removing the values of the keys on the translation yml files. This has the nasty effect that it will suppress all the flash message which we may not  want to do.

psd-web/app/forms/two_factor_authentication_form.rb
scruti: Try to use default scope but didn't find the correct scope for it after trying to place the `otp_code` messages under different scopes :man_shrugging:

psd-web/app/controllers/users/two_factor_authentication_controller.rb
scruti: Guess we don't want to globally suppress the "You're already signed in" message from Devise through removing it from the translation file.
I can combine both approaches by doing a flash delete filtering by the particular message so only gets suppressed in this particular action.

edwardhorsford: I don't actually know what benefit a `You're already signed in` message would have - so it might be fine to disable globally.

PR #340, opened 2020-03-04
pgcrypto in our databases

slawosz:
Document pgcrypto and add it to review apps

### DIFF
--- a/psd-web/db/schema.rb
+++ b/psd-web/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_02_25_163932) do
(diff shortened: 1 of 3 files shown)


psd-web/db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_02_25_163932) do
+ActiveRecord::Schema.define(version: 2020_02_25_163952) do
```
frankieroberto: This shouldn't have changed as you’ve not added any migrations in this branch...? 🤔

slawosz: Its me breaking the rules. It was done to trigger db creation.

PR #347, opened 2020-03-06
Refactor: move conditionals to the existing job class

frankieroberto:
This is a possible refactor of the controller code, for discussion.

### DIFF
--- a/psd-web/app/jobs/send_user_invitation_job.rb
+++ b/psd-web/app/jobs/send_user_invitation_job.rb
@@ -3,11 +3,27 @@ require "notifications/client"
 class SendUserInvitationJob < ApplicationJob
   GOV_UK_NOTIFY_TEMPLATE_ID = "22b3799c-aa3d-43e8-899d-3f30307a488f".freeze
 
-  def perform(user_id, user_inviting_id)
+  # If the user_inviting_id is present, it is implied that the invitation
+  # is being sent (or resent) by a colleague, and so the invited_at time
+  # is reset.
+  #
+  # If absent, it is implied that the user themselves requested a resend of the
+  # invitation (eg by attempting to reset their password), and so the invited_at
+  # time should not be altered, and a different email should be sent if the invitation
+  # has already expired.
+  def perform(user_id, user_inviting_id = nil)
     user = User.find(user_id)
-    user_inviting = User.find(user_inviting_id)
 
-    NotifyMailer.invitation_email(user, user_inviting).deliver_now
-    user.update!(has_been_sent_welcome_email: true, invited_at: Time.current)
+    if user_inviting_id
+      user_inviting = User.find(user_inviting_id)
+
+      NotifyMailer.invitation_email(user, user_inviting).deliver_now
+      user.update!(has_been_sent_welcome_email: true, invited_at: Time.current)
+
+    elsif user.invitation_expired?
+      NotifyMailer.expired_invitation_email(user).deliver_now
+    else
+      NotifyMailer.invitation_email(user, nil).deliver_now
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


psd-web/app/jobs/send_user_invitation_job.rb
```diff
@@ -3,11 +3,27 @@
...
+
+      NotifyMailer.invitation_email(user, user_inviting).deliver_now
+      user.update!(has_been_sent_welcome_email: true, invited_at: Time.current)
+
+    elsif user.invitation_expired?
+      NotifyMailer.expired_invitation_email(user).deliver_now
+    else
+      NotifyMailer.invitation_email(user, nil).deliver_now
+    end
   end
```
scruti: As an alternative of having all this logic in the job (not sure myself how good/bad practice is this, even when I suggested it), we could extract this into an instance method at user model.

Then the "other user re-sending the invite would fall under the current `User.resend_invite` while the `User#resend_invite` would contain the logic for their own instance like:
```ruby
 def resend_invite
  if invitation_expired?
    ...
  else
    ...
 end
```

But thinking more about it, as the logic growths, I see more and more justification to just extract a Service Object like `ResendUserInvitation(invited_user/email, inviting_user)` that could wrap all this logic + the current `User.resend_invite` logic.

Currently the Job is short of emulating it (besides the bit in the `User.resend_invite`).

Or even looking further we could have a `SendUserInvitation` service that even deals with absolutely every case and check given a few parameters (email, inviting_user...) and just have a single place dealing with all this possibilities.
Eg: "If inviting user email and given email are the same then is the own user requesting re-invitation"...

slorek:
I agree with moving the business logic code out of the controller, but once the job starts implementing business logic it's time to encapsulate that within a service.

psd-web/app/jobs/send_user_invitation_job.rb
```diff
@@ -3,11 +3,27 @@
...
 
-  def perform(user_id, user_inviting_id)
+  # If the user_inviting_id is present, it is implied that the invitation
+  # is being sent (or resent) by a colleague, and so the invited_at time
+  # is reset.
+  #
+  # If absent, it is implied that the user themselves requested a resend of the
+  # invitation (eg by attempting to reset their password), and so the invited_at
+  # time should not be altered, and a different email should be sent if the invitation
+  # has already expired.
```
scruti: Nice explanation of the reason behind the code implementation :clap:

psd-web/app/jobs/send_user_invitation_job.rb
```diff
@@ -3,11 +3,27 @@
...
+  def perform(user_id, user_inviting_id = nil)
```
scruti: Suggestion, should we use a named param for the `user_inviting` ?

frankieroberto:
@slorek I personally see the job classes as a type of service object (just happens to implement a slightly different API, using the `#perform` method).

We _could_ refactor again so that the job just delegates to a different service object:

```ruby
class SendUserInvitationJob < ApplicationJob
  def perform(user_id, user_inviting_id = nil)
     SendUserInvitation.call(user_id, user_inviting_id = nil)
  end
end
```

...but I don’t see a huge benefit?

psd-web/app/jobs/send_user_invitation_job.rb
frankieroberto: I’m not entirely sure that keyword arguments work for jobs which are serialised to Sidekiq and back again, sadly.

psd-web/app/jobs/send_user_invitation_job.rb
frankieroberto: Heh, I figured it probably wasn’t obvious how the method arguments related to the different contexts in which this job could be called, so needed a detailed explanation... _Possibly_ a sign that should be two different classes (`SendInvitationForAnotherUser` and `ResendUserTheirOwnInvitation` or something), but they're so closely related, that might be even more confusing... 😬

PR #353, opened 2020-03-06
Fix error when deleting an attachment from a product

slorek:
https://trello.com/c/NrwtYep4/404-deleting-an-attachment-on-a-product-shows-an-error-message

## Description
Fixes an error which was exposed to the user when deleting an attachment from a product. This was caused by a missing `return` in the controller.

Fixes exceptions of type `ActiveRecord::AssociationTypeMismatch` as reported in Sentry. Example here: https://sentry.io/organizations/beis/issues/1544652422/?environment=prod&project=1329381&query=is%3Aunresolved

I have added a feature spec to cover this functionality, which is broadly similar to the existing delete attachment from case spec, which has been renamed. The test fails before the fix is applied.

## Checklist:


psd-web/spec/features/delete_attachment_from_product_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def expect_to_be_on_product_attachments_page
+    expect(page).to have_selector("h2", text: "Attachments")
+    expect(page).to have_selector("h2", text: document.title)
+    expect(page).to have_selector("p",  text: document.description)
+  end
+
+  def expect_to_be_on_remove_attachment_confirmation_page
+    expect(page).to have_current_path("/products/#{product.id}/documents/#{document.id}/remove")
```
frankieroberto: I've been trying to pass in dependent variables into helper methods, eg:

```ruby
def expect_to_be_on_remove_attachment_confirmation_page_for_document(document)
```

...mainly so that be shared between feature specs a bit more easily. That’s possibly a bit premature here, but it might be good to be consistent anyway?

slorek: I agree - these methods were copied and lightly altered from the existing test for case attachments so I didn't want to spend too much time refactoring here. We probably need to go over the existing tests and refactor to meet current standards, for example using full paths.

psd-web/app/controllers/documents_controller.rb
```diff
@@ -34,7 +34,7 @@ def remove; end
   # DELETE /documents/1
   def destroy
     @file.destroy
-    redirect_to @parent, flash: { success: "File was successfully removed" } unless @parent.is_a? Investigation
+    return redirect_to @parent, flash: { success: "File was successfully removed" } unless @parent.is_a? Investigation
```
scruti: ouch!

PR #354, opened 2020-03-06
Refactor health check to use controller within the app

frankieroberto:
This replaces the health check view at "/health/all" which is generated using the [OKComputer](https://github.com/sportngin/okcomputer) gem with an equivalent view generated within the app from a new `HealthController` class.

The rationale behind this is to enable any exceptions to be caught by the application in the usual way, so they can be reported by tools such as Sentry.

The checks are the same as previously, except for:

`Keycloak::Internal.get_users` has been replaced with `KeycloakClient.instance.all_organisations` as there are less organisations than users, so it should return faster.  We'll be able to remove this soon too.

I'm also not entirely convinced that checking the sidekiq queue latency should be part of the health check (and reported as part of our uptime metrics). I think it would be better to remove this, and monitor the queue latency separately. We could do that a future PR though.

See https://trello.com/c/jY3vkces/402-improve-healthcheck-endpoint


psd-web/app/controllers/health_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+    # Check redis services
+    Redis.new(Rails.application.config_for(:redis_session)).info
+    Redis.new(Rails.application.config_for(:redis_store)).info
+
+    # Check database connection
+    ActiveRecord::Migrator.current_version
+
+    # Check Keycloak service is available
+    # TODO: remove once Keycloak is no longer required.
+    KeycloakClient.instance.all_organisations
```
slorek: Could catch a potential issue with no data being returned (albeit this will be short-lived):

```suggestion
    raise "No Keycloak organisation data found" if KeycloakClient.instance.all_organisations.count < 1
```

psd-web/app/controllers/health_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  http_basic_authenticate_with name: ENV.fetch("HEALTH_CHECK_USERNAME", "health"), password: ENV.fetch("HEALTH_CHECK_PASSWORD", "check")
+
+  def show
+    # Check redis services
+    Redis.new(Rails.application.config_for(:redis_session)).info
+    Redis.new(Rails.application.config_for(:redis_store)).info
+
+    # Check database connection
+    ActiveRecord::Migrator.current_version
```
slorek: I'm not sure if creating a new connection is better than querying the existing one, e.g. `Rails.application.config.session_store` and `Rails.application.config.cache_store` and `ActiveRecord::Base.connection`? We might get false positives if we hit a connection pool limit.

psd-web/app/controllers/health_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    # Check Keycloak service is available
+    # TODO: remove once Keycloak is no longer required.
+    KeycloakClient.instance.all_organisations
+
+    # Check Elasticsearch cluster health
+    raise "Elastic search is down" if Elasticsearch::Client.new(Rails.application.config_for(:elasticsearch)).cluster.health[:status] == "red"
```
slorek:
```suggestion
    raise "Elasticsearch is down" if Elasticsearch::Client.new(Rails.application.config_for(:elasticsearch)).cluster.health[:status] == "red"
```
Same potential issue with connection limits as well.

psd-web/app/controllers/health_controller.rb
frankieroberto: I'm not sure either – I just copied and pasted the existing behaviour from the okcomputer gem. 🤷‍♂

PR #368, opened 2020-03-11
bump ruby to 2.6.5

StupidCodeFactory:
## Description
Simple bump to ruby 2.6.5, should really be seamless, and we should see an improvement in terms of performance, not that it matters much.

I've not bumped to 2.7.0 there was way to many warnings about the new changes made to keyword arguments. Rails libraries were spitting loads of warnings, enough to be annoying.

## Checklist:


slorek:
Thanks for doing this. Could you please make use of the checklist.

PR #375, opened 2020-03-12
Incude a space before the visually hidden text within actions

frankieroberto:
This makes sure that a space is inserted before the visually-hidden text within action links in the summary list, so that screen-reader software announces the links correctly, eg as "Change name" instead of "Changename".

See https://trello.com/c/y1pqZ7yr/420-bugfix-change-links-on-summary-lists-check-your-answers-pages-dont-contain-a-space-between-action-and-hidden-text


slorek:
This seems like unexpected behaviour 🤷‍♂

PR #377, opened 2020-03-12
Update cookies banner to fix accessibility issues

frankieroberto:
Fixes 2 accessibility issues within the cookies banner:

* Makes the blue links darker, so that they have higher contrast against the light blue background colour. (See [Trello card](https://trello.com/c/BhpqZVSb/417-accessibility-issue-improve-contrast-of-the-links-in-the-cookies-banner) )
* Adds some visually-hidden text so that the "hide this message" link makes sense when announced out of context. (see [Trello card](https://trello.com/c/xCHsFhOk/418-accessibility-issue-make-the-dismiss-cookies-banner-link-make-sense-of-out-context) )

Also hides the "hide this message" link when javascript disabled, as it requires javascript to work.

Note: I've moved this view from the [shared govuk-design-system-rails repo](https://github.com/UKGovernmentBEIS/govuk-design-system-rails/blob/master/app/views/components/_govuk_cookie_banner.html.slim) to within the application, as this isn't part of the GOV.UK Design System, and feels more application-specific.

### DIFF
--- a/psd-web/app/views/layouts/_cookie_banner.html.erb
+++ b/psd-web/app/views/layouts/_cookie_banner.html.erb
@@ -7,7 +7,7 @@
         or
         <%= link_to "#", id: "hideLink", class: "govuk-link" do %>
           hide this
-          <span class="govuk-visually-hiddenx">cookie</span>
+          <span class="govuk-visually-hidden">cookie</span>
           message
         <% end %>
       </span>
--- a/psd-web/app/views/layouts/_cookie_banner.html.erb
+++ b/psd-web/app/views/layouts/_cookie_banner.html.erb
@@ -1,0 +1,16 @@
+<% if !cookies[:seen_cookie_message] %>
+  <div class="cookie-banner" id="global-cookie-message">
+    <p class="cookie-banner__message govuk-width-container">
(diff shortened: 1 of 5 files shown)


psd-web/app/views/layouts/_cookie_banner.html.erb
```diff
@@ -0,0 +1,16 @@
+<% if !cookies[:seen_cookie_message] %>
+  <div class="cookie-banner" id="global-cookie-message">
+    <p class="cookie-banner__message govuk-width-container">
+      GOV.UK uses cookies to make the site simpler.
+      <%= link_to "Find out more about cookies", help_privacy_notice_path(anchor: "cookies-policy"), class: "govuk-link" %>
+      <span class="js-only">
+        or
+        <%= link_to "#", id: "hideLink", class: "govuk-link" do %>
+          hide this
+          <span class="govuk-visually-hiddenx">cookie</span>
```
edwardhorsford:
```suggestion
          <span class="govuk-visually-hidden">cookie</span>
```

frankieroberto: Opps, well spotted!

PR #382, opened 2020-03-13
Show error summary when adding comment

frankieroberto:
The "add comment to a case" flow currently doesn’t use the standard error summary component, and instead uses a flash message. This fixes that.

Also makes sure that `Error: ` is prefixed the the page `<title>` tag.

See https://trello.com/c/S40mJNig/425-add-comment-to-case-doesnt-show-an-error-summary-when-there-are-validation-errors

## Before

Screenshot 2020-03-13 at 15 18 16

## After

Screenshot 2020-03-13 at 16 00 40

### DIFF
--- a/psd-web/app/controllers/comments_controller.rb
+++ b/psd-web/app/controllers/comments_controller.rb
@@ -2,22 +2,26 @@ class CommentsController < ApplicationController
   before_action :set_investigation
 
   def create
-    activity = @investigation.activities.new(comment_activity_params.merge(investigation: @investigation))
-    activity.source = UserSource.new(user: current_user)
+    @activity = @investigation.activities.new(comment_activity_params.merge(investigation: @investigation))
+    @activity.source = UserSource.new(user: current_user)
 
     respond_to do |format|
-      if activity.save
+      if @activity.save
         format.html do
           redirect_to investigation_url(@investigation), flash: { success: "Comment was successfully added." }
         end
         format.json { render :show, status: :created, location: @activity }
       else
         format.html do
-          redirect_to new_investigation_activity_comment_url(@investigation), flash: { warning: "Comment was not successfully added." }
+          render :new
         end
-        format.json { render json: activity.errors, status: :unprocessable_entity }
+        format.json { render json: @activity.errors, status: :unprocessable_entity }
       end
     end
+  end
+
+  def new
+    @activity = @investigation.activities.new
   end
 
 private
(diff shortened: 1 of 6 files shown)


edwardhorsford:
Looks good!

I note that the textarea has a hidden label - whereas we could make the H1 the label. Up to you if you want to fix as part of this though.

psd-web/app/controllers/comments_controller.rb
```diff
@@ -2,24 +2,28 @@ class CommentsController < ApplicationController
   before_action :set_investigation
 
   def create
-    activity = @investigation.activities.new(comment_activity_params.merge(investigation: @investigation))
-    activity.source = UserSource.new(user: current_user)
+    @comment = @investigation.activities.new(comment_activity_params.merge(investigation: @investigation))
+    @comment.source = UserSource.new(user: current_user)
```
StupidCodeFactory: Maybe avoid a mix of assignment styles, do either:
```ruby
@investigation.activities.new(
  comment_activity_params.merge(
    investigation: @investigation, 
    source: UserSource.new(user: current_user)
  )
)
```
or 
```ruby
@comment = @investigation.activities.new(comment_activity_params)
@comment.investigation = @investigation
@comment.source = UserSource.new(user: current_user)
```

psd-web/app/controllers/comments_controller.rb
```diff
@@ -2,24 +2,28 @@ class CommentsController < ApplicationController
...
+      if @comment.save
         format.html do
           redirect_to investigation_url(@investigation), flash: { success: "Comment was successfully added." }
         end
-        format.json { render :show, status: :created, location: @activity }
+        format.json { render :show, status: :created, location: @comment }
       else
         format.html do
-          redirect_to new_investigation_activity_comment_url(@investigation), flash: { warning: "Comment was not successfully added." }
+          render :new
```
StupidCodeFactory: I think here we'd prefer a redirect, @edwardhorsford suggested following the `post - redirect - get`. I personally don't mind, I prefer avoid redirect whenever possible.

frankieroberto: We should definitely use "post - redirect - get" for successful form submissions, but for invalid submissions I think it's probably ok (if not ideal) to render, as it’s tricky to persist the validation errors (and the form data) through a redirect.

psd-web/app/controllers/comments_controller.rb
frankieroberto: Agree, this is weird...  was trying to avoid doing too much refactoring, but I think this is worth it.

psd-web/app/controllers/comments_controller.rb
StupidCodeFactory: Yeah I agree with you, probably why there was a flash message rather than the form validation errors.

PR #383, opened 2020-03-16
Improve error message display on document upload for products

frankieroberto:
This fixes some usability issues around the display of error messages on the document upload pages (for products).

Namely:

* The page title is now prefixed with `Error: ` if there are errors present
* The link from the error summary now works (focuses the file file)
* The file upload field has has a red border to the left and repeats the error message where there are errors.

Issues raised in the Accessibility Audit under the ID `DAC_usability_issue5`.

See https://trello.com/c/7HDbwwnD/259-file-uploads-missing-error-styling-and-error-messages

## Before

Screenshot 2020-03-16 at 11 44 19

## After

Screenshot 2020-03-16 at 11 44 41

## Implementation / review notes

* I converted the template from Slim to ERb as a first commit. This may the overall diff harder to read, so it might be worth reviewing commit-by-commit.
* There’s already a feature spec (`add_attachment_spec.rb`) covering this feature (including the validation error). Whilst it doesn’t cover all the subtleties of the UX (the `Error: ` prefix and so on), I’m not sure how valuable it’d be to add these extra assertions right now?
* I decided to avoid making any changes to the controller code, even though it could definitely be improved, as it’s fairly complex and hard to follow (involving a shared `DocumentsHelper` and `FileConcern`). We could tackle this as a future task.


psd-web/app/views/documents_flow/upload.html.erb
```diff
@@ -0,0 +1,34 @@
+<% @parent = @parent.decorate %>
+<% title = "Add attachment" %>
+<% content_for :page_title, title_for(OpenStruct.new(errors: @errors), "Add attachment") %>
```
frankieroberto: This struct is a bit of hack, required because `title_for` expects an activerecord object (reasonably enough), but for whatever reason, the existing `DocumentsController` code builds an isolated `ActiveModel::Errors` object (`@errors`) not attached to any model.

I’d like to refactor this at a future point to make the errors actually attached to the model object they describe, but that’s a bigger job.

psd-web/app/views/documents_flow/upload.html.erb
```diff
@@ -0,0 +1,34 @@
...
+    <h1 class="govuk-heading-l"><%= title %></h1>
+
+    <%= form_with scope: :document, url: wizard_path, method: :put, local: true, multipart: true do |form| %>
+
+      <% if @errors.any?
+        errorMessage = {
+          text: @errors.full_messages.join(", ")
+        }
+        end
+      %>
```
frankieroberto: With the existing controller code, there should only ever be one error message present, so I was tempted to only display the first one (`@errors.full_messages.first`) but including them all, comma-separated seems more future-proof. Possibly we should use `to_sentence`?

Any suggestions for better ways to build this hash only if there are any errors welcome.

psd-web/app/views/documents_flow/upload.html.erb
```diff
@@ -0,0 +1,34 @@
...
+
+      <%= govukFileUpload(
+        id: "base",
```
frankieroberto: `base` is required to be the id of the file upload field here, as the error messages are added to a `base` key in the controller, and so that is what the `error_summary` helper links to.

psd-web/app/views/documents_flow/upload.html.erb
StupidCodeFactory: yeah use `to_sentence(last_word_connector: " and ")`

psd-web/app/views/documents_flow/upload.html.erb
```diff
@@ -0,0 +1,34 @@
...
+        name: "document[file][file]",
+        label: {
+          text: "Upload a file"
+        }, errorMessage: errorMessage
```
StupidCodeFactory: seem this could be indented to new line to be consistent with the rest of the keys

psd-web/app/views/documents_flow/upload.html.erb
StupidCodeFactory: This part of the code is absolutely horrendous... Hopefully we can trash this sometime. 

It's actually far more sensible to decouple the helper from the ActiveRecord object and have it passed an ActiveModel::Errors  to avoid future problems like this.
Not all objects to be displayed will be database records, they could simply be ActiveModel object, or maybe just plain ruby object that have ActiveModel::Errors.

psd-web/app/views/documents_flow/upload.html.erb
```diff
@@ -0,0 +1,34 @@
...
+    <span class="govuk-caption-l"><%= @parent.pretty_description %></span>
+    <h1 class="govuk-heading-l"><%= title %></h1>
+
+    <%= form_with scope: :document, url: wizard_path, method: :put, local: true, multipart: true do |form| %>
+
+      <% if @errors.any?
+        errorMessage = {
+          text: @errors.full_messages.join(", ")
+        }
+        end
```
StupidCodeFactory: maybe you can extract some of the code in this helper https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/382/files#diff-3a9651b5997ab974e9d52961cd84ed6bR8-R13 and avoid this `if` statement

psd-web/app/views/documents_flow/upload.html.erb
frankieroberto: Given that all `title_for` needs to do is know whether there _are_ any errors or not, I'd be tempted to refactor it so that we’re just passing it a truthy value:

```ruby
def page_title(title, errors: false)  
   title = "Error: #{title}" if errors
   title
 end
```

then in most pages we could call it using

```ruby
  content_for(:page_title, page_title(title, errors: thing.errors.any?))
```

StupidCodeFactory: that's a good idea, I'd add the call to content_for insed the page_title:
```ruby
def page_title(title, errors: false)
  title = "Error: #{title}" if errors

  content_for(:page_title, title)
end
```

frankieroberto: I kinda thought it was slightly unexpected to call `content_for` from a helper. But I don't feel strongly, and I guess it reduces code in the templates...

psd-web/app/views/documents_flow/upload.html.erb
frankieroberto: I've moved it inline using a ternary. Just about easier to follow, I think.

PR #396, opened 2020-03-18
Include error prefix in page titles

frankieroberto:
Adding an "Error: " prefix to the `<title>` tag on pages with validation errors, so that screen readers announce the error as soon as possible. See [GOV.UK Design System pattern on validation](https://design-system.service.gov.uk/patterns/validation/).

To implement this, I've switched all views to use a new `page_title` helper, which accepts an optional `error` boolean that adds the prefix if truthy. The helper also includes the `content_for` call to DRY the views.

➡️ [Trello card](https://trello.com/c/F9rSFMOo/428-include-error-prefix-in-page-titles-where-there-are-validation-errors)


psd-web/app/views/businesses/show.html.slim
```diff
@@ -1,4 +1,4 @@
-- content_for :page_title, "Business - #{@business.trading_name}"
+= page_title"Business - #{@business.trading_name}"
```
StupidCodeFactory:
```suggestion
= page_title "Business - #{@business.trading_name}"
```

frankieroberto: Opps, 👍

psd-web/app/views/investigations/activities/new.html.slim
```diff
@@ -1,4 +1,4 @@
-- content_for :page_title, "Add activity"
+= page_title "Add activity", errors: @activity_type_empty.present?
```
edwardhorsford: Curious why this isn't doing `.errors.any?` like the others?

psd-web/app/views/investigations/corrective_actions/confirmation.html.slim
```diff
@@ -1,5 +1,5 @@
 - page_title = "Confirm corrective action details"
```
edwardhorsford: Not directly part of this pr, but we're inconsistent with how we declare page titles and then pass to the helper. Here it's `page_title`, in other cases it's `title`, in other cases it's directly passed through.

psd-web/app/views/investigations/created.html.slim
```diff
@@ -1,6 +1,6 @@
 - page_heading = "Case created"
```
edwardhorsford: `page_heading` here - inconsistent with others.

psd-web/app/views/investigations/ts_investigations/_repeatable_info.html.slim
```diff
@@ -1,6 +1,6 @@
 - page_heading = title
```
edwardhorsford: Urrgh this circular title > heading > title. Something to clean up later.

psd-web/app/views/investigations/activities/new.html.slim
frankieroberto: This page seems to handle errors somewhat differently... not sure why. One to investigate at another point!

psd-web/app/views/investigations/corrective_actions/confirmation.html.slim
frankieroberto: Yeah, I was tempted to move all page title strings to be variables, so that they can be used within the `<h1>`s, but that seemed a little bit out of scope for this. Worth doing at another point though.

psd-web/app/views/investigations/ts_investigations/_repeatable_info.html.slim
frankieroberto: Yeah it'd be good to come back and make all these variable names consistent...

PR #402, opened 2020-03-19
Enable 2FA when user attempts to reset password

scruti:
Trello ticket: https://trello.com/c/rwtnglov/427-2fa-on-forgot-password

When an user reaches the reset password page after receiving the link by email, it will be redirected to the 2FA page as a security check before being able to reset its password.

In order to achieve this we use the following process:
1 - Anonymous user lands at "edit password" and its email gets identified.
2 - The user gets signed in based on its email address.
3 - User session gets marked as "needs 2FA".
4 - A 2FA code is sent to the user.
5 - User gets redirected to the 2FA page so can introduce its code.
6 - Once the correct code is introduced, the user gets redirected back to the "Edit password" page keeping the same "reset password token" that had in the reset password email link.
7 - When landing as a completely signed in user, the user session gets destroyed so the user depends only int the "reset password token" for updating the password.
8 - User can update its password.

## Security and implementation considerations
The solution partially signs in the user in order to go through the 2FA authentication.
Users partially signed in with "needs 2FA" state won't be able to navigate outside 2FA page until the correct code its introduced.
When being redirected back to the edit password page, the user session gets automatically destroyed, so should be safe.

## Overloaded functionality
No new overloaded methods introduced but all this behaviour  heavily depends on overloading devise methods and setting values for Warden user session.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/psd-web/app/controllers/users/passwords_controller.rb
+++ b/psd-web/app/controllers/users/passwords_controller.rb
@@ -66,7 +66,10 @@ module Users
     def passed_two_factor_authentication?
       return true if !Rails.configuration.two_factor_authentication_enabled
 
-      current_user && is_fully_authenticated?
+      request.referer &&
+        URI(request.referer).path == user_two_factor_authentication_path &&
+        current_user &&
+        is_fully_authenticated?
     end
 
     def resend_invitation_link_for(user)
--- a/psd-web/app/controllers/users/passwords_controller.rb
+++ b/psd-web/app/controllers/users/passwords_controller.rb
@@ -1,10 +1,29 @@
 module Users
   class PasswordsController < Devise::PasswordsController
-    skip_before_action :assert_reset_token_passed, only: :edit
+    skip_before_action :assert_reset_token_passed,
+                       :require_no_authentication,
+                       :has_accepted_declaration,
+                       :has_viewed_introduction,
+                       only: :edit
 
+
+    # TO DO: Hide user navigation header?
+    # TO DO: Do we really want to skip declaration/introduction? Done for tests. Need to figure out desired behaviour.
+    # TO DO: Is this safe to do?
+    # TO DO: Refactor/Extract 2FA logic bit
     def edit
-      return render :invalid_link, status: :not_found if params[:reset_password_token].blank? || reset_token_invalid?
-      return render :expired, status: :gone if reset_token_expired?
+      # TO DO: Check if this conditional is correct
+      if !current_user || (current_user && !is_fully_authenticated?)
+        return render :invalid_link, status: :not_found if params[:reset_password_token].blank? || reset_token_invalid?
+        return render :expired, status: :gone if reset_token_expired?
+
+        sign_in(user_with_reset_token)
+        warden.session(:user)[TwoFactorAuthentication::NEED_AUTHENTICATION] = true
+        warden.session(:user)["two_factor_authentication_referrer"] = request.path
+        user_with_reset_token.send_new_otp
+
+        return redirect_to user_two_factor_authentication_path
+      end
 
       super
     end
--- a/psd-web/app/controllers/users/passwords_controller.rb
+++ b/psd-web/app/controllers/users/passwords_controller.rb
@@ -7,7 +7,7 @@ module Users
                        only: :edit
 
     def edit
-      return render :invalid_link, status: :not_found if params[:reset_password_token].blank? || reset_token_invalid?
+      return render :invalid_link, status: :not_found if reset_token_invalid?
       return render :expired, status: :gone if reset_token_expired?
 
       if passed_two_factor_authentication?
@@ -67,7 +67,7 @@ module Users
     end
 
     def reset_token_invalid?
-      user_with_reset_token.blank?
+      params[:reset_password_token].blank? || user_with_reset_token.blank?
     end
 
     def reset_token_expired?
--- a/psd-web/app/controllers/users/passwords_controller.rb
+++ b/psd-web/app/controllers/users/passwords_controller.rb
@@ -6,24 +6,24 @@ module Users
                        :has_viewed_introduction,
                        only: :edit
 
-    # TO DO: Do we really want to skip declaration/introduction? Done for tests. Need to figure out desired behaviour.
-    # TO DO: Is this safe to do?
-    # TO DO: Refactor/Extract 2FA logic bit
     def edit
-      # TO DO: Check if this conditional is correct
-      if !current_user || (current_user && !is_fully_authenticated?)
-        return render :invalid_link, status: :not_found if params[:reset_password_token].blank? || reset_token_invalid?
-        return render :expired, status: :gone if reset_token_expired?
+      return render :invalid_link, status: :not_found if params[:reset_password_token].blank? || reset_token_invalid?
+      return render :expired, status: :gone if reset_token_expired?
 
+      if passed_two_factor_authentication?
+        # Password update is based on the "reset_password_token", not in the user session.
+        sign_out(:user)
+        super
+      else
         sign_in(user_with_reset_token)
         warden.session(:user)[TwoFactorAuthentication::NEED_AUTHENTICATION] = true
-        store_location_for(:user, edit_user_password_path)
+        # Will redirect back to #edit after passing 2FA.
+        # fullpath contains "reset_password_token", necessary for the further update.
+        store_location_for(:user, request.fullpath)
         user_with_reset_token.send_new_otp
 
-        return redirect_to user_two_factor_authentication_path
+        redirect_to user_two_factor_authentication_path
       end
-
-      super
     end
 
     def create
@@ -54,6 +54,12 @@ module Users
     end
 
   private
+
+    def passed_two_factor_authentication?
+      return true if !Rails.configuration.two_factor_authentication_enabled
+
+      current_user && is_fully_authenticated?
+    end
 
     def resend_invitation_link_for(user)
       SendUserInvitationJob.perform_later(user.id, nil)
--- a/psd-web/app/controllers/users/passwords_controller.rb
+++ b/psd-web/app/controllers/users/passwords_controller.rb
@@ -68,7 +68,7 @@ module Users
 
       request.referer &&
         URI(request.referer).path == user_two_factor_authentication_path &&
-        current_user &&
+        user_signed_in? &&
         is_fully_authenticated?
     end
 
--- a/psd-web/app/controllers/users/passwords_controller.rb
+++ b/psd-web/app/controllers/users/passwords_controller.rb
@@ -66,10 +66,7 @@ module Users
     def passed_two_factor_authentication?
       return true if !Rails.configuration.two_factor_authentication_enabled
 
-      request.referer &&
-        URI(request.referer).path == user_two_factor_authentication_path &&
-        user_signed_in? &&
-        is_fully_authenticated?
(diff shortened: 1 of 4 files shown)


psd-web/app/controllers/users/two_factor_authentication_controller.rb
```diff
@@ -68,7 +72,11 @@ def after_two_factor_success_for(resource)
 
       resource.pass_two_factor_authentication!
 
-      redirect_to after_two_factor_success_path_for(resource)
+      if referrer
+        redirect_to referrer
+      else
+        redirect_to after_two_factor_success_path_for(resource)
+      end
```
frankieroberto: Can you use [`store_location_for`](https://github.com/heartcombo/devise/blob/715192a7709a4c02127afb067e66230061b82cf2/lib/devise/controllers/store_location.rb#L36) in the passwords controller to avoid using the referrer here? `after_two_factor_success_path_for` seems to use that by default.

psd-web/app/controllers/users/passwords_controller.rb
```diff
@@ -1,10 +1,29 @@
...
-      return render :invalid_link, status: :not_found if params[:reset_password_token].blank? || reset_token_invalid?
-      return render :expired, status: :gone if reset_token_expired?
+      # TO DO: Check if this conditional is correct
+      if !current_user || (current_user && !is_fully_authenticated?)
+        return render :invalid_link, status: :not_found if params[:reset_password_token].blank? || reset_token_invalid?
+        return render :expired, status: :gone if reset_token_expired?
+
+        sign_in(user_with_reset_token)
+        warden.session(:user)[TwoFactorAuthentication::NEED_AUTHENTICATION] = true
+        warden.session(:user)["two_factor_authentication_referrer"] = request.path
```
frankieroberto: Could you do something like this:

```suggestion
        store_location_for(:user, user_edit_password_path)
```
?

psd-web/app/controllers/users/two_factor_authentication_controller.rb
scruti: Best tip ever. Didn't know about `store_location_for`. Makes things so much easier :ok_hand:

psd-web/app/controllers/users/two_factor_authentication_controller.rb
```diff
@@ -49,6 +49,10 @@ def otp_code_param
       @otp_code_param ||= resource_params.permit(:otp_code)[:otp_code]
     end
 
+    def referrer
+      warden.session(:user)["two_factor_authentication_referrer"]
+    end
+
```
frankieroberto: Think you can remove this now?

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -134,10 +151,49 @@ def request_password_reset
     end
   end
 
+  # TO DO: Resolve this awful lazy duplication
+  def request_password_reset_with_2fa
+    user.update!(reset_password_token: reset_token)
```
frankieroberto: Feels odd to be doing this within a feature spec. Can't you request the password reset link via the user interface?

psd-web/app/controllers/users/passwords_controller.rb
```diff
@@ -1,10 +1,30 @@
...
```
frankieroberto: Can you remove this line now too?

scruti: wops!

psd-web/spec/features/password_reset_spec.rb
scruti: I think this has something to do with the token being hashed by Devise (hence the stub). But will double check if we can avoid it.

psd-web/app/controllers/users/passwords_controller.rb
```diff
@@ -32,15 +49,25 @@ def update
...
+    def hide_nav?
+      true
+    end
+
   private
 
+    def passed_two_factor_authentication?
+      return true if !Rails.configuration.two_factor_authentication_enabled
+
+      current_user && is_fully_authenticated?
```
StupidCodeFactory: What are you trying to defend against? `is_fully_authenticated?` will take care of no that. If there is no current user the user can not be fully authenticated

psd-web/spec/requests/user_resets_password_spec.rb
```diff
@@ -1,25 +1,39 @@
...
-        get(edit_user_password_path(reset_password_token: reset_token))
       end
 
-      it "return a 200 status code" do
-        expect(response).to have_http_status(:ok)
+      # rubocop:disable RSpec/AnyInstance
+      context "when user already passed two factor authentication" do
+        before do
+          allow_any_instance_of(Users::PasswordsController).to receive(:current_user).and_return(user)
+          allow_any_instance_of(Users::PasswordsController).to receive(:is_fully_authenticated?).and_return(true)
```
StupidCodeFactory: :flushed: Why not `sign_in(user)`? 

The fact that you need to use `allow_any_instance_of` is  a clear sign this has to be tested using a feature test. 
Being fully authenticated requires at least 2 requests, /sign-in and /two-fatctor , and then another 3 request to edit the password. Request specs are great for API or anything stateless for anything else they're pretty poor.

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -134,10 +143,25 @@ def request_password_reset
     end
   end
 
+  def complete_two_factor_authentication
+    expect_to_be_on_two_factor_authentication_page
+
+    fill_in "Enter security code", with: user.reload.direct_otp
+    click_on "Continue"
+  end
```
frankieroberto: I suggest moving this expectation to the scenario, instead of having it within the helper method.

I think elsewhere we've also suggested passing in the values to fill in to the helper methods, rather than looking them up from the helper method, eg:

```ruby
def complete_two_factor_authentication_with_code(security_code)
  fill_in "Enter security code", with: security_code
  click_on "Continue"
end  
```

and then using this within the scenario:

```ruby
complete_two_factor_authentication_with_code(user.reload.direct_otp)
```

psd-web/app/controllers/users/passwords_controller.rb
```diff
@@ -32,15 +49,25 @@ def update
       end
     end
 
+    def hide_nav?
+      true
+    end
```
frankieroberto: Maybe add a quick note to explain what this does (as it's kinda unconventional). Hopefully we can refactor this into an instance variable or something soon.

psd-web/app/controllers/users/passwords_controller.rb
```diff
@@ -1,12 +1,29 @@
...
 
     def edit
-      return render :invalid_link, status: :not_found if params[:reset_password_token].blank? || reset_token_invalid?
+      return render :invalid_link, status: :not_found if reset_token_invalid?
       return render :expired, status: :gone if reset_token_expired?
 
-      super
+      if passed_two_factor_authentication?
+        # Password update is based on the "reset_password_token", not in the user session.
+        sign_out(:user)
```
frankieroberto: Feel like we need a comment to explain why we're signing out the user at this point...

psd-web/spec/requests/user_resets_password_spec.rb
scruti: The idea was to just set the preconditions that would allow the user to land in the "edit" page as the same controller action send to edit/2fa depending if the user already got authenticated.
The current feature specs already cover it so happy to remove this context for simplicity.

psd-web/app/controllers/users/passwords_controller.rb
scruti: that there is a logged user and does not have pending 2FA.
We will use this check to allow the user to edit its password.

StupidCodeFactory: I wasn't clear but you can remove the call to `current_user` also to perform this check devise has the `user_signed_in?` method but it's redundant as `is_fully_signed_in?` would do all of that.

StupidCodeFactory: Here's a link to the documentation
https://www.rubydoc.info/github/plataformatec/devise/Devise%2FControllers%2FHelpers.define_helpers

scruti: problem is that `is_fully_authenticated?` returns `true` if there is no user session, so need to check that there is a session and given there is a session that does not need 2FA authentication.

`user_signed_in?` happens to be an alias to `!!current_user`, but the name seems more intuitive so will use it :+1:

StupidCodeFactory: yeah looking at the implementation I'm not surprised, this gem is really terrible :rofl:
```ruby
def is_fully_authenticated?
  !session["warden.user.user.session"].try(:[], TwoFactorAuthentication::NEED_AUTHENTICATION)
end
```

slawosz:
@scruti should reset password functionality log out all devices as well?

psd-web/app/controllers/users/passwords_controller.rb
```diff
@@ -1,12 +1,35 @@
...
-      super
+      if passed_two_factor_authentication?
+        # Devise password update requires the user to be signed out, as it relies on the "reset password token"
+        # parameter and an user attempting to reset its password because does not remember it is not supposed to be
+        # already signed in.
+        # In order to be able to enforce 2FA, we need to to sign the user in.
+        # Given this contradiction between needing it for 2FA but needing the opposite for the password update,
+        # when the user gets redirected back after 2FA, we have to sign out the users before allowing them to
+        # update their password.
+        sign_out(:user)
```
frankieroberto: Good explanation!

psd-web/app/controllers/users/passwords_controller.rb
```diff
@@ -32,15 +55,30 @@ def update
...
+      true
+    end
+
   private
 
+    def passed_two_factor_authentication?
+      return true if !Rails.configuration.two_factor_authentication_enabled
+
+      request.referer &&
+        URI(request.referer).path == user_two_factor_authentication_path &&
```
slorek: `referer` is unreliable as it is client supplied data. Some proxies/private browsing modes strip it from the request headers. Do we need to check this?

edwardhorsford:
When you get past the 2fa page to the password entry page, if you refresh the page it takes you back to 2fa again. Is this expected?

psd-web/app/controllers/users/passwords_controller.rb
frankieroberto: I wondered about this too. I _think_ generally proxies and private browsing mode might still send it for same-origin referers, but it still might not be 100% safe.

Save some data to the session instead?

scruti:
Clarified refresh page behaviour with @edwardhorsford , caused by the refresh request taken as a new landing from the email and re-initiating the 2FA process.
Tried to avoid it by allowing sessions to be kept when submitting a new password. But caused issues on the Devise password update behaviour.
So leaving it as it is as the lesser evil.

psd-web/app/controllers/users/passwords_controller.rb
scruti: Giving it a second thought, if a signed user reaches this point will have passed the check over its `password_reset_token` in the URL, so someone cannot just manually type the path and access the password reset.
Going to remove the referrer check :+1:

PR #406, opened 2020-03-20
Password manager login

slorek:
https://trello.com/c/5qErVTcP/438-improve-chances-of-passwords-getting-saved-correctly-autocomplete

## Description
Allows password managers to capture the username when completing registration for an account.

We did not want the username to be visible to the user.

Hidden inputs were ignored, as were elements inside a `govuk-visually-hidden` class. Multiple iterations were made to arrive at this solution, which is tested in Chrome and Firefox with 1Password, OSX Keychain and the Okta Password Manager on a BEIS managed laptop. It is not accessible via keyboard or screen readers.

## Checklist:

- [x] Works keyboard only

### DIFF
--- a/psd-web/app/assets/stylesheets/main.scss
+++ b/psd-web/app/assets/stylesheets/main.scss
@@ -58,8 +58,3 @@ $govuk-assets-path: '~govuk-frontend/govuk/assets/';
   padding-top: 2em;
   border-top: 1px solid govuk-colour("light-grey");
 }
-
-.app-\!-password-manager-hidden {
-  opacity: 0;
-  height: 0;
-}
--- a/psd-web/app/assets/stylesheets/main.scss
+++ b/psd-web/app/assets/stylesheets/main.scss
@@ -61,5 +61,5 @@ $govuk-assets-path: '~govuk-frontend/govuk/assets/';
 
 .app-\!-password-manager-hidden {
   opacity: 0;
-  height: 0px;
+  height: 0;
 }
--- a/psd-web/app/assets/stylesheets/main.scss
+++ b/psd-web/app/assets/stylesheets/main.scss
@@ -58,3 +58,8 @@ $govuk-assets-path: '~govuk-frontend/govuk/assets/';
   padding-top: 2em;
   border-top: 1px solid govuk-colour("light-grey");
 }
+
+.app-\!-password-manager-hidden {
+  opacity: 0;
+  height: 0px;
+}
--- a/psd-web/app/views/users/complete_registration.html.erb
+++ b/psd-web/app/views/users/complete_registration.html.erb
@@ -14,22 +14,25 @@
           classes: "govuk-visually-hidden"
         },
         label: { text: "Email address" },
-        value: @user.email
+        value: @user.email,
+        autocomplete: "off"
         %>
 
       <%= render "form_components/govuk_input", key: :name, form: form,
         classes: "app-!-max-width-two-thirds",
         label: { text: "Full name" },
-        value: @user.name
+        value: @user.name,
+        autocomplete: "name"
         %>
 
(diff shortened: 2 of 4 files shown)


psd-web/app/views/users/complete_registration.html.erb
```diff
@@ -12,16 +12,22 @@
...
+        value: @user.name,
+        autocomplete: "name"
         %>
 
       <%= render "form_components/govuk_input", key: :mobile_number, form: form,
         classes: "app-!-max-width-two-thirds",
         label: { text: "Mobile number" },
-        value: @user.mobile_number
+        value: @user.mobile_number,
+        autocomplete: "tel-national"
```
frankieroberto: We should possibly add `type: "tel"` here too, see [telephone numbers pattern](https://design-system.service.gov.uk/patterns/telephone-numbers/) – think that might trigger the numeric keypad on mobile / tablets?

psd-web/app/views/users/complete_registration.html.erb
```diff
@@ -12,16 +12,22 @@
...
         %>
 
-      <%= password_input @user %>
+      <%= content_tag :div, style: "opacity: 0; height: 0px" do -%>
+        <%= text_field_tag :username, @user.email, "disabled" => true, "tabindex" => "-1", "aria-hidden" => true, autocomplete: "username"  %>
+      <% end -%>
```
frankieroberto: Nice! I take it that a hidden field (`type="hidden"`) didn't work?

psd-web/app/assets/stylesheets/main.scss
```diff
@@ -58,3 +58,8 @@ $govuk-assets-path: '~govuk-frontend/govuk/assets/';
   padding-top: 2em;
   border-top: 1px solid govuk-colour("light-grey");
 }
+
+.app-\!-password-manager-hidden {
+  opacity: 0;
+  height: 0;
+}
```
frankieroberto: Could you move this to `helpers/_password_manager_hidden.scss` or similar?

psd-web/app/views/users/complete_registration.html.erb
```diff
@@ -12,16 +12,28 @@
...
+        type: "tel"
         %>
 
-      <%= password_input @user %>
+      <%#
+        This field is to enable password managers to capture the username as
+        well as the password, but should not be visible to users, nor should
+        the controller process the input.
+      %>
+      <%= content_tag :div, class: "app-!-password-manager-hidden" do -%>
```
edwardhorsford: @frankieroberto might correct me, but I think using `!` in a class name is meant to imply it's an override using `! important`. Since that's not the case here, I think it should probably be `app-password-manager-hidden`.

slorek: it's a bit confusing. `app-` is supposed to be used to define custom components, I believe. I'm not sure this fits the criteria. https://design-system.service.gov.uk/get-started/extending-and-modifying-components/#avoid-overwriting-gov-uk-design-system-code

frankieroberto: I think that _might_ be a govuk-frontend convention, rather than a BEM convention? So I'm not entirely sure what it implies.

But I agree that it probably makes more sense to use `app-password-manager-hidden` here. Or even `app-hidden`?

edwardhorsford: I'd probably err against `app-hidden` - as this isn't a general purpose class. I might more expect `app-hidden` to do `display:none` or something.

edwardhorsford: @slorek I just see `app` as our namespace. Though we're inconsistently using `psd` or `app` or possibly also `opss` namespaces right now.

slorek: I've updated the class name.

PR #408, opened 2020-03-20
Add 2fa to registration flow

frankieroberto:
This requires users to complete a mobile phone number verification step (entering a code received via text message) when completing their account details after being invited to use the service.

Because entering name, password and mobile number is a previous step, the user can potentially be left in a state where they have set those details but haven't verified their mobile number. In this (hopefully rare) state, they:

* Can still go back and start their account setup process again (from the link in the e-mail), provided it’s within the expiry period.
* Cannot login - if they do, they get a generic error message (this prevents a user from ending up in a state where they have an active account but can't sign in because their mobile number is invalid).

See https://trello.com/c/Huo9xSPj/389-2fa-on-registration-page

## Implementation approach

To keep track of whether the mobile number has been verified (via 2FA), a new `mobile_number_verified` boolean attribute is added to the User model. This is set to true after a user completes 2FA. The sign in page checks that this is true. Until it becomes `true`, the invitation to set up an account remains valid.

## Alternative approaches 

* #398 doesn’t save the name, password and mobile number to the database at all until the 2FA is complete. Instead it saves it to the redis session store.
* #404 as above, but saving to the Rails cache instead (which I think is also backed by Redis).

## Checklist:

### DIFF
--- a/psd-web/app/controllers/users/sessions_controller.rb
+++ b/psd-web/app/controllers/users/sessions_controller.rb
@@ -18,6 +18,14 @@ module Users
 
       self.resource = warden.authenticate(auth_options)
 
+
+      # Stop users from signing in if they’ve not completed 2FA verification
+      # of their mobile number during account set up process.
+      if resource && !resource.mobile_number_verified
+        resource.errors.add(:email, I18n.t(:wrong_email_or_password, scope: "sign_user_in.email"))
+        return render :new
+      end
+
       if resource
         sign_in(resource_name, resource)
         return respond_with resource, location: after_sign_in_path_for(resource)
--- a/psd-web/app/controllers/users/sessions_controller.rb
+++ b/psd-web/app/controllers/users/sessions_controller.rb
@@ -18,7 +18,6 @@ module Users
 
       self.resource = warden.authenticate(auth_options)
 
-
       # Stop users from signing in if they’ve not completed 2FA verification
       # of their mobile number during account set up process.
       if resource && !resource.mobile_number_verified
--- a/psd-web/db/migrate/20200320144542_add_mobile_number_verified_to_users.rb
+++ b/psd-web/db/migrate/20200320144542_add_mobile_number_verified_to_users.rb
@@ -10,6 +10,5 @@ class AddMobileNumberVerifiedToUsers < ActiveRecord::Migration[5.2]
         end
       end
     end
-
   end
 end
--- a/psd-web/db/migrate/20200320144542_add_mobile_number_verified_to_users.rb
+++ b/psd-web/db/migrate/20200320144542_add_mobile_number_verified_to_users.rb
@@ -1,0 +1,15 @@
+class AddMobileNumberVerifiedToUsers < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured do
+      add_column :users, :mobile_number_verified, :boolean, default: false, null: false
+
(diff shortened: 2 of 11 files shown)


psd-web/db/migrate/20200320144542_add_mobile_number_verified_to_users.rb
```diff
@@ -0,0 +1,14 @@
+class AddMobileNumberVerifiedToUsers < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured do
+      add_column :users, :mobile_number_verified, :boolean, default: false, null: false
```
slawosz: How about making this `datetime`, this way we would know when it was verified?

psd-web/app/controllers/users/sessions_controller.rb
```diff
@@ -18,6 +18,14 @@ def create
 
       self.resource = warden.authenticate(auth_options)
 
+
```
scruti: :scissors:

psd-web/spec/features/creating_an_account_from_an_invitation_spec.rb
```diff
@@ -14,6 +14,28 @@
...
+    expect_to_be_on_the_homepage
+
+    click_link "Sign in to your account"
+
+    fill_in "Email address", with: invited_user.email
+    fill_in "Password", with: "testpassword123@"
+    click_on "Continue"
+
+    # Skips 2FA as cookie was set to not require
+    # 2FA for 7 days.
```
scruti: Nice explanation :clap:

psd-web/spec/requests/two_factor_authentication_spec.rb
```diff
@@ -67,6 +69,11 @@
           follow_redirect!
           expect(response.body).to include("Sign out")
         end
+
+        it "marks the mobile number as verified" do
+          submit_2fa
+          expect(user.reload.mobile_number_verified).to be true
```
scruti: Suggestion: 
Suggestion: As @StupidCodeFactory would say... Rspec provides nice helpers for this 
```suggestion
          expect(user.reload).to be_mobile_number_verified
```

psd-web/spec/requests/user_completes_registration_spec.rb
```diff
@@ -101,19 +113,23 @@
         expect(user).not_to be_account_activated
       end
 
-      it "redirects to the root path" do
-        expect(response).to redirect_to(root_path)
+      it "does not set the mobile number as verified" do
+        expect(user.mobile_number_verified).to be false
```
scruti: Suggestion (showing the possibility to be considered more than requesting any change):
```suggestion
        expect(user).not_to be_mobile_number_verified
```

psd-web/spec/requests/two_factor_authentication_spec.rb
frankieroberto: I personally like the rspec helpers, but only where they make linguistic sense. "be mobile number verified" feels a bit awkward and ungrammatical to me?

psd-web/db/migrate/20200320144542_add_mobile_number_verified_to_users.rb
frankieroberto: Could do. I can’t immediately think of a reason why we'd need to know the time it was verified, but perhaps it’d be useful.

PR #412, opened 2020-03-23
Prevent brute force on login page

slawosz:
Locks account after several attempts. Currently set to 10.

https://trello.com/c/KPrSkHF4/383-5-prevent-brute-forcing-of-passwords-at-sign-in


psd-web/app/controllers/users/sessions_controller.rb
```diff
@@ -16,6 +16,11 @@ def create
         return render :new
       end
 
+      user = User.find_by(email: sign_in_form.email)
```
slorek: Please change: Seems like this would be retrieved as `self.resource` one line later, and we could put this check alongside the 2FA one on line 26.

psd-web/spec/features/sign_in_spec.rb
```diff
@@ -56,6 +62,51 @@ def fill_in_credentials
     end
   end
 
+  context "when using wrong credentials over and over again", :with_2fa do
```
slorek: We can combine all three of these scenarios to improve performance as they are duplicating the same steps.

psd-web/spec/features/sign_in_spec.rb
```diff
@@ -56,6 +62,51 @@ def fill_in_credentials
...
+    scenario "locks user account" do
+      visit "/sign-in"
+      fill_in_credentials(password_override: "XXX")
+      fill_in_credentials(password_override: "XXX")
```
slorek:
```suggestion
      ENV['LOCK_MAXIMUM_ATTEMPTS'].to_i.times do
        fill_in_credentials(password_override: "XXX")
      end
```

psd-web/.env.test.example
```diff
@@ -17,3 +17,4 @@ ANTIVIRUS_USERNAME=av
 ANTIVIRUS_PASSWORD=password
 
 TWO_FACTOR_AUTHENTICATION_ENABLED=false
+LOCK_MAXIMUM_ATTEMPTS=2
```
slorek: Consider setting this inline in the relevant tests.

psd-web/app/models/user.rb
```diff
@@ -179,6 +179,17 @@ def has_viewed_introduction!
...
 
+  def send_unlock_instructions
+    raw, enc = Devise.token_generator.generate(self.class, :unlock_token)
+    self.unlock_token = enc
+    save(validate: false)
+    reset_password_token = set_reset_password_token
+    NotifyMailer.account_locked(self,
+                                unlock_token: raw,
+                                reset_password_token: reset_password_token).deliver_later
+    raw
```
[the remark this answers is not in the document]
slawosz: I will move it to `LockableOverride`, what do you think?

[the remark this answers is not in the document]
scruti: I am up for the idea (I believe @slorek and possibly others prefer to have everything together in the User model though) but I would be consistent and either extract all Devise related logic bits from the User model into concerns or leave all in the model.

psd-web/spec/support/mailer.rb
```diff
@@ -1,8 +1,29 @@
 # frozen_string_literal: true
 
+class TestNotifyEmail
```
scruti: :clap:

psd-web/app/controllers/users/sessions_controller.rb
slawosz: @slorek `self.resource` will be only available when user is successfully authorized. I need this to find user by email only.

psd-web/app/models/user.rb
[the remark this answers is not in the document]
slawosz: @slorek @frankieroberto @StupidCodeFactory what is your opinion here?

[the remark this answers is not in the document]
StupidCodeFactory: I personally think **absolutely any* of this should be in the model at all.
The database really should not know anything about sending email, generate an encrypted token or validation except it own constraints.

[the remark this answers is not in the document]
slorek: Prefer not to use modules if possible. Makes it easier to find things.

psd-web/spec/features/sign_in_spec.rb
slawosz: It works. Regarding having this being configurable in test, I don't see much value here, especially after using `Devise.maximum_attempts`.

psd-web/spec/features/sign_in_spec.rb
slawosz: I combined 2, as it is not really possible to combine reset password.

psd-web/app/models/user.rb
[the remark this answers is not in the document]
frankieroberto: Yeah I'm happy to keep it all in the model for now. That's what devise does anyhow: https://github.com/heartcombo/devise/blob/master/lib/devise/models/recoverable.rb#L89-L96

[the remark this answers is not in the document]
slawosz: @frankieroberto devise is using concerns, so similar approach. I would prefer to keep everything in concerns, but #democracy

psd-web/app/controllers/users/sessions_controller.rb
```diff
@@ -16,6 +16,11 @@ def create
...
+      if user&.access_locked?
+        return redirect_to account_locked_path
```
frankieroberto: I kinda think that we should render the account locked view directly, rather than redirect to it.  Otherwise we’ll end up with the account unlocked page (see [example in review app](https://psd-pr-412.london.cloudapps.digital/account-locked)) being public and accessible by anyone, which could be confusing if it gets indexed by google or gets into a user’s browser history?

psd-web/app/controllers/users/unlocks_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      return true if !Rails.configuration.two_factor_authentication_enabled
+
+      user_signed_in? && user_with_unlock_token == current_user && is_fully_authenticated?
+    end
+
+    def user_with_unlock_token
+      @user_with_unlock_token ||= begin
+        unlock_token = Devise.token_generator.digest(self, :unlock_token, params[:unlock_token])
+
+        User.find_by(unlock_token: unlock_token)
```
frankieroberto: Please fix: make sure this renders a 404 page if the token is invalid, [see example](https://psd-pr-412.london.cloudapps.digital/unlock?unlock_token=ZVdiu7Bswvhzo1DU5Y_6sgsdg).

```suggestion
        User.find_by!(unlock_token: unlock_token)
```

psd-web/app/models/user.rb
```diff
@@ -170,6 +170,28 @@ def need_two_factor_authentication?(_request)
     Rails.configuration.two_factor_authentication_enabled
   end
 
+  def send_unlock_instructions
+    raw, enc = Devise.token_generator.generate(self.class, :unlock_token)
```
frankieroberto: These variables aren’t super obvious. Presumably `enc` is the `unlock_token`, but what is `raw`?

psd-web/app/models/user.rb
```diff
@@ -170,6 +170,28 @@ def need_two_factor_authentication?(_request)
...
+    NotifyMailer.account_locked(self,
+                                unlock_token: raw,
+                                reset_password_token: reset_password_token).deliver_later
+    raw
+  end
+
+  # Don't reset password attempts yet, it will happen on next successful login
+  def unlock_access!
+    self.locked_at = nil
+    self.unlock_token = nil if respond_to?(:unlock_token=)
```
frankieroberto: I'm a bit unclear why `if respond_to?(:unlock_token=)` is needed here?

psd-web/app/models/user.rb
```diff
@@ -170,6 +170,28 @@ def need_two_factor_authentication?(_request)
...
+    save(validate: false)
+  end
+
+  def active_for_authentication?
+    true
+  end
```
frankieroberto: Could use a comment to explain what this does? Presume it's a method from the super class that we’re overriding? But what does it _do_?

psd-web/app/views/users/sessions/account_locked.html.erb
```diff
@@ -0,0 +1,12 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+    <h1 class="govuk-heading-l">Check your email</h1>
```
frankieroberto: Please change: set the page `<title>` too:

```suggestion
<% title = "Check your email" %>
<%= page_title(title) %>
<div class="govuk-grid-row">
  <div class="govuk-grid-column-two-thirds-from-desktop">
    <h1 class="govuk-heading-l"><%= title %></h1>
```

psd-web/config/initializers/devise.rb
```diff
@@ -293,6 +293,15 @@
   # changed. Defaults to true, so a user is signed in automatically after changing a password.
   # config.sign_in_after_change_password = true
 
+
+  config.lock_strategy = :failed_attempts
+  config.unlock_strategy = :email
+  if Rails.env.test?
+    config.maximum_attempts = 2
+  else
+    config.maximum_attempts = ENV.fetch("LOCK_MAXIMUM_ATTEMPTS", 10).to_i
```
frankieroberto: Is there any need for this to be configurable at runtime via an environment variable?  I feel like we could happily hardcode this.

psd-web/db/migrate/20200319121332_add_lockable_to_devise.rb
```diff
@@ -0,0 +1,12 @@
+class AddLockableToDevise < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured do
+      add_column :users, :failed_attempts, :integer, default: 0, null: false # Only if lock strategy is :failed_attempts
+      add_column :users, :locked_at, :datetime
+
+      # Add these only if unlock strategy is :email or :both
```
frankieroberto: Remove comments here.

psd-web/spec/features/sign_in_spec.rb
```diff
@@ -5,18 +5,22 @@
 
   let(:investigation) { create(:project) }
   let(:user) { create(:user, :activated, has_viewed_introduction: true) }
+  let(:password) { "2538fhdkvuULE36f" }
 
-  def fill_in_credentials
-    visit "/sign-in"
-
+  def fill_in_credentials(password_override: nil)
```
frankieroberto: I think ideally we should always pass in both the email and the password to this helper?

```suggestion
  def fill_in_credentials(email:, password:)
```

psd-web/spec/features/sign_in_spec.rb
```diff
@@ -56,6 +62,80 @@ def fill_in_credentials
...
+      end
+
+      Devise.maximum_attempts.times do
+        visit "/sign-in"
+        fill_in_credentials(password_override: "XXX")
+      end
+
+      expect(page).to have_css("p", text: "We’ve locked this account to protect its security.")
+
+      visit unlock_path
```
frankieroberto: Please change: use hardcoded path in feature specs rather than relying on the helpers.

psd-web/spec/features/sign_in_spec.rb
```diff
@@ -56,6 +62,80 @@ def fill_in_credentials
...
+
+      expect(page).to have_css("h1", text: "Check your phone")
+
+      fill_in "Enter security code", with: user.reload.direct_otp
+      click_on "Continue"
+
+      fill_in_credentials
+
+      expect(page).to have_css("h2", text: "Your cases")
+      expect(page).to have_link("Sign out", href: destroy_user_session_path)
```
frankieroberto: please use hardcoded path, or just `have_link("Sign out")`

psd-web/Gemfile
```diff
@@ -57,6 +57,7 @@ group :development, :test do
   gem "listen"
   gem "m"
   gem "pry"
+  gem "pry-byebug"
```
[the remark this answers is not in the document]
slawosz: Yes, `pry-byebug` is combination of pry and byebug.

psd-web/app/models/user.rb
slawosz: `raw` is the original, generated token. I want to keep them as they are since they part of device interface.

psd-web/app/models/user.rb
```diff
@@ -170,6 +170,28 @@ def need_two_factor_authentication?(_request)
     Rails.configuration.two_factor_authentication_enabled
   end
 
+  def send_unlock_instructions
+    raw, enc = Devise.token_generator.generate(self.class, :unlock_token)
+    self.unlock_token = enc
+    save(validate: false)
```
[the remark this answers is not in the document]
slawosz: I think both methods are fine. I will stay with this one, as its the same as devise.

psd-web/app/models/user.rb
slawosz: Copy paste from devise.

psd-web/config/initializers/devise.rb
slawosz: It is useful for testing locally.

psd-web/spec/features/sign_in_spec.rb
slawosz: I think we can wait until we need it.

psd-web/spec/features/sign_in_spec.rb
slawosz: Its dynamic, set in line 67.

psd-web/app/views/users/unlocks/invalid_link.html.erb
```diff
@@ -0,0 +1,12 @@
+<% title = "Invalid link" %>
+<% content_for :page_title, title %>
```
frankieroberto: Please use the `page_title(title)` helper.

psd-web/app/controllers/users/sessions_controller.rb
slawosz: Done

psd-web/spec/features/sign_in_spec.rb
```diff
@@ -56,6 +62,86 @@ def fill_in_credentials
...
+        expect(page).to have_css("h2", text: "Your cases")
+
+        visit unlock_path
+        expect(page).to have_css("h1", text: "Check your phone")
+      end
+    end
+
+    scenario "shows invalid link page for invalid link" do
+      visit "/unlock?unlock_token=wrong-token"
+      expect(page).to have_css("h1", text: "Invalid link")
```
frankieroberto: Be good to have an expectation around the response status here too, ie that it should be a `404` response.

slawosz: I am not sure if this is part of acceptance test. Do we have former requirements for this eg. due accessibility?

frankieroberto: It's probably not a huge deal, but it might affect some things, eg logging/analytics and possibly whether the browser autocompletes to it or not.

psd-web/spec/features/sign_in_spec.rb
```diff
@@ -5,18 +5,22 @@
...
-    visit "/sign-in"
-
+  def fill_in_credentials(password_override: nil)
     fill_in "Email address", with: user.email
-    fill_in "Password", with: "2538fhdkvuULE36f"
+    if password_override
+      fill_in "Password", with: password_override
+    else
+      fill_in "Password", with: password
+    end
```
scruti: Tiny suggestion (you may not like it so feel free to ignore it):
```suggestion
    fill_in "Password", with: password_override.presence || password
```

psd-web/spec/features/sign_in_spec.rb
```diff
@@ -56,6 +62,87 @@ def fill_in_credentials
...
+        visit "/sign-in"
+        fill_in_credentials
+        fill_in "Enter security code", with: user.reload.direct_otp
+        click_on "Continue"
+
+        expect(page).to have_css("h2", text: "Your cases")
+
+        visit unlock_path
+        expect(page).to have_css("h1", text: "Check your phone")
+      end
```
scruti: Have we checked with Ed this is what they expect to happen? In case they want to align in with the "complete registration" and "reset password" behaviour (asking the signed in user if it wants to proceed as the user that got the link)

psd-web/manifest.review.yml
```diff
@@ -12,6 +12,7 @@ applications:
     SIDEKIQ_QUEUE: ((sidekiq-queue))
     SENTRY_CURRENT_ENV: ((sentry-current-env))
     KEYCLOAK_REDIRECT_URI: https://((psd-host))/users/auth/openid_connect/callback
+    LOCK_MAXIMUM_ATTEMPTS: 3
```
scruti: Suggestion:
Shouldn't we use [user provided service instances](https://docs.cloudfoundry.org/devguide/services/user-provided.html) for setting this value? (unless is needed for the deployment somewhat).

slawosz: I found it much easier to do it this way. This is not needed in production.

psd-web/app/controllers/users/unlocks_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+      else
+        sign_in(user_with_unlock_token)
+        warden.session(:user)[TwoFactorAuthentication::NEED_AUTHENTICATION] = true
+        # Will redirect back to #edit after passing 2FA.
+        # fullpath contains "reset_password_token", necessary for the further update.
+        store_location_for(:user, request.fullpath)
+        user_with_unlock_token.send_new_otp
+
+        redirect_to user_two_factor_authentication_path
+      end
```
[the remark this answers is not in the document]
slawosz: I think this is `good` duplication.

psd-web/manifest.review.yml
scruti: Definitely is easier (hated the otehr way). Didn't realise is not needed in production. Happy the way it is then.

psd-web/spec/features/sign_in_spec.rb
edwardhorsford: I'm not too fussed about this situation as it should be very rare - but did raise that other developers might prefer a consistent approach.

nasirkhanbeis:
Over all functionality works fine as expected.
1. When a pre-logged in as different user tries to click on un-lock link, it will first prompt for 2fa and then prompts to sign-in.
2. When an account is brute forced locked from other devise whilst real user session is active, it doesn't force log-out but will continue session as normal and would lock when try to login after log out. user will need to click on unlock link as expected
3. Password reset works fine by link in email or explicitly as expected when an account is locked and lock is release.

Few minor niggles need to be aware of
1. Email sometimes lands in promotion folder of my google account, not sure if email client will treat as spam, in that case probably need to fix email content may be.
2. When user pre-logged in and for some reason account is locked in another device, then user upon clicking unlock link in same browser where user is logged in currently will let user re-sign in without 2fa (2fa session is retained if its same user and same browser)

psd-web/app/mailers/notify_mailer.rb
```diff
@@ -102,4 +103,16 @@ def user_added_to_team(email,
...
+
+  def account_locked(user, tokens)
+    set_template(TEMPLATES[:account_locked])
+
+    personalization = {
+      name: user.name,
+      edit_user_password_url_token: edit_user_password_url(reset_password_token: tokens[:reset_password_token]),
+      unlock_user_url_token: user_unlock_url(unlock_token: tokens[:unlock_token])
+    }
+    set_personalisation(personalization)
```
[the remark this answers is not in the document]
slorek: 👍 This would be consistent with the other methods' usage of `set_personalisation`.

psd-web/app/views/users/sessions/account_locked.html.erb
slorek: Looks like the page title is still duplicated.

PR #425, opened 2020-03-25
Fix seeds

StupidCodeFactory:
**CHANGE:**
Add mobile verified to seeds

### DIFF
--- a/psd-web/db/seeds.rb
+++ b/psd-web/db/seeds.rb
@@ -423,8 +423,22 @@ if run_seeds
   Team.create!(name: "OPSS Incident Management",  team_recipient_email: nil, "organisation": organisation)
   Team.create!(name: "OPSS Testing", team_recipient_email: nil, "organisation": organisation)
 
-  user1 = User.create!(name: "Test User", email: "user@example.com", password: "testpassword", password_confirmation: "testpassword", organisation: organisation, teams: [enforcement])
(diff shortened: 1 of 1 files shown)


psd-web/db/seeds.rb
```diff
@@ -423,8 +423,22 @@ def create_blob(filename, title: nil, description: nil)
...
   Team.create!(name: "OPSS Testing", team_recipient_email: nil, "organisation": organisation)
 
-  user1 = User.create!(name: "Test User", email: "user@example.com", password: "testpassword", password_confirmation: "testpassword", organisation: organisation, teams: [enforcement])
-  user2 = User.create!(name: "Team Admin", email: "admin@example.com", password: "testpassword", password_confirmation: "testpassword", organisation: organisation, teams: [processing])
+  user1 = User.create!(
+    name: "Test User",
+    email: "user@example.com",
+    password: "testpassword",
+    password_confirmation: "testpassword",
+    mobile_number_verified: true,
```
frankieroberto: Should we add a dummy mobile number too?

StupidCodeFactory: I thought of that too, but I was going to do it in a separate PR. 
My thinking was to add a `ENV.fetch("DEVELOPEMENT_TWO_FACTOR_AUTH_MOBILE")`, add it to the .env.example and document this in the README. so that a new developer would know it needs to adds his on mobile number.

PR #426, opened 2020-03-26
DO NOT MERGE: Disable 2FA for testing users

scruti:
## Please do not merge
PR created just for enable @nasirkhanbeis to run regression tests over `keycloak-replacement` branch without 2FA for a couple users.

This will be deleted after testing.

### DIFF
--- a/psd-web/app/models/user.rb
+++ b/psd-web/app/models/user.rb
@@ -213,7 +213,8 @@ class User < ApplicationRecord
   end
 
   def need_two_factor_authentication?(_request)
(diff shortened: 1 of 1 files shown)


psd-web/app/models/user.rb
```diff
@@ -213,7 +213,8 @@ def send_two_factor_authentication_code(code)
   end
 
   def need_two_factor_authentication?(_request)
```
slawosz: instead I would do
```
def authenticate_otp
  return true if email.in?(TEST_EMAILS)
  super
end
```

PR #427, opened 2020-03-26
Bugfix ensure user cant sign in with unverified mobile number

frankieroberto:
The existing test checked that the user gets an immediate "Enter correct email address and password" error if they try to login after having abandoned the registration flow before verifying their e-mail.

However, if the user subsequently tried to visit another URL, such as the homepage, they were then presented with the 2FA page, as `warden.authenticate(auth_options)` will have signed them in.

This fixes that, but signing the user back out again from within the controller.

### DIFF
--- a/psd-web/app/controllers/users/sessions_controller.rb
+++ b/psd-web/app/controllers/users/sessions_controller.rb
@@ -20,7 +20,11 @@ module Users
 
       # Stop users from signing in if they’ve not completed 2FA verification
       # of their mobile number during account set up process.
-      if resource && !resource.mobile_number_verified
+      if Rails.configuration.two_factor_authentication_enabled && !resource.mobile_number_verified
+        # Need to sign the user out here as they will have been signed in by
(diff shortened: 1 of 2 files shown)


psd-web/app/controllers/users/sessions_controller.rb
```diff
@@ -20,7 +20,11 @@ def create
 
       # Stop users from signing in if they’ve not completed 2FA verification
       # of their mobile number during account set up process.
-      if resource && !resource.mobile_number_verified
+      if Rails.configuration.two_factor_authentication_enabled && !resource.mobile_number_verified
+        # Need to sign the user out here as they will have been signed in by
+        # warden.authenticate(auth_options) above.
+        sign_out
+        User.current = nil
```
slorek: Suggestion: override `sign_out` to make it set `User.current = nil`.

frankieroberto: Good idea, but let's do that as a separate PR.

PR #431, opened 2020-03-26
Ask confirmation when different user open reset password link

scruti:
When multiple users share a computer and one of them request the password reset, if when opening the link to reset the password another user is logged in the site, they will be ask to either:
- Continue as the logged user, cancelling the password reset.
- Proceed to reset the password as the user who requested it.

This behaviour imitates the current flow for completing the user registration when already signed in as another user.
Screenshot from 2020-03-30 12-22-57
Screenshot from 2020-03-30 12-23-16
Screenshot from 2020-03-30 12-23-34
Screenshot from 2020-03-30 12-23-45


nasirkhanbeis:
Issue "already signed-in" flash message is not related this ticket its a regression and surfaces only when re-directing from /sign-in to home page when already sign-n. Ed raised a separate card for this.

psd-web/app/views/users/passwords/signed_in_as_another_user.html.erb
```diff
@@ -0,0 +1,18 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+    <h1 class="govuk-heading-l"><%= page_title %></h1>
+
+    <p class="govuk-body">Do you want to</p>
+
+    <ul class="govuk-list govuk-list--bullet">
+      <li><%= link_to "Continue as #{current_user.name}", root_path %></li>
+      <li><%= link_to "Reset password for #{@user_with_reset_token.name}",
```
edwardhorsford: Both these links should have the `govuk-link` class added.

psd-web/app/controllers/users/passwords_controller.rb
```diff
@@ -34,6 +35,11 @@ def edit
       end
     end
 
+    def sign_out_before_resetting_password
+      sign_out
```
frankieroberto: @scruti I think we should avoid automatically signing users out from a `GET` request, as that introduces a minor security risk, whereby someone could force a user to sign out (eg by embedding an iframe or getting them to click a link) in order to more easily obtain their password when they sign back in (via some other mechanism).

Probably not a blocker, but would be good to fix if possible, eg by doing the sign out via a form doing a POST.

psd-web/app/views/users/passwords/signed_in_as_another_user.html.erb
```diff
@@ -0,0 +1,18 @@
+<% page_title = "You are already signed in to the Product safety database" %>
+<% content_for :page_title, page_title %>
```
StupidCodeFactory: Can you change this use the `page_title` helper?

psd-web/app/views/users/passwords/signed_in_as_another_user.html.erb
```diff
@@ -0,0 +1,19 @@
...
```
frankieroberto: please use `<%= page_title(page_title) %>` helper here (mainly for consistency).

psd-web/app/views/users/passwords/signed_in_as_another_user.html.erb
```diff
@@ -0,0 +1,19 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+    <h1 class="govuk-heading-l"><%= page_title %></h1>
+
+    <p class="govuk-body">Do you want to</p>
```
frankieroberto: Should this end with a colon?

```suggestion
    <p class="govuk-body">Do you want to:</p>
```

psd-web/app/controllers/users/passwords_controller.rb
scruti: Good point. Changed it.

psd-web/app/views/users/passwords/signed_in_as_another_user.html.erb
```diff
@@ -9,9 +9,12 @@
...
     <ul class="govuk-list govuk-list--bullet">
       <li><%= link_to "Continue as #{current_user.name}", root_path, class: "govuk-link" %></li>
-      <li><%= link_to "Reset password for #{@user_with_reset_token.name}",
-              sign_out_before_resetting_password_path(reset_password_token: reset_password_token),
-              class: "govuk-link" %>
+      <li>
+        <%= form_with url: sign_out_before_resetting_password_path(reset_password_token: reset_password_token), method: :post do %>
+          <button class="button-link govuk-link">
+            <%= "Reset password for #{@user_with_reset_token.name}" %>
+          </button>
```
frankieroberto: There's a helper for this:

```suggestion
            <%= govukButton(text:"Reset password for #{@user_with_reset_token.name}") %>
```

scruti: @frankieroberto does this helper keep the button displayed as a link? Trying to keep the original list with 
- link a
- ink b

frankieroberto: Oh no, I didn't spot that it was styled as a link. 🤦‍♂

PR #433, opened 2020-03-26
Bugfix: Don’t prefill name and mobile number after abandoned registration

frankieroberto:
If you've completed the first step of registration (adding name, password and mobile number) but didn't complete the 2FA, and then you later go back and attempt to start again using the invite link, it probably makes sense not to prefill the form with your name and mobile number from last time...

See https://trello.com/c/dm7gIfWt/449-account-details-pre-filled-after-abandoning-account-registration

### DIFF
--- a/psd-web/app/controllers/users_controller.rb
+++ b/psd-web/app/controllers/users_controller.rb
@@ -7,6 +7,12 @@ class UsersController < ApplicationController
   def complete_registration
     @user = User.find(params[:id])
 
+    # Reset name and mobile number in case they’ve been remembered
+    # from a previous registration that was abandoned before the mobile number
+    # was verified via two-factor authentication.
+    @user.name = ""
+    @user.mobile_number = ""
+
     return render :signed_in_as_another_user if user_signed_in? && !signed_in_as?(@user)
 
     # Some users will bookmark the invitation URL received on the email and may re-use
--- a/psd-web/app/controllers/users_controller.rb
+++ b/psd-web/app/controllers/users_controller.rb
@@ -7,12 +7,6 @@ class UsersController < ApplicationController
   def complete_registration
     @user = User.find(params[:id])
 
-    # Reset name and mobile number in case they’ve been remembered
-    # from a previous registration that was abandoned before the mobile number
-    # was verified via two-factor authentication.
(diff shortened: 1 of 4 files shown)


psd-web/app/controllers/users_controller.rb
```diff
@@ -7,6 +7,12 @@ class UsersController < ApplicationController
   def complete_registration
     @user = User.find(params[:id])
 
+    # Reset name and mobile number in case they’ve been remembered
+    # from a previous registration that was abandoned before the mobile number
+    # was verified via two-factor authentication.
+    @user.name = ""
+    @user.mobile_number = ""
```
StupidCodeFactory: You could wrap this in a method which would hopefully be descriptive enough to avoid adding comments.
Also why an empty string over just nil?

psd-web/spec/features/creating_an_account_from_an_invitation_spec.rb
```diff
@@ -64,6 +64,47 @@
...
+             mobile_number_verified: false)
+    end
+
+    scenario "it shouldn’t show values entered previously" do
+      visit "/users/#{invited_user.id}/complete-registration?invitation=#{invited_user.invitation_token}"
+
+      expect_to_be_on_complete_registration_page
+
+      # Form should NOT contain values from previous abandoned registration
+      expect(find_field("Full name").value).to eql("")
```
StupidCodeFactory: unless I'm mistaken or this has changed since I started working on the project we use `eq` everywhere.

frankieroberto: Switched, thanks. Forgot about the subtle difference (type conversion).

psd-web/app/controllers/users_controller.rb
frankieroberto: Needs to be an empty string otherwise the rails helper doesn’t seem to output the attribute at all. Which is maybe ok, but might cause browsers to prefill using saved values?

Felt like it needed a lengthy comment here as it’s such an edge case that it’s not obvious why you’d do this.

nasirkhanbeis:
@frankieroberto I'm getting this error after clicking on login button 
image

PR #434, opened 2020-03-27
Always set the two factor authentication options

frankieroberto:
By always setting these we resolve a bug whereby, if the `TWO_FACTOR_AUTHENTICATION_ENABLED` environment variable is missing, then 2FA is turned on (as the default is true in `application.rb` but the config setting for 2fa weren’t being set in the `devise.rb` initializer.


scruti:
This seems to break the Rspec test suite. Are we sure `Rails.configuration.two_factor_authentication_enabled` is available on Devise initialisation stage?

frankieroberto:
@scruti yeah, I wasn't expecting it to break loads of tests... Looking into it now.

PR #440, opened 2020-03-27
Add interstitial page after changing password

frankieroberto:
This should help make it clearer to the user that their password was successfully updated, rather than just redirecting to the homepage.

Screenshot:

Screenshot 2020-03-27 at 16 26 53

### DIFF
--- a/psd-web/app/controllers/users/password_changed_controller.rb
+++ b/psd-web/app/controllers/users/password_changed_controller.rb
@@ -1,0 +1,8 @@
+module Users
+  class PasswordChangedController < ApplicationController
+    skip_before_action :has_accepted_declaration,
+                       :has_viewed_introduction
+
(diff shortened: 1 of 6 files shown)


psd-web/app/controllers/users/password_changed_controller.rb
```diff
@@ -0,0 +1,12 @@
+module Users
+  class PasswordChangedController < ApplicationController
```
slawosz: For this, I would add an action in `PasswordsController` especially that is only single page. `CRUD` is nice but at the same time having controller just to display static page is overkill.

frankieroberto: I started out doing that, but because that controller inherits from devise, it’s really tricky to make the view work for logged in users only.

PR #443, opened 2020-03-30
Add product categories relating to coronavirus

frankieroberto:
This adds two product categories relating to the coronavirus response.

As one of these is a rename (`Protective equipment` -> `Personal protective equipment (PPE)`), this also includes a data migration as a separate rake task to update the existing products and cases (as the categories are stored as strings).

## Screenshot

Screenshot 2020-03-30 at 16 10 45


LMallows:
Looks good until we know better how to deal with specific CORVID related categories

PR #445, opened 2020-03-31
Use RAILS_MAX_THREADS env var for db concurrency

frankieroberto:
This fixes an issue we’ve [seen in production (~4000 times)](https://sentry.io/organizations/beis/issues/1588837243/?project=1329381) whereby a Sidekiq worker process throws a `ActiveRecord::ConnectionTimeoutError (could not obtain a connection from the pool within 5 seconds)` error.

This happens because currently we’ve set the database connection pool to a fixed size of 5, but the number of Sidekiq threads to 10.

To fix this, the database connection pool size now defaults to the value of `RAILS_MAX_THREADS` if set (or 5 otherwise), and this env var is also used to set the Sidekiq concurrency and the number of Puma threads (which was already the case).

The `RAILS_MAX_THREADS` is set to 5 for Puma (web process) and 10 for Sidekiq (worker process) as currently, with these values fed through from the Github deploy scripts.

See:

* [Sidekiq advanced options: concurrency](https://github.com/mperham/sidekiq/wiki/Advanced-Options#concurrency)
* [Heroku: deploying Rails applications with Puma](https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#database-connections)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -48,6 +48,8 @@ jobs:
         WORKER_MEMORY: 1G
         WEB_INSTANCES: 1
         WORKER_INSTANCES: 1
+        WEB_MAX_THREADS: 5
+        WORKER_MAX_THREADS: 10
         CF_USERNAME: ${{ secrets.PaaSUsernameStaging }}
         CF_PASSWORD: ${{ secrets.PaaSPasswordStaging }}
       run: |
@@ -105,6 +107,8 @@ jobs:
         WORKER_MEMORY: 1G
         WEB_INSTANCES: 4
         WORKER_INSTANCES: 2
+        WEB_MAX_THREADS: 5
+        WORKER_MAX_THREADS: 10
         CF_USERNAME: ${{ secrets.PaaSUsernameProduction }}
         CF_PASSWORD: ${{ secrets.PaaSPasswordProduction }}
       run: |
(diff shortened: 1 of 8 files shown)


.github/workflows/deploy.yml
```diff
@@ -48,6 +48,8 @@ jobs:
         WORKER_MEMORY: 1G
         WEB_INSTANCES: 1
         WORKER_INSTANCES: 1
+        WEB_MAX_THREADS: 5
```
StupidCodeFactory: We could certainly bump that number a bit

.github/workflows/deploy.yml
```diff
@@ -48,6 +48,8 @@ jobs:
...
+        WORKER_MAX_THREADS: 10
```
StupidCodeFactory: Can you bind this var to the sidekiq config as well?

https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/config/sidekiq.yml#L2

slorek: I think this is done.

StupidCodeFactory: the 10 is still hard coded in sidekiq.yml ^^

slorek: https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/445/files#diff-61f87bff54cc0dce2d1209ab5f5eee5eR2

frankieroberto: Yeah the only place these values should be set now is in the GitHub workflow config. Although there are default values elsewhere that haven't changed.

.github/workflows/deploy.yml
frankieroberto: Maybe, but if so we can do that as a separate PR? Would have to look at memory etc.

.github/workflows/deploy.yml
StupidCodeFactory: 🤦‍♂ I was looking at master...

PR #449, opened 2020-03-31
Fix submitting wrong credentials twice on sign in

StupidCodeFactory:
**CHANGES:**

Upon submitting the wrong credentials twice the form is generated to use PATCH
as the user as an id

https://trello.com/c/8KwrwkSm/472-page-not-found-error-with-half-registered-user 

## Checklist:


nasirkhanbeis:
looking good as expected but in backend it is sending 2fa code on each time trying to login

nasirkhanbeis:
as @StupidCodeFactory  confirmed spamming of 2fa codes invalid user/pwd error seems a separate issue so will raise a separate ticket.

PR #450, opened 2020-03-31
Rails admin

slawosz:
Quick PoC for admin panel. Currently allows changing password and other user attributes.
Requires new role for the user, `superadmin`:
```
user = User.find(...)
user.user_roles.create name: 'superadmin'
```

To make it mergeable:
- [x] make sure it is accessible only via authentication

Screenshot from 2020-03-31 17-45-07

Screenshot from 2020-03-31 17-45-27


psd-web/config/initializers/rails_admin.rb
```diff
@@ -0,0 +1,103 @@
...
+  config.authenticate_with do
+    warden.authenticate! scope: :user
+  end
+  config.authorize_with do |_controller|
+    if warden.session(:user)[TwoFactorAuthentication::NEED_AUTHENTICATION]
+      redirect_to "/two-factor"
+    end
+    if !current_user.is_superadmin?
+      redirect_to main_app.root_path
+    end
```
scruti, line 91: Suggestion:
Could we invert the checks so we ensure the user `is_superadmin?` before sending them to 2FA page if needed?
It seems like the user will be asked for 2FA (when needed by its session) even if is not a super admin. What would be the point?

slawosz: Good point. I will change it. However it doesn't matter as the user in such case will end up in the same page.

psd-web/app/controllers/application_controller.rb
```diff
@@ -15,6 +15,13 @@ class ApplicationController < ActionController::Base
 
   helper_method :nav_items, :secondary_nav_items, :previous_search_params, :current_user
 
+
+  def current_admin
+    if current_user && current_user.is_superadmin?
+      current_user
+    end
+  end
```
scruti, line 23: Do we use this anywhere or need it at all?

psd-web/config/initializers/rails_admin.rb
```diff
@@ -0,0 +1,103 @@
...
+  config.included_models = %w[User]
+end
+
+module RailsAdminTwoFactor
+  def user_two_factor_authentication_path
+    "/two-factor"
+  end
+end
+
+RailsAdmin::MainController.include RailsAdminTwoFactor
```
scruti, line 103: A little bit confused about why are we declaring this here but at the same time we are manually redirecting to `/two-factor`  inside `config.authorize_with` block.

Little bit of context? :)

psd-web/spec/features/admin_spec.rb
```diff
@@ -0,0 +1,41 @@
+require "rails_helper"
+
+RSpec.feature "Admin Panel", :with_elasticsearch, :with_stubbed_mailer, :with_stubbed_notify, type: :feature do
+  include ActiveSupport::Testing::TimeHelpers
+
+  let(:investigation) { create(:project) }
+  let(:user) { create(:user, :activated, has_viewed_introduction: true) }
+  let(:password) { "2538fhdkvuULE36f" }
```
scruti, line 8: Please change: 
I believe this is the same as the password for user factory. Being that the case we can avoid this explicit declaration and just use `user.password` in `fill_in_credentials` as we do with `user.email`.

psd-web/spec/features/admin_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def fill_in_credentials(password_override: nil)
+    fill_in "Email address", with: user.email
+    if password_override
+      fill_in "Password", with: password_override
+    else
+      fill_in "Password", with: password
+    end
+    click_on "Continue"
+  end
```
scruti, line 18: Suggestion:
Being the same as the `sign-in` feature spec... time to extract and share the helper?
`def fill_in_credentials(user, password_override: nil)` could be a generic interface that allows us to fill any sign-in form.

psd-web/spec/features/admin_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  context "when user tries to access admin", :with_2fa do
+    it "needs to sign in" do
```
scruti, line 21: Suggestion:
Move to `feature` / `scenario` syntax. Did we reach any agreement on this? :thinking: 
Trying to normalise what syntax we use for feature tests.

psd-web/spec/features/admin_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      visit "/admin"
+      expect(page).to have_css("h1", text: "Sign in")
+      visit "/sign-in"
```
scruti, line 24: :scissors: 
We are already in the `Sign in page`. It doesn't seem we need to do a further request here, right?

psd-web/spec/features/admin_spec.rb
slawosz: But the thing is, you don't have access to `user.password`, as its write only attribute.

psd-web/spec/features/admin_spec.rb
scruti, line 21:
```suggestion
    it "needs to be signed in as super admin" do
```

psd-web/spec/features/admin_spec.rb
scruti: We do in the factory generated user: 
```
 user.password
=> "2538fhdkvuULE36f"
```

psd-web/config/initializers/rails_admin.rb
scruti: Yeah is more about the user experience turning weird :D

psd-web/spec/features/admin_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      fill_in_credentials
```
scruti, line 25: Suggestion:
add expectation about not being on the admin panel after submitting credentials.

psd-web/config/initializers/rails_admin.rb
```diff
@@ -0,0 +1,106 @@
...
+    end
+  end
+
+  config.included_models = %w[User]
+end
+
+# Due two_factor_authentication_gem implementation, which is being hooked to
+# RailsAdmin controller, we need to define this path.
+module RailsAdminTwoFactor
+  # This method will be called inside RailsAdmin, but is only defined for host app.
```
scruti, line 100: Thanks for clarifying this!

psd-web/spec/features/admin_spec.rb
```diff
@@ -0,0 +1,31 @@
+require "rails_helper"
+
+RSpec.feature "Admin Panel", :with_elasticsearch, :with_stubbed_mailer, :with_stubbed_notify, type: :feature do
+  include ActiveSupport::Testing::TimeHelpers
```
scruti, line 4: :scissors: I don't think we are using any of the helpers right?

nasirkhanbeis:
Re-tested this in review app with permission as super user, I can manage user by being able to change phone  number and name, all good. 
At moment review app is on new db so you  will see you only fewer users.

PR #452, opened 2020-04-01
Remove keycloak client and connector

frankieroberto:
Removes the remaining bits of keycloak code from the app, including the gem.


psd-web/test/controllers/teams_controller_test.rb
```diff
@@ -3,7 +3,6 @@
 class TeamsControllerTest < ActionDispatch::IntegrationTest
   setup do
     stub_notify_mailer
-    stub_user_management
```
scruti, line 6: Please change: 
We need to remove the `stub_user_management` definition in lines [91-95](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/452/files#diff-1a34cb4d3c24eb19b90e28d3b622b709R91-R95)

frankieroberto: Thanks. Think these got added back in when I "fixed" the merge conflict.

PR #453, opened 2020-04-01
Resend invitation bug fix

StupidCodeFactory:
**CHANGES:**

Regenerates invitation token upon resending an invitation


psd-web/app/models/user.rb
```diff
@@ -59,9 +63,9 @@ def self.create_and_send_invite!(email_address, team, inviting_user)
 
   def self.resend_invite(email_address, inviting_user)
     # Only want to allow resending invites to users that share a team with the inviting user.
-    user = User.joins(:teams)
-               .where(teams: { id: inviting_user.teams.pluck(:id) })
-               .find_by!(email: email_address)
+    user = find_user_in_teams_with!(email_address, inviting_user.teams)
+
+    user.update!(invitation_token: SecureRandom.hex(15))
```
frankieroberto, line 68: Please change: we only want to generate the token if there isn’t one already. That way, any existing email invites will still contain a valid link. This helps in the scenario where someone invites a colleague, but the email doesn’t arrive immediately, so they request a re-invite, and then the first email arrives.

psd-web/app/models/user.rb
```diff
@@ -38,6 +38,10 @@ def self.activated
     where(account_activated: true)
   end
 
+  def self.find_user_in_teams_with!(email, teams)
```
frankieroberto, line 41: Suggestion: rename and use keyword arguments:

```suggestion
  def self.find_user_within_teams_with_email!(teams:, email:)
```

psd-web/spec/models/user_spec.rb
```diff
@@ -179,12 +179,33 @@
     before { allow(SendUserInvitationJob).to receive(:perform_later) }
 
     context "when both users belong to same team" do
-      let(:invited_user) { create(:user, teams: inviting_user.teams, organisation: inviting_user.organisation) }
+      let!(:inviation_token) { SecureRandom.hex }
```
slorek, line 182:
```suggestion
      let!(:invitation_token) { SecureRandom.hex }
```

PR #455, opened 2020-04-02
Add homepage banner

frankieroberto:
This adds a homepage banner, to notify users about the coronavirus priority.

See https://trello.com/c/93g1J8ot/479-1-add-banner-to-homepage

### Screenshot

Screenshot 2020-04-02 at 15 08 19

#### Focused style

Screenshot 2020-04-02 at 15 38 06

## Implementation notes

This is implemented as a custom "banner" component, so that it can be reused elsewhere, in anticipation of an upcoming story which uses a similar banner.

## Checklist:

### General testing (author)
- [x] Test without JavaScript
- [x] Test on small screen

- [x] Works keyboard only


psd-web/spec/helpers/banner_helper_spec.rb
```diff
@@ -0,0 +1,34 @@
+require "rails_helper"
+
+RSpec.describe BannerHelper do
+  describe "#banner" do
+    context "with a text param" do
+      subject(:banner) { helper.banner(text: "There is a problem") }
```
StupidCodeFactory, line 6: if not using  `is_expected` rspec helper there is no point using `subject` over `let`. I'd suggest to `s/subject/let/`

psd-web/app/views/homepage/_coronavirus_banner.html.erb
```diff
@@ -0,0 +1,15 @@
...
+
+      <%= tag.p do %>
+        <%= link_to("https://www.gov.uk/guidance/opss-coronavirus-covid-19-guidance-for-business-and-local-authorities") do %>
+          Read more
+          <%= tag.span("about the coronavirus guidance for local authorities", class: "govuk-visually-hidden") %>
+          on GOV.UK
+        <% end %>
+      <% end %>
+    <% end %>
+  <% end %>
```
StupidCodeFactory, line 14: Why not just use regular HTML?  Wasn't the whole point of moving away from `less` to be more designer friendly and produce templates that are basically only HTML sprinkled with a tad of `erb` tags. 
There is no benefit in doing this there's nothing dynamic except the link.

psd-web/app/helpers/banner_helper.rb
```diff
@@ -0,0 +1,23 @@
+module BannerHelper
+  # Use the banner component to display an important site-wise message.
+  #
+  # Example:
+  #
+  # <%= banner(text: "We are experiencing temporary performance issues") %>
+  #
+  # The component can contain HTML, given either as a `html` param or a block:
```
StupidCodeFactory, line 8: I think the documentation is superfluous. But if you want to keep it it's not accurately describing what it does.
1. if there is a block passed it takes precedence and  and will never render either `html` nor `text`
2. you should specify that if you pass both an HTML or element it will not do anything with the text.

Overall I think the API of this helper is weird, I appreciate the effort to abstract some of this. But it feels like this should be re-worked and improved, may checking combinations of arguments and issuing appropriate warnings.
or completely remove the documentation and let the developer work out how this works.

psd-web/spec/helpers/banner_helper_spec.rb
scruti: I still prefer to keep using named subjects as it highlights what is the subject under test.
Is like `describe` and context`. They are effectively the same but they are still used for different purposes.

psd-web/app/views/homepage/_coronavirus_banner.html.erb
frankieroberto: Good point. Some of this was originally within the helper, but I forgot to convert it back to regular HTML when moving it into the partial. Will fix.

psd-web/spec/helpers/banner_helper_spec.rb
frankieroberto: I've switched them from `subject` to `let`, but kept them outside of the `it` blocks as I think that improves readability?

psd-web/app/helpers/banner_helper.rb
frankieroberto: I've (hopefully) improved the documentation a bit, to make clearer that the block takes precedence over the `html` argument, which takes precedence over the `text` argument.

Supporting both `html` and `text` arguments is to follow the convention of the Design System components. Supporting `&block` follows the convention of Rails helpers (eg [`link_to`](https://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to)) – also similar to the `call` feature in Nunjucks macros, which some of the Design System components support.

We _could_ raise an `ArgumentError` if more than one (or none) of `text`, `html` or `&block` are specified? Doesn’t feel like it’d happen that often though, and that kind of defensive programming feels a bit un-Ruby like perhaps?

LMallows:
Looks good :)

PR #456, opened 2020-04-02
Add coronavirus flag to investigations table

slorek:
https://trello.com/c/b3K7oCvN/466-add-page-asking-if-case-is-related-to-coronavirus-for-trading-standards

## Description
Adds the column required to record whether an investigation is related to the 2020 coronavirus pandemic. This is required for several other stories so I have separated it from the above linked story to unblock work on those.

Also contains schema updates missing from https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/444

## Checklist:

### DIFF
--- a/psd-web/db/schema.rb
+++ b/psd-web/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_03_20_144542) do
+ActiveRecord::Schema.define(version: 2020_04_02_142659) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
@@ -161,6 +161,7 @@ ActiveRecord::Schema.define(version: 2020_03_20_144542) do
     t.text "hazard_description"
     t.string "hazard_type"
     t.boolean "is_closed", default: false
+    t.boolean "coronavirus_related", default: false
     t.boolean "is_private", default: false, null: false
     t.text "non_compliant_reason"
     t.string "pretty_id"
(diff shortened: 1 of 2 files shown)


psd-web/db/migrate/20200402142659_add_is_coronavirus_related_to_investigations.rb
```diff
@@ -0,0 +1,7 @@
+class AddIsCoronavirusRelatedToInvestigations < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured do
+      add_column :investigations, :is_coronavirus_related, :boolean, default: false
```
StupidCodeFactory, line 4: NITPICK:  I do like following the ruby guidelines on method naming https://rubystyle.guide/#bool-methods-prefix
```suggestion
      add_column :investigations, :coronavirus_related, :boolean, default: false
```

psd-web/db/schema.rb
```diff
@@ -161,6 +161,7 @@
     t.text "hazard_description"
     t.string "hazard_type"
     t.boolean "is_closed", default: false
+    t.boolean "is_coronavirus_related", default: false
```
frankieroberto, line 164: Please change: This no longer matches the migration... 😄

psd-web/db/migrate/20200402142659_add_is_coronavirus_related_to_investigations.rb
```diff
@@ -0,0 +1,7 @@
+class AddIsCoronavirusRelatedToInvestigations < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured do
+      add_column :investigations, :coronavirus_related, :boolean, default: false
+    end
```
frankieroberto, line 5: We _could_ avoid locking the whole table here (to add the default value), using one of the strategies [mentioned here](https://github.com/ankane/strong_migrations#adding-a-column-with-a-default-value) but it’s probably overkill at our scale...

slorek: Interestingly this raises a `Rails/ReversibleMigration` rubocop failure so I think I'll keep it as was.

frankieroberto: 👍

PR #458, opened 2020-04-02
 Bugfix: Avoid sending 2FA code when not requesting 2FA

scruti:
Trello Card: https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/458

# Short? description

We don't allow to login into the service users who didn't complete the 2FA validation for their mobile phone when completing their registration.
But when they attempt to use their credentials, on top of the login error page, they get a 2FA code in their mobile phone.

This happens because the code send 2FA codes is hooked to warden authentication method. And this method is being called by Devise user helpers that are being invoked by our service in `before_action` `ApplicationController` declarations.

The solution consists in delaying these calls until we effectively want to authorise the user.

# Detailed description

We only want to send sms with 2FA code for users that are going to be
redirected to 2FA page in order to fil the form with the received code.

Our sign in business logic dictates some cases where we don't want
the sign in submission to reach 2FA, like when the user didn't verified
its phone during the registration from an invitation email link.

The ruby gem used for 2FA adds a hook to successful Warden
authentication calls, so whenever a Warden authentication happens, a
2FA code gets send to the user (if the user is marked as requiring 2FA).

Devise helpers like "current_user" and "user_signed_in?" internally
do a "warden.authenticate", triggering the 2FA code being sent.

Our application controller has multiple before filters calling those
Devise methods. So, when submitting a Sign in form, the filters where
being triggered and the 2FA code was being sent even before we could
run our business logic to dictate if an user gets directed to 2FA page
or not.

This was causing users not being directed to 2FA page and being shown
a login error (as the business logic dictates) but still getting an
SMS with a 2FA code for each login submission.

The solution consists in suppressing those calls to Devise methods
based in existing user until the point where we actually want to sign
in an user. In that point we set up those variables/invoke those calls.

This way the 2FA code will be sent only when we explicitly authenticate
the user, instead of being an uncontrolled side effect.

## Refactor session creation method

The method in charge of handling user sign-in submissions has become
complex with multiple branching for dealing with edge cases and
different ways to redirect after specific failures.

As a result of all this mixed logic, I found the method hard to follow and
change.

As an attempt to improve its maintainability:

- Re-shaped it so the different edge cases and errors provoke an early
  return one by one, while defaulting at the end to the happy path with
  a successful sign in. This is done to allow the reader to follow a
  linear process of the checks that would cause a sign-in failure.

- Documented all these checks in the method declaration, so future
  maintainers can understand better how our Sign-in process work.

- Extracted a couple private methods encapsulating checks/logic to
  show the intention of the code bit through the naming of the method.

- Added comments where considered necessary.

### DIFF
--- a/psd-web/app/controllers/users/sessions_controller.rb
+++ b/psd-web/app/controllers/users/sessions_controller.rb
@@ -3,7 +3,7 @@ module Users
     skip_before_action :has_accepted_declaration
     skip_before_action :has_viewed_introduction
     # These methods trigger Warden authentication.
-    # We don't want this to happen until we explicitely attempt to authenticate the user.
+    # We don't want this to happen until we explicitly attempt to authenticate the user.
     skip_before_action :set_current_user, :set_raven_context, :authorize_user, only: :create
 
     def new
--- a/psd-web/app/controllers/users/sessions_controller.rb
+++ b/psd-web/app/controllers/users/sessions_controller.rb
@@ -2,6 +2,8 @@ module Users
   class SessionsController < Devise::SessionsController
     skip_before_action :has_accepted_declaration
     skip_before_action :has_viewed_introduction
+    # These methods trigger Warden authentication.
+    # We don't want this to happen until we explicitely attempt to authenticate the user.
     skip_before_action :set_current_user, :set_raven_context, :authorize_user, only: :create
 
     def new
--- a/psd-web/app/controllers/users/sessions_controller.rb
+++ b/psd-web/app/controllers/users/sessions_controller.rb
@@ -10,36 +10,18 @@ module Users
       super { self.resource = resource.decorate }
     end
 
-    # Submission of the sign-in form.
-    #
-    # This method follows a sequence of steps checking possible errors and edge
-    # cases that would impede the user from being signed in.
-    # The checks are, in listed order:
-    # 1. Are the sumbitted values invalid?
-    # 2. Do the credentials correspond to an user that didn't verify its mobile
-    #    number through 2FA when completing its user registration?
-    # 3. In case of failing authentication. Did the user acount become locked?
-    # 4. Were the credentials wrong in the authentication attempt?
-    # 5. On successful authentication. Is the user missing its mobile number?
-    #
-    # When the sign-in submission does not fall under any of these checks,
-    # the user will be successfully set and signed in.
     def create
       set_resource_as_new_user_from_params
 
-      # Checks against form attributes validations
       if sign_in_form.invalid?
-        resource.errors.merge!(sign_in_form.errors)
+        handle_invalid_form(resource)
         return render :new
       end
 
       matching_user = User.find_by(email: sign_in_form.email)
 
-      # Stop users from signing in if they’ve not completed 2FA verification
-      # of their mobile number during account set up process.
-      if user_missing_2fa_mobile_verification?(matching_user)
-        sign_out
-        add_wrong_credentials_errors(resource)
+      if missing_mobile_verification?(matching_user)
+        handle_missing_mobile_verification(resource)
         return render :new
       end
 
@@ -72,6 +54,15 @@ module Users
       return render :new
     end
 
+    def handle_invalid_form(resource)
+      resource.errors.merge!(sign_in_form.errors)
+    end
+
+    def handle_missing_mobile_verification(resource)
+      sign_out
+      add_wrong_credentials_errors(resource)
+    end
+
     def sign_in_form
       @sign_in_form ||= SignInForm.new(sign_in_params)
     end
@@ -83,7 +74,7 @@ module Users
       resource.errors.add(:password, nil)
     end
 
-    def user_missing_2fa_mobile_verification?(user)
+    def missing_mobile_verification?(user)
       Rails.configuration.two_factor_authentication_enabled && user && !user.mobile_number_verified
     end
 
--- a/psd-web/app/controllers/users/sessions_controller.rb
+++ b/psd-web/app/controllers/users/sessions_controller.rb
@@ -2,6 +2,7 @@ module Users
   class SessionsController < Devise::SessionsController
     skip_before_action :has_accepted_declaration
     skip_before_action :has_viewed_introduction
+    skip_before_action :set_current_user, :set_raven_context, :authorize_user, only: :create
 
     def new
       super { self.resource = resource.decorate }
@@ -17,13 +18,10 @@ module Users
       end
 
       user = User.find_by(email: sign_in_form.email)
-      return render "account_locked" if user&.access_locked?
-
-      self.resource = warden.authenticate(auth_options)
 
       # Stop users from signing in if they’ve not completed 2FA verification
       # of their mobile number during account set up process.
-      if Rails.configuration.two_factor_authentication_enabled && resource && !resource.mobile_number_verified
+      if Rails.configuration.two_factor_authentication_enabled && user && !user.mobile_number_verified
         # Need to sign the user out here as they will have been signed in by
         # warden.authenticate(auth_options) above.
         sign_out
@@ -32,7 +30,13 @@ module Users
         return render :new
       end
 
+      self.resource = warden.authenticate(auth_options)
+      return render "account_locked" if user&.reload&.access_locked?
+
       if resource&.mobile_number?
+        set_current_user
+        set_raven_context
+        authorize_user
         sign_in(resource_name, resource)
         return respond_with resource, location: after_sign_in_path_for(resource)
       elsif resource
--- a/psd-web/app/controllers/users/sessions_controller.rb
+++ b/psd-web/app/controllers/users/sessions_controller.rb
@@ -43,21 +43,18 @@ module Users
         return render :new
       end
 
-      # Attempts to authenticate the user with its credentials.
-      # The resource will be null if the authentication fails.
       self.resource = warden.authenticate(auth_options)
 
-      # On authentication failure
-      if !resource
-        # User may have become locked
-        return render "account_locked" if matching_user&.reload&.access_locked?
-
-        set_resource_as_new_user_from_params
-        add_wrong_credentials_errors
-        return render :new
+      if resource
+        handle_authentication_success
+      else
+        handle_authentication_failure(matching_user)
       end
+    end
 
-      # On successful authentication
+  private
+
+    def handle_authentication_success
       return redirect_to missing_mobile_number_path if !resource.mobile_number?
 
       set_current_user
@@ -67,7 +64,14 @@ module Users
       respond_with resource, location: after_sign_in_path_for(resource)
     end
 
-  private
+    def handle_authentication_failure(user)
+      return render "account_locked" if user&.reload&.access_locked?
+
+      set_resource_as_new_user_from_params
+      add_wrong_credentials_errors
+      return render :new
+    end
+
 
     def sign_in_form
       @sign_in_form ||= SignInForm.new(sign_in_params)
--- a/psd-web/app/controllers/users/sessions_controller.rb
+++ b/psd-web/app/controllers/users/sessions_controller.rb
@@ -8,44 +8,61 @@ module Users
       super { self.resource = resource.decorate }
     end
 
+    # Submission of the sign-in form.
+    #
+    # This method follows a sequence of steps checking possible errors and edge
+    # cases that would impede the user from being signed in.
+    # The checks are, in listed order:
+    # 1. Are the sumbitted values invalid?
+    # 2. Do the credentials correspond to an user that didn't verify its mobile
+    #    number through 2FA when completing its user registration?
+    # 3. In case of failing authentication. Did the user acount become locked?
+    # 4. Were the credentials wrong in the authentication attempt?
+    # 5. On successful authentication. Is the user missing its mobile number?
+    #
+    # When the sign-in submission does not fall under any of these checks,
+    # the user will be successfully set and signed in.
     def create
-      self.resource = resource_class.new(sign_in_params)
+      set_dummy_devise_resource_for_display_errors
 
+      # Checks against form attributes validations
       if sign_in_form.invalid?
         resource.errors.merge!(sign_in_form.errors)
-
         return render :new
       end
 
-      user = User.find_by(email: sign_in_form.email)
+      matching_user = User.find_by(email: sign_in_form.email)
 
       # Stop users from signing in if they’ve not completed 2FA verification
       # of their mobile number during account set up process.
-      if Rails.configuration.two_factor_authentication_enabled && user && !user.mobile_number_verified
-        # Need to sign the user out here as they will have been signed in by
-        # warden.authenticate(auth_options) above.
+      if user_missing_2fa_mobile_verification?(matching_user)
         sign_out
-        User.current = nil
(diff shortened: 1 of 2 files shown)


psd-web/spec/features/sign_in_spec.rb
```diff
@@ -21,13 +20,14 @@ def expect_incorrect_email_or_password
...
+    expect(page).to have_css("span#password-error", text: "")
 
     expect(page).not_to have_link("Cases")
   end
 
-  context "when succeeeding signin in", :with_2fa do
+  context "when succeeding signin in", :with_2fa do
     context "when in two factor authentication page" do
-      it "allows user to sign in with correct two factor authentication code" do
+      scenario "allows user to sign in with correct two factor authentication code" do
```
scruti, line 30: Found a mixture of `it` and `scenario` declarations through all this file. 
I defaulted everything to `scenario` as they're feature specs and wanted to keep a cohesive approach. They're not following the `feature/scenario` approach but is closer than before? Guess this file still needs some improvements.

psd-web/spec/features/sign_in_spec.rb
```diff
@@ -166,64 +166,46 @@ def expect_incorrect_email_or_password
...
+    scenario "doesn’t let them sign in and does not send them a 2FA code by sms" do
       visit "/sign-in"
 
-      fill_in "Email address", with: user.email
-      fill_in "Password", with: "2538fhdkvuULE36f"
-      click_on "Continue"
-
-      expect_incorrect_email_or_password
-
-      # tries again with another wrong password
```
scruti, line 178: Actually we were using the right password! (Copied from the factory).

psd-web/spec/features/sign_in_spec.rb
```diff
@@ -166,64 +166,46 @@ def expect_incorrect_email_or_password
...
-      fill_in "Email address", with: user.email
-      fill_in "Password", with: "2538fhdkvuULE36f"
-      click_on "Continue"
-
+      fill_in_credentials
       expect_incorrect_email_or_password
+      expect(notify_stub).not_to have_received(:send_sms)
```
scruti, line 174: Added expectation over the sms as is part of the user interface (a quite invasive one to be honest).

psd-web/spec/features/sign_in_spec.rb
```diff
@@ -5,14 +5,13 @@
 
   let(:investigation) { create(:project) }
   let(:user) { create(:user, :activated, has_viewed_introduction: true) }
-  let(:password) { "2538fhdkvuULE36f" }
```
scruti, line 8: removed this as is just the same value as the one hard-coded in the user factory.

psd-web/app/controllers/users/sessions_controller.rb
```diff
@@ -2,46 +2,67 @@ module Users
...
-      elsif resource
-        return redirect_to missing_mobile_number_path
+      # Attempts to authenticate the user with its credentials.
+      # The resource will be null if the authentication fails.
+      self.resource = warden.authenticate(auth_options)
+
+      # On authentication failure
+      if !resource
+        # User may have become locked
+        return render "account_locked" if matching_user&.reload&.access_locked?
```
scruti, line 51: Made a conscious decision of adding this second conditional level as I found easier to read the sign-in process as `when the authentication failed, the user may have become locked`.

I tried to highlight a clear point where the credential authentication is attempted. Then, based on the authentication passing or failing, we may attempt further specific checks.

scruti:
[On invalid user/pwd error- it is spamming with 2fa codes on each time when it throws error](https://trello.com/c/WLFcz91P/475-on-invalid-user-pwd-error-it-is-spamming-with-2fa-codes-on-each-time-when-it-throws-error)

psd-web/spec/features/sign_in_spec.rb
frankieroberto: I like using `scenario` in feature specs, but if you rename them I think it’d be good to rename the description slightly too, so that it reads as a scenario, eg:

```suggestion
      scenario "user signing in with correct two factor authentication code" do
```

StupidCodeFactory:
@scruti is there a reason why you did use devise's `#bypass_sign_on` https://www.rubydoc.info/github/plataformatec/devise/Devise%2FControllers%2FSignInOut:bypass_sign_in ?

psd-web/app/controllers/users/sessions_controller.rb
```diff
@@ -2,46 +2,69 @@ module Users
...
+
+        set_dummy_devise_resource_for_display_errors
```
StupidCodeFactory, line 55: method name is not great, it's not a dummy user, how about something along the lines of `set_new_resource_with_errors`

psd-web/app/controllers/users/sessions_controller.rb
```diff
@@ -2,46 +2,69 @@ module Users
...
+        add_wrong_credentials_errors
+        return render :new
       end
 
-      self.resource = resource_class.new(sign_in_params).decorate
-      add_wrong_credentials_errors
```
StupidCodeFactory, line 43: This method is far from ideal, there is an implicite dependency on the fact that th `resource` must be set beforehand.
It is IMHO better to the object to be manipulated as and argument as such

```add_wrong_credentials_errors(resource)```

File: psd-web/app/controllers/users/sessions_controller.rb [hunk 8]
```diff
@@ -2,46 +2,69 @@ module Users
...
-      elsif resource
-        return redirect_to missing_mobile_number_path
+      # Attempts to authenticate the user with its credentials.
+      # The resource will be null if the authentication fails.
+      self.resource = warden.authenticate(auth_options)
+
+      # On authentication failure
+      if !resource
+        # User may have become locked
+        return render "account_locked" if matching_user&.reload&.access_locked?
```
StupidCodeFactory, line 53: can we wrap this into a method `handle_account_locked` ?
return `false` in `handle_account_locked` so that it does not return
 
```handle_account_locked && return```

File: psd-web/app/controllers/users/sessions_controller.rb [hunk 9]
```diff
@@ -2,46 +2,69 @@ module Users
...
+        set_dummy_devise_resource_for_display_errors
+        add_wrong_credentials_errors
+        return render :new
       end
 
-      self.resource = resource_class.new(sign_in_params).decorate
-      add_wrong_credentials_errors
-      render :new
+      # On successful authentication
+      return redirect_to missing_mobile_number_path if !resource.mobile_number?
```
StupidCodeFactory, line 61: same here `handle_missing_mobile_number && return`

psd-web/app/controllers/users/sessions_controller.rb
```diff
@@ -2,46 +2,69 @@ module Users
...
         return render :new
       end
 
-      if resource&.mobile_number?
-        sign_in(resource_name, resource)
-        return respond_with resource, location: after_sign_in_path_for(resource)
-      elsif resource
-        return redirect_to missing_mobile_number_path
+      # Attempts to authenticate the user with its credentials.
+      # The resource will be null if the authentication fails.
```
StupidCodeFactory, line 47: I don't think this comment is necessary, it more noisy than anything. Give the next line is an `if !resource` this is a clear indication this method can either return something or not.
There is already a massive amount of comments in this overly bloated class.

psd-web/app/controllers/users/sessions_controller.rb
```diff
@@ -54,5 +77,13 @@ def add_wrong_credentials_errors
       resource.errors.add(:email, I18n.t(:wrong_email_or_password, scope: "sign_user_in.email"))
       resource.errors.add(:password, nil)
     end
+
+    def user_missing_2fa_mobile_verification?(user)
+      Rails.configuration.two_factor_authentication_enabled && user && !user.mobile_number_verified
+    end
+
+    def set_dummy_devise_resource_for_display_errors
```
frankieroberto, line 85: Naming suggestion (I think "dummy" is a bit colloquial?):

```suggestion
    def set_resource_as_new_user_from_params
```

psd-web/app/controllers/users/sessions_controller.rb
```diff
@@ -2,46 +2,69 @@ module Users
...
+    #
+    # This method follows a sequence of steps checking possible errors and edge
+    # cases that would impede the user from being signed in.
+    # The checks are, in listed order:
+    # 1. Are the sumbitted values invalid?
+    # 2. Do the credentials correspond to an user that didn't verify its mobile
+    #    number through 2FA when completing its user registration?
+    # 3. In case of failing authentication. Did the user acount become locked?
+    # 4. Were the credentials wrong in the authentication attempt?
+    # 5. On successful authentication. Is the user missing its mobile number?
```
StupidCodeFactory, line 23: Opinion: Had we kept the use of interactors and an organiser (or at least a few service classes) this 5 points would have easily been translated to expressive, testable, reusable and maintainable code. I think we can all agree this controller has gotten wildly out of hand.
Only sticking with models, views and controller with this amount of complexity is an absolute challenge, I probably would not have done better myself, don't get me wrong!

This is a clear example that sticking to only MVC gets out of hand very very quickly should add a bit more business logic.

frankieroberto, line 23: Helpful comments, thanks! I wonder if it’d be even better to move these numbered steps within the comments so that they appear inline next to the code which does each bit?

psd-web/spec/features/sign_in_spec.rb
```diff
@@ -21,13 +20,14 @@ def expect_incorrect_email_or_password
     expect(page).to have_css("h2#error-summary-title", text: "There is a problem")
     expect(page).to have_link("Enter correct email address and password", href: "#email")
     expect(page).to have_css("span#email-error", text: "Error: Enter correct email address and password")
+    expect(page).to have_css("span#password-error", text: "")
 
     expect(page).not_to have_link("Cases")
   end
 
-  context "when succeeeding signin in", :with_2fa do
+  context "when succeeding signin in", :with_2fa do
```
frankieroberto, line 28:
```suggestion
  context "when successfully signing in", :with_2fa do
```

psd-web/spec/features/sign_in_spec.rb
```diff
@@ -166,64 +166,46 @@ def expect_incorrect_email_or_password
...
-
+      fill_in_credentials
       expect_incorrect_email_or_password
+      expect(notify_stub).not_to have_received(:send_sms)
     end
   end
 
   context "when email address does not belong to any user", :with_2fa do
-    it "shows a generic error message" do
+    scenario "shows a generic error message" do
```
frankieroberto, line 179: Remove this line and unnest it within the previous `context` block (renamed as a `scenario`)?

psd-web/app/controllers/users/sessions_controller.rb
```diff
@@ -2,46 +2,69 @@ module Users
   class SessionsController < Devise::SessionsController
     skip_before_action :has_accepted_declaration
     skip_before_action :has_viewed_introduction
+    # These methods trigger Warden authentication.
+    # We don't want this to happen until we explicitely attempt to authenticate the user.
```
frankieroberto, line 6: Typo:

```suggestion
    # We don't want this to happen until we explicitly attempt to authenticate the user.
```

psd-web/app/controllers/users/sessions_controller.rb
StupidCodeFactory: that's what I'm trying to do in the later comments.

psd-web/spec/features/sign_in_spec.rb
scruti: Implemented it through all the file.

psd-web/app/controllers/users/sessions_controller.rb
scruti: Sure! I didn't think much about the name :sweat_smile:

psd-web/app/controllers/users/sessions_controller.rb
scruti: I agree. The reason why I re-shaped the method and heavily commented it is precisely how difficult I found to understand the behaviour and align it with the actual business cases we wanted to cover.

Plus dealing with the early session triggering 2FA codes was an absolute nightmare given how brittle the method ended being with so much logic together.

psd-web/app/controllers/users/sessions_controller.rb
scruti: I was experimenting with something similar to this approach yesterday. What didn't convince me is that when writing it like 
```ruby
def method_name
  handle_account_locked(user) && return render "account locked"

  fallback_behaviour
end
```
I don't find it easy to follow it as a conditional. As "we handle this" it looks as some sort of affirmative statement.
I like to do it after the check but not including the condition itself? 
Something like
```ruby
return handle_account_locked if user.locked?
```

psd-web/app/controllers/users/sessions_controller.rb
```diff
@@ -2,46 +2,69 @@ module Users
...
       end
 
-      user = User.find_by(email: sign_in_form.email)
-      return render "account_locked" if user&.access_locked?
-
-      self.resource = warden.authenticate(auth_options)
+      matching_user = User.find_by(email: sign_in_form.email)
 
       # Stop users from signing in if they’ve not completed 2FA verification
       # of their mobile number during account set up process.
```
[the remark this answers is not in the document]
scruti: Done!

psd-web/app/controllers/users/sessions_controller.rb
```diff
@@ -2,57 +2,84 @@ module Users
...
 
-      # Stop users from signing in if they’ve not completed 2FA verification
-      # of their mobile number during account set up process.
-      if Rails.configuration.two_factor_authentication_enabled && resource && !resource.mobile_number_verified
-        # Need to sign the user out here as they will have been signed in by
-        # warden.authenticate(auth_options) above.
-        sign_out
-        User.current = nil
-        add_wrong_credentials_errors
+      if mobile_not_verified?(matching_user)
```
StupidCodeFactory, line 23: you could inline this:
```ruby
 return render :new if mobile_not_verified?(matching_user) && handle_mobile_not_verified(resource)
```

psd-web/app/controllers/users/sessions_controller.rb
scruti: Once I extracted methods I think it became easier enough to understand through the code. So ended removing this comment.

psd-web/app/controllers/users/sessions_controller.rb
scruti: Not really convinced of adding the handler to the condition :/
I tried the following code, but found it a little bit too long for a one liner and decided to set is back as a block.
```ruby
handle_mobile_not_verified(resource) and return render :new if mobile_not_verified?(matching_user)
```

In case we're really want a one liner I would go for `do_x and do_y if z?` rather than `do_y if z? && do_x`

scruti:
@StupidCodeFactory on the `bypass_sign_in`.
It would be useful if we had the unwanted side effect in the `sign_in` call. But at the point of the sign in is precisely when we actually want the warden hooks to be triggered.

Sadly, the unwanted behaviour is happening when `current_user` is being set by the `ApplicationController` before executing the  `SessionsController#create`.

PR #460, opened 2020-04-03
Add badge for coronavirus-related cases

frankieroberto:
This adds a badge on cases that are coronavirus (COVID-19) related, on both list views and on the case page itself.

See https://trello.com/c/RmjgZzsC/478-1-highlight-covid-19-cases

### List view

Screenshot 2020-04-03 at 16 21 40

### Case page view

Screenshot 2020-04-03 at 16 24 25

## Implementation notes

I've resisted refactoring this too much, except for moving the text of the badges into the locale file to avoid duplication, and converting a Slim partial to ERb.

## Checklist:

### General testing (author)
- [x] Test on small screen


edwardhorsford:
I've pushed some changes to:

* Move the coronavirus colour to a `_colours.scss` file.
* Invert the colours for this special case.

slorek:
Don't merge this until the cases on production which should be updated to `true` have been updated.

edwardhorsford:
Should get a dev to review as I've also worked on this.

psd-web/config/locales/en.yml
```diff
@@ -31,6 +31,10 @@
 
 en:
   enquiries_email: "opss.enquiries@beis.gov.uk"
+  case_badges:
```
slorek, line 34: Nitpick suggestion: Use same namespace as https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/462/files#diff-07308eeb711dc6bb0c2fa4d380cfb013R34 i.e. `case.badges` otherwise we will have two `case` and `case_badges` namespaces.

psd-web/app/views/investigations/_pages_top.html.slim
```diff
@@ -2,13 +2,17 @@
   .govuk-grid-row
     .govuk-grid-column-one-half
       = render "breadcrumbs_navigation", breadcrumbs: @breadcrumbs
-    - if @investigation.is_private? || @investigation.is_closed?
+    - if @investigation.is_private? || @investigation.is_closed? || @investigation.coronavirus_related?
```
slorek, line 5: Can't we re-use the partial here?

psd-web/config/locales/en.yml
frankieroberto: Good idea.

psd-web/app/views/investigations/_pages_top.html.slim
frankieroberto: I was trying to resist refactoring just yet 😄

frankieroberto: Ok, I un-resisted. f27e075e4a8ef05987fd2b2912b076e5073957f0

psd-web/app/views/investigations/_case_card_status_area.html.erb
```diff
@@ -0,0 +1,5 @@
+<% if investigation.is_private? || investigation.is_closed? || investigation.coronavirus_related? %>
+  <div class="psd-case-card__status-area">
+    <%= render "investigations/case_badges", investigation: investigation, classes: "psd-case-card__badge" %>
```
slawosz, line 3: From the `_case_badges` I see that you are checking for this conditions twice. I would just move all HTML from here to `_case_badges` and display it always, and let conditions do the rest.

psd-web/app/views/investigations/_pages_top.html.slim
```diff
@@ -2,13 +2,8 @@
...
```
slawosz, line 5: As comment above, and how about moving this to erb?

psd-web/app/views/investigations/_case_card_status_area.html.erb
frankieroberto: @slawosz `_case_badges` only checks the conditions individually. This checks whether to render the container element and the partial? We _could_ move the conditions and the container element into the partial, but the HTML class for the container element differs depending on the context, so I thought it was better to leave it outside.

psd-web/app/views/investigations/_pages_top.html.slim
frankieroberto: I was trying to avoid unnecessary refactoring so as not to make the PR too big & hard to follow, but given that this is only a 9 line partial, I'm tempted...

PR #462, opened 2020-04-03
Add coronavirus question

slorek:
https://trello.com/c/b3K7oCvN/466-add-page-asking-if-case-is-related-to-coronavirus-for-trading-standards

## Description
Implements the question to ask MSA users if the case is related to the coronavirus.

I added an extra scenario to the existing feature spec since the existing one only tested the happy path and with every field completed. The new scenario tests a minimal input data set with some of the validations around those. Each scenario has a different value for the coronavirus question. Ideally there would be a few more scenarios with different input combinations but that is beyond the scope of this change. I also fixed the user having the wrong role in this feature spec.

Needless to say, the wizard controller implementation is pretty terrible, but I have not attempted to improve this, just the minimum required to get it working.

I have manually updated those live cases which should be set to `true` on production.

Notes:
- the banner is missing the 'read more' link as on the home page, as per the design.
- the "Report an unsafe or non-compliant product" caption has been added to the design to be consistent with the other pages in the journey.

### Default state:

Screenshot 2020-04-03 at 16 37 25

### Error state:

Screenshot 2020-04-03 at 16 38 11

### Activity log:

Screenshot 2020-04-03 at 16 38 58

### Case overview:

Screenshot 2020-04-03 at 16 39 19

## Checklist:

### General testing (author)
- [x] Test without JavaScript
- [x] Test on small screen

- [x] Works keyboard only

### DIFF
--- a/psd-web/app/controllers/investigations/ts_investigations_controller.rb
+++ b/psd-web/app/controllers/investigations/ts_investigations_controller.rb
@@ -9,7 +9,7 @@ class Investigations::TsInvestigationsController < ApplicationController
   set_attachment_names :file
   set_file_params_key :file
 
-  steps :product, :why_reporting, :which_businesses, :business, :has_corrective_action, :corrective_action,
+  steps :coronavirus, :product, :why_reporting, :which_businesses, :business, :has_corrective_action, :corrective_action,
         :other_information, :test_results, :risk_assessments, :product_images, :evidence_images, :other_files,
         :reference_number
 
@@ -35,7 +35,7 @@ class Investigations::TsInvestigationsController < ApplicationController
     %i[has_corrective_action risk_assessments product_images evidence_images other_files].include? step
   end
   before_action :store_product, only: %i[update], if: -> { step == :product }
-  before_action :store_investigation, only: %i[update], if: -> { %i[why_reporting reference_number].include? step }
+  before_action :store_investigation, only: %i[update], if: -> { %i[coronavirus why_reporting reference_number].include? step }
   before_action :store_why_reporting, only: %i[update], if: -> { step == :why_reporting }
   before_action :store_selected_businesses, only: %i[update], if: -> { step == :which_businesses }
   before_action :store_pending_businesses, only: %i[update], if: -> { step == :which_businesses }
@@ -221,9 +221,16 @@ private
   end
 
   def investigation_request_params
+    # This must be done first because the browser will send no params if no radio is selected
+    if step == :coronavirus
+      params[:investigation] ||= { coronavirus_related: nil }
+    end
+
     return {} if params[:investigation].blank?
 
     case step
+    when :coronavirus
+      params.require(:investigation).permit(:coronavirus_related)
     when :why_reporting
       params[:investigation][:hazard_description] = nil unless params[:investigation][:unsafe] == "1"
       params[:investigation][:hazard_type] = nil unless params[:investigation][:unsafe] == "1"
@@ -452,6 +459,9 @@ private
 
   def records_valid?
     case step
+    when :coronavirus
+      # Ideally we would only validate the coronavirus_related attribute here, but there are currently no other validations and this is simpler
+      @investigation.validate
     when :product
       @product.validate
     when :why_reporting
--- a/psd-web/app/decorators/investigation_decorator.rb
+++ b/psd-web/app/decorators/investigation_decorator.rb
@@ -49,6 +49,10 @@ class InvestigationDecorator < ApplicationDecorator
         actions: []
       },
       {
+        key: { text: "Coronavirus related", classes: classes },
+        value: { text: I18n.t(coronavirus_related, scope: "case.coronavirus_related"), classes: classes },
+      },
+      {
         key: { text: "Created by", classes: classes },
         value: { text: created_by, classes: classes },
         actions: []
@@ -76,10 +80,10 @@ class InvestigationDecorator < ApplicationDecorator
       rows[0][:actions] = [
         { href: h.status_investigation_path(investigation), text: "Change", classes: classes, visually_hidden_text: "status" }
       ]
-      rows[2][:actions] = [
+      rows[3][:actions] = [
         { href: h.new_investigation_assign_path(investigation), text: "Change", classes: classes, visually_hidden_text: "assigned to" }
       ]
-      rows[4][:actions] = [
+      rows[5][:actions] = [
         { href: h.new_investigation_activity_path(investigation), text: "Add activity", classes: classes }
       ]
     end
--- a/psd-web/app/models/investigation.rb
+++ b/psd-web/app/models/investigation.rb
@@ -12,6 +12,7 @@ class Investigation < ApplicationRecord
 
   validates :description, presence: true, on: :update
   validates :assignable_id, presence: { message: "Select assignee" }, on: :update
+  validates :coronavirus_related, inclusion: { in: [true, false] }
 
   validates_length_of :user_title, maximum: 100
   validates_length_of :description, maximum: 10000
@@ -64,6 +65,17 @@ class Investigation < ApplicationRecord
     self.assignable_id = entity&.id
     self.assignable_type = "User" if entity.is_a?(User)
     self.assignable_type = "Team" if entity.is_a?(Team)
+  end
+
+  def coronavirus_related=(value)
+    if value.is_a?(String)
+      value = case value
+              when "yes" then true
+              when "no" then false
+              end
+    end
+
+    super(value)
   end
 
   def status
--- a/psd-web/app/models/investigation.rb
+++ b/psd-web/app/models/investigation.rb
@@ -67,17 +67,6 @@ class Investigation < ApplicationRecord
     self.assignable_type = "Team" if entity.is_a?(Team)
   end
 
-  def coronavirus_related=(value)
-    if value.is_a?(String)
-      value = case value
-              when "yes" then true
-              when "no" then false
-              end
-    end
-
-    super(value)
-  end
-
   def status
     is_closed? ? "Closed" : "Open"
   end
--- a/psd-web/app/views/investigations/ts_investigations/coronavirus.html.erb
+++ b/psd-web/app/views/investigations/ts_investigations/coronavirus.html.erb
@@ -9,13 +9,13 @@
 <%= form_with scope: :investigation, model: @investigation, url: wizard_path, method: :put, local: true do |form| %>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds-from-desktop">
-      <%= render "form_components/govuk_error_summary", form: form %>
+      <%= error_summary @investigation.errors  %>
       <span class="govuk-caption-l">Report an unsafe or non-compliant product</span>
 
       <%= render "form_components/govuk_radios", form: form, key: :coronavirus_related,
               fieldset: { legend: { text: page_heading, classes: "govuk-fieldset__legend--l", isPageHeading: true } },
               hint: { text: "For example, cases relating to hand sanitiser or personal protective equipment (PPE)" },
-              items: [{ text: "Yes, it is (or could be)", value: "true" },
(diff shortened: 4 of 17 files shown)


psd-web/app/models/investigation.rb
```diff
@@ -66,6 +67,17 @@ def assignee=(entity)
...
+    if value.is_a?(String)
+      value = case value
+              when "yes" then true
+              when "no" then false
+              end
+    end
+
+    super(value)
+  end
+
```
frankieroberto, line 80: I _think_ if you set the radio values to `"true"` and `"false"` instead of `"yes"` and `"no"` then Rails should coerce them to boolean for you automatically?

psd-web/app/decorators/investigation_decorator.rb
```diff
@@ -76,10 +80,10 @@ def investigation_summary_list(include_actions: true, classes: "")
       rows[0][:actions] = [
         { href: h.status_investigation_path(investigation), text: "Change", classes: classes, visually_hidden_text: "status" }
       ]
-      rows[2][:actions] = [
+      rows[3][:actions] = [
```
slawosz, line 83: How about getting rid of `magic numbers` and using some `consts`?

slorek: maybe another time :)

psd-web/spec/models/audit_activity/investigation/add_allegation_spec.rb
```diff
@@ -18,5 +18,13 @@
     it "stores the body" do
       expect(activity.body).to eq("**Allegation details**<br><br>Hazard type: **#{allegation.hazard_type}**<br><br>#{allegation.description}<br><br>Assigned to #{allegation.assignee.display_name}.")
     end
+
+    context "when case is coronavirus related" do
+      let(:allegation) { create(:allegation, coronavirus_related: true) }
+
+      it "adds text to the body" do
+        expect(activity.body).to eq("**Allegation details**<br><br>Case is related to the coronavirus outbreak.<br>Hazard type: **#{allegation.hazard_type}**<br><br>#{allegation.description}<br><br>Assigned to #{allegation.assignee.display_name}.")
```
slawosz, line 26: I personally would like to have body as `let` with here document:
```
let(:expected_body) do
<<-BODY
  foo bar
BODY
end
```

psd-web/app/views/investigations/ts_investigations/coronavirus.html.erb
```diff
@@ -0,0 +1,24 @@
...
+<%= page_title page_heading, errors: @investigation.errors.any? %>
+
+<%= banner do %>
+  <p>Coronavirus (COVID-19): OPSS is urging all market surveillance authorities to prioritise coronavirus-related cases above other work.</p>
+<% end %>
+
+<%= form_with scope: :investigation, model: @investigation, url: wizard_path, method: :put, local: true do |form| %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds-from-desktop">
+      <%= render "form_components/govuk_error_summary", form: form %>
```
frankieroberto, line 12: Please change, use the new helper, as the "form_components/govuk_error_summary" partial has some bugs:

```suggestion
      <%= error_summary errors: @investigation.errors  %>
```

The error summary link should target the first radio button `<input>`, so that it focuses upon click (see [Design System documentation](https://design-system.service.gov.uk/components/error-summary/#linking-from-the-error-summary-to-each-answer).

psd-web/app/views/investigations/ts_investigations/coronavirus.html.erb
```diff
@@ -0,0 +1,24 @@
...
+      <span class="govuk-caption-l">Report an unsafe or non-compliant product</span>
+
+      <%= render "form_components/govuk_radios", form: form, key: :coronavirus_related,
+              fieldset: { legend: { text: page_heading, classes: "govuk-fieldset__legend--l", isPageHeading: true } },
+              hint: { text: "For example, cases relating to hand sanitiser or personal protective equipment (PPE)" },
+              items: [{ text: "Yes, it is (or could be)", value: "true" },
+                      { text: "No, this is business as usual", value: "false" }] %>
+
+      <%= form.submit "Continue", class: "govuk-button" %>
```
frankieroberto, line 21: Please change, use the helper:

```suggestion
      <%= govukButton(text: "Continue") %>
```

(This triggers [some javascript](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/button/button.js) to make the button respond to the space bar)

psd-web/spec/features/report_product_spec.rb
```diff
@@ -317,14 +448,19 @@ def expect_case_activity_page_to_show_test_result(test)
     expect(item).to have_text(test[:details])
   end
 
+  def fill_in_coronavirus_page(answer)
+    choose answer ? "Yes, it is (or could be)" : "No, this is business as usual"
```
frankieroberto, line 452: Suggest scoping the the answer to a fieldset, to help make sure that the HTML is correct and improve readability of the spec:

```suggestion
    within_fieldset("Is this case related to the coronavirus outbreak?") do
        choose answer ? "Yes, it is (or could be)" : "No, this is business as usual"
    end
```

psd-web/spec/features/report_product_spec.rb
```diff
@@ -317,14 +448,19 @@ def expect_case_activity_page_to_show_test_result(test)
     expect(item).to have_text(test[:details])
   end
 
+  def fill_in_coronavirus_page(answer)
```
frankieroberto, line 451: Suggest renaming of the helper method:

```suggestion
  def answer_is_case_coronavirus_related_with(answer)
```

psd-web/app/controllers/investigations/ts_investigations_controller.rb
```diff
@@ -221,9 +221,16 @@ def product_session_params
   end
 
   def investigation_request_params
+    # This must be done first because the browser will send no params if no radio is selected
+    if step == :coronavirus
+      params[:investigation] ||= { coronavirus_related: nil }
```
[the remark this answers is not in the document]
slorek: I decided to do it this way because it is then explicit that the user has neglected to select an option. A boolean attribute should not have a third `nil` state to manage. I think the best way to avoid this would be to pre-select yes or no in the form, but that would change the interaction with the user fairly significantly. Otherwise we should have custom validation steps for each page on the form, but that would be a rewrite of the wizard.

psd-web/app/views/investigations/ts_investigations/coronavirus.html.erb
slorek: Should we not fix the partial instead? This is used throughout the other pages on the wizard.

psd-web/app/views/investigations/ts_investigations/coronavirus.html.erb
slorek: This is used on all of the other pages on the wizard so I'll have to change it on all pages.

psd-web/spec/features/report_product_spec.rb
slorek: I kept the naming consistent with all of the other `fill_in_page` helpers so I'd prefer to keep it the same for consistency. We can look at overall naming consistency across all tests separately.

psd-web/app/controllers/investigations/ts_investigations_controller.rb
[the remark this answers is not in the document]
frankieroberto: Yeah, I'm not sure we should pre-select Yes or No.  I had assumed that there was a single `Investigation.new()` call at the start of the flow where we could set it to `nil`, but looking at the code, that's not how it works... 😬  So this seems reasonable, especially given the validations!

psd-web/app/views/investigations/ts_investigations/coronavirus.html.erb
frankieroberto: I think we could probably find-and-replace `<%= render "form_components/govuk_error_summary", form: form %>` with `<%= error_summary errors: form.object.errors  %>` across the templates.

This wouldn't fix all the accessibility issues though, as the error summary helper in `error_summary()` just links to the anchor with the same name as the attribute (eg `#coronavirus_related` here) - but we have to manually set that on the _first_ radio button in order to get the right behaviour (and the helper doesn’t know the order of the radio/checkbox elements on the page)...

slorek: @frankieroberto the helper doesn't seem to work for radios. The first radio `id` is `investigation_coronavirus_related_true` but the helper is trying to target `coronavirus_related`. Suggest looking at this separately.

psd-web/app/decorators/investigation_decorator.rb
```diff
@@ -48,6 +48,10 @@ def investigation_summary_list(include_actions: true, classes: "")
         value: { text: status, classes: classes },
         actions: []
       },
+      {
+        key: { text: "Coronavirus related", classes: classes },
+        value: { text: I18n.t(coronavirus_related, scope: "case.coronavirus_related"), classes: classes },
```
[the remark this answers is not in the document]
slorek: @frankieroberto I did test this but it doesn't work (I wouldn't expect/want it to, as the values should be scoped to `coronavirus_related`).

psd-web/spec/models/audit_activity/investigation/add_allegation_spec.rb
slorek: There's no newlines in the output text so I don't think this would be more readable.

psd-web/app/views/investigations/ts_investigations/coronavirus.html.erb
frankieroberto: @slorek I think you can fix this by manually setting the ID on the first radio button item in the call to `<%= render "form_components/govuk_radios" %>` below, eg:

```
              items: [{ text: "Yes, it is (or could be)", value: "true", id: "coronavirus_related" },
```

Think we should fix this as part of the PR, as otherwise we'll be introducing an accessibility fail (an anchor link to an ID which doesn't exist).

frankieroberto: @slorek I _think_ you should be able to change the automatically-generated ID on the first radio item?

ie:

```ruby
      <%= render "form_components/govuk_radios", form: form, key: :coronavirus_related,
               fieldset: { legend: { text: page_heading, classes: "govuk-fieldset__legend--l", isPageHeading: true } },
               hint: { text: "For example, cases relating to hand sanitiser or personal protective equipment (PPE)" },
               items: [{ text: "Yes, it is (or could be)", value: "true", id: "coronavirus_related"},
                       { text: "No, this is business as usual", value: "false", id: "not_coronavirus_related" }] %>
```

slorek: The current implementation links to `#investigation_coronavirus_related-error` which is the ID for the error message in the fieldset.

psd-web/app/decorators/investigation_decorator.rb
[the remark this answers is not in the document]
frankieroberto: Doh, I didn't spot that the the key was a reference to a boolean rather than a symbol, sorry. 🤦‍♂

psd-web/app/views/investigations/ts_investigations/coronavirus.html.erb
frankieroberto: Yeah, this was one of the bugs reported in the accessibility audit. If you switch to the `<%= error_summary errors: @investigation.errors  %>` helper _and_ manually set the IDs for the radios as in the code snippet above, it should fix the issue (by linking to the ID of the first radio).

We have this issue across the application (but have been slowly fixing it, eg in the new devise views).

If this is taking too long, let's merge and create a card to fix later.

slorek: Fixed.

PR #469, opened 2020-04-06
Include Coronavirus flag in export

scruti:
Add Coronavirus flag to XLS exported investigations file.


psd-web/test/controllers/investigations_xls_export_test.rb
```diff
@@ -23,13 +23,26 @@ class InvestigationsXlsExportTest < ActionDispatch::IntegrationTest
...
+    Investigation::Allegation.new(coronavirus_related: true).save
+    Investigation.import refresh: true, force: true
+
+    file = get_file(path)
+    coronavirus_cell_title = file.sheet("Cases").cell(1, 3)
+    coronavirus_cell_content = file.sheet("Cases").cell(2, 3)
+    assert_equal "Coronavirus_Related", coronavirus_cell_title
+    assert_equal "Yes", coronavirus_cell_content
+    File.delete(path)
+  end
```
scruti, line 44: Added this test through TDD and left it for PR review as a good way to prove that "works" to reviewers. Happy to remove it before merging if is considered an expensive overkill test-wise.

scruti:
[(1) Include coronavirus flag in export.](https://trello.com/c/SdZAS6y1/470-1-include-coronavirus-flag-in-export)

psd-web/test/controllers/investigations_xls_export_test.rb
```diff
@@ -30,6 +30,19 @@ class InvestigationsXlsExportTest < ActionDispatch::IntegrationTest
...
+    Investigation::Allegation.new(coronavirus_related: true).save
+    Investigation.import refresh: true, force: true
+
+    file = get_file(path)
+    coronavirus_cell_title = file.sheet("Cases").cell(1, 8)
+    coronavirus_cell_content = file.sheet("Cases").cell(2, 8)
+    assert_equal "Coronavirus_Related", coronavirus_cell_title
+    assert_equal "true", coronavirus_cell_content
+    File.delete(path)
+  end
```
scruti, line 44: Added this test through TDD and left it for PR review as a good way to prove that "works" to reviewers. Happy to remove it before merging if is considered an expensive overkill test-wise.

slorek: I think it's worthwhile but we should do all new tests in RSpec and not MiniTest.

scruti: Going to migrate the file to a RSpec request spec then.

nasirkhanbeis:
functionally works fine as expected, not sure about performance impact though.

psd-web/test/controllers/investigations_xls_export_test.rb
scruti: @slorek moved them to Rspec.

StupidCodeFactory:
> functionally works fine as expected, not sure about performance impact though.

It should not have any impact on the performance as this only adds a column from a row that's already been fetch from the database.

psd-web/spec/requests/export_investigations_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    let(:user) { create(:user, :activated, :psd_user, :viewed_introduction) }
+    let(:temp_dir) { "spec/tmp/" }
+    let(:export_path) { Rails.root + temp_dir + "export_cases.xlsx" }
+    let(:exported_data) do
+      File.open(export_path, "w") { |f| f.write response.body }
+      Roo::Excelx.new(export_path).sheet("Cases")
+    end
+
+    before do
+      Dir.mkdir(temp_dir) unless Dir.exist?(temp_dir)
```
StupidCodeFactory, line 15: Can you use `Dir::mktmpdir` with a block as it cleans itself up?

scruti: Nice suggestion. Going to change it.

psd-web/app/views/investigations/index.xlsx.axlsx
```diff
@@ -14,6 +14,7 @@ book.add_worksheet name: "Cases" do |sheet_investigations| # rubocop:disable Met
       investigation.description,
       investigation.product_category,
       investigation.hazard_type,
+      investigation.coronavirus_related?,
```
slawosz, line 17: I think you can use microsevices in case we will have possible performance issues in 3 years :trollface:

psd-web/spec/requests/export_investigations_spec.rb
```diff
@@ -0,0 +1,64 @@
+require "rails_helper"
+
+RSpec.describe "Export investigations as XLSX file", :with_elasticsearch, :with_stubbed_notify, :with_stubbed_mailer, type: :request do
```
slawosz, line 3: :1st_place_medal:

psd-web/app/views/investigations/index.xlsx.axlsx
scruti: LOL

psd-web/spec/requests/export_investigations_spec.rb
scruti: Ended reverting it as, when using the `Dir::mktmpdir` inside an `around` block, was generating a dynamically named folder that would force `File.open` to be also called inside the around block in order to have the correct dir/file name for writing the test XLSX file.  Didn't want to run the whole test case inside a `File.open` scope.

PR #474, opened 2020-04-09
Create team task

slorek:
https://trello.com/c/daoKianX/488-add-herefordshire-council-and-tracey-as-team-admin

## Description
Adds a Rake task to enable a new org, team and admin user to be added easily. I have created a service so that this business logic is encapsulated somewhere sensible and in future this could potentially be re-used by the planned platform admin feature.

Currently there is no validation for uniqueness on team or org names. I will address this separately.

I've added usage docs, of course anyone who runs this in prod will need sufficient access.

Also fixed some bugs in `SendUserInvitationJob` and restructured the specs.

## Checklist:


psd-web/app/services/create_organisation.rb
```diff
@@ -0,0 +1,25 @@
+class CreateOrganisation
+  include Interactor
+
+  def call
+    context.fail!(error: "No organisation name supplied") unless context.org_name
```
frankieroberto, line 5: There's already a validation on the Organisation model for the name. Could you rely on that, rather than adding a custom message here?

psd-web/spec/services/create_organisation_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+RSpec.describe CreateOrganisation, :with_stubbed_mailer do
+  let(:email) { Faker::Internet.safe_email }
+  let(:org_name) { Faker::Team.name }
+
+  describe ".call" do
+    context "with no parameters" do
+      let(:result) { described_class.call }
+
+      it "returns a failure" do
+        expect(result.failure?).to be true
```
StupidCodeFactory, line 12:
```suggestion
        expect(result).to be_a_failure
```

psd-web/spec/services/create_organisation_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+        end
+
+        it "does not save the user" do
+          expect(created_user).to be_nil
+        end
+      end
+
+      context "when no exceptions are raised" do
+        it "returns success" do
+          expect(result.success?).to be true
```
StupidCodeFactory, line 63:
```suggestion
          expect(result).to be_a_success
```

psd-web/app/services/create_organisation.rb
```diff
@@ -0,0 +1,25 @@
...
+    context.fail!(error: "No team admin email supplied") unless context.admin_email
+
+    ActiveRecord::Base.transaction do
+      context.org = Organisation.create!(name: context.org_name)
+      context.team = context.org.teams.create!(name: context.org_name)
```
frankieroberto, line 10: Do we also need to support setting the `team_recipient_email` (optionally)?

psd-web/app/services/create_organisation.rb
```diff
@@ -0,0 +1,25 @@
+class CreateOrganisation
```
frankieroberto, line 1: It took me a few mins to realise that this sets up a bunch of different models, so I wonder whether a more verbose class name would help:

```suggestion
class CreateOrganisationWithTeamAndAdminUser
```

psd-web/spec/services/create_organisation_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+
+        it "sends an email to the user" do
+          result
+          expect(invitation_email.recipient).to eq(email)
+        end
+
+        it "includes the invitation URL in the email" do
+          result
+          expect(invitation_email.personalization_path(:invitation_url)).to eq complete_registration_user_path(created_user.id, invitation: created_user.invitation_token)
+        end
```
StupidCodeFactory, line 116: Nitpick: it's a bit odd to test the content of the email as this does not send the email at all
I'd rather test that the job has been enqueued with correct argument:
```ruby
expect { result } .to have_enqueued_job(SendUserInvitationJob).at(:no_wait).on_queue("psd").with(user.id)
```
This actually matches what the service is doing.

psd-web/lib/tasks/create_organisation.rake
```diff
@@ -0,0 +1,7 @@
+namespace :organisation do
+  desc "Creates an organisation, team, and team admin user, and sends an invitation email to the user"
+  task create: :environment do
```
frankieroberto, line 3: Could use rake task arguments here, instead of environment variables? (not necessarily easier though)

```suggestion
  task :create, [:organisation_name, :admin_email] => :environment do
```

psd-web/app/services/create_organisation.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    ActiveRecord::Base.transaction do
+      context.org = Organisation.create!(name: context.org_name)
+      context.team = context.org.teams.create!(name: context.org_name)
+
+      context.user = context.team.users.create!(
+        email: context.admin_email,
+        organisation: context.org,
+        skip_password_validation: true,
+        invitation_token: SecureRandom.hex(15)
```
frankieroberto, line 16: I think we also need to set an `invited_at` timestamp, as the existing `SendUserInvitationJob` assumes that if there is no `user_inviting_id` present, the user requested a resend of their own invitation, and so the timestamp isn't set (and `has_been_sent_welcome_email` is also not set).

Alternatively, we could adjust the `SendUserInvitationJob` (which wasn't really designed for this use-case) to accept an additional param, or create a new `SendTeamAdminUserInvitationJob` or something? (Depending on if we think the e-mail invite should be different)

frankieroberto: Hmm actually it looks like if the `invited_at` timestamp is `nil`, then the invitation be valid and will never expire... 😄  That feels like a slightly accidental feature...

psd-web/app/services/create_organisation.rb
slorek: I think it's better to make the interface clear (as I don't think interactor has a pattern to specify input arguments) and escape early.

psd-web/app/services/create_organisation.rb
slorek: Not within the scope of request on the Trello ticket, if needed we can add this later.

psd-web/app/services/create_organisation.rb
slorek: I'll modify the job to set these attributes, I think this use case was missed.

psd-web/lib/tasks/create_organisation.rake
slorek: We _could_ do this but I find the interface is pretty poor. In this case we would have to write it as:

```bash
bundle exec rake organisation:create\[test\,\ test7@example.com\]
```

No named arguments, and having to remember to do nasty escaping.

psd-web/app/services/create_organisation_with_team_and_admin_user.rb
```diff
@@ -0,0 +1,25 @@
+class CreateOrganisationWithTeamAndAdminUser
+  include Interactor
```
slawosz, line 2: Whats the benefit or getting `Interactor` here? It does not improve nothing :( Only adds complexity over adding arguments to method which has literally 16 lines of code. Please read: https://martinfowler.com/bliki/AnemicDomainModel.html :)

slorek: It provides the success/fail interface, and is consistent with our decision to use `interactor` for service objects which will provide a consistent interface.

PR #475, opened 2020-04-09
Apply Coronavirus question page to all opss flows

scruti:
[Trello ticket](https://trello.com/c/qudWvkHJ/468-2-apply-coronavirus-question-page-to-all-opss-flows)

## Description

Apply coronavirus question page to all OPSS flows:
- Allegation
- Enquiry
- Project

This work is heavily based on previous work done by @slorek [adding the question for trading standards](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/462)

## Implementation details
- Big part of the work was to switch Projects from a single page controller to a Wizard with steps. 
- Introduced a new Form Object for Coronavirus question.
- Changed the Coronavirus question edition page to use the newly introduced Form Object.
- All the flows  share a controller concern to use the Coronavirus form in their `coronavirus` step.
- Few Minitest system tests have been migrated to their corresponding Rspec feature tests.
 

## Checklist:

image


scruti:
[(2) Apply coronavirus question page to all OPSS flows](https://trello.com/c/qudWvkHJ/468-2-apply-coronavirus-question-page-to-all-opss-flows)

edwardhorsford:
Three minor issues:

OPSS flows have back links at the top:
Screenshot 2020-04-14 at 11 33 28

The line of text  `Report an unsafe or non-compliant product` is only for TS flows.
Screenshot 2020-04-14 at 11 35 35

Perhaps we can see how other pages that are shared between the flows include this line.

The URL path for projects seems to have changed from `domain/project/foo` to `domain/investigation/project/foo` -  is this intentional?

psd-web/app/controllers/investigations/project_controller.rb
```diff
@@ -1,27 +1,67 @@
...
+  def new
+    session.delete :investigation
+    redirect_to wizard_path(steps.first)
+  end
+
+  # POST /xxx
   def create
-    if @investigation.valid?
+    if @investigation.valid?(step)
       @investigation.save
```
StupidCodeFactory, line 22: Will this still save if it does not pass validation from broader context?

psd-web/app/controllers/investigations/project_controller.rb
```diff
@@ -1,27 +1,67 @@
...
+    end
+  end
+
+  # PATCH/PUT /xxx
+  def update
+    return render_wizard unless @investigation.valid?(step)
+
+    session[:investigation] = @investigation.attributes
+
+    if step == steps.last
```
StupidCodeFactory, line 35: why isn't the last step posting instead of patching? There is something wrong with this pattern calling the create from the update seems counterintuitive at best.

psd-web/app/controllers/investigations/project_controller.rb
```diff
@@ -1,27 +1,67 @@
...
+      create
+    else
+      redirect_to next_wizard_path
     end
   end
 
 private
 
+  def investigation_session_params
+    session[:investigation] || {}
```
StupidCodeFactory, line 45: Can't we resend the params between each wizard state like in normal web application instead of messing with the session?

psd-web/app/controllers/investigations/project_controller.rb
```diff
@@ -1,27 +1,67 @@
...
+    investigation_session_params.merge(investigation_request_params)
+  end
+
   # Never trust parameters from the scary internet, only allow the white list through.
-  def project_params
+  def investigation_request_params
+    # This must be done first because the browser will send no params if no radio is selected
+    if step == :coronavirus
+      params[:investigation] ||= { coronavirus_related: nil }
+    end
```
StupidCodeFactory, line 57: same comment as above this can be remove by adding a hidden input in the form

psd-web/app/views/investigations/coronavirus.html.erb
```diff
@@ -0,0 +1,24 @@
...
+<%= banner do %>
+  <p>Coronavirus (COVID-19): OPSS is urging all market surveillance authorities to prioritise coronavirus-related cases above other work.</p>
+<% end %>
+
+<%= form_with scope: scope, model: @investigation, url: wizard_path, method: :put, local: true do |form| %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds-from-desktop">
+      <%= error_summary @investigation.errors  %>
+      <span class="govuk-caption-l">Report an unsafe or non-compliant product</span>
+
```
StupidCodeFactory, line 14:
```suggestion
form.hidden_field :coronavirus_related
```

psd-web/spec/features/create_enquiry_as_opss_user_spec.rb
```diff
@@ -88,15 +97,45 @@
...
+      .to have_sibling("dd", text: "Coronavirus related case")
+  end
+
+  def expect_details_on_activity_page(contact, enquiry)
+    expect(page).to have_text("Enquiry logged: #{enquiry.fetch(:enquiry_title)}")
+    expect(page).to have_text(enquiry.fetch(:enquiry_description))
+    expect(page).to have_text("Attachment: testImage.png")
+    expect(page).to have_text("Name: #{contact.fetch(:contact_name)}")
+    expect(page).to have_text("Email address: #{contact.fetch(:contact_email)}")
+    expect(page).to have_text("Phone number: #{contact.fetch(:contact_phone)}")
```
StupidCodeFactory, line 137: We're using `have_css` everywhere else maybe here also just for the sake on the test being a bit more robust to unwanted presentation/HTML changes.

psd-web/app/controllers/investigations/creation_flow_controller.rb
```diff
@@ -140,6 +145,11 @@ def complainant_request_params
 
   # Never trust parameters from the scary internet, only allow the white list through.
   def investigation_request_params
+    # This must be done first because the browser will send no params if no radio is selected
+    if step == :coronavirus
+      params[model_key] ||= { coronavirus_related: nil }
+    end
```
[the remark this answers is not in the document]
scruti: Problem is that we don't want to have any value pre-selected when displaying the Coronavirus question, so if this gets removed, the radio-button checks "No" as default instead.
Discussion where this was introduced for TS flow: https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/462#discussion_r403108740

psd-web/app/controllers/investigations/project_controller.rb
```diff
@@ -1,27 +1,67 @@
 class Investigations::ProjectController < ApplicationController
-  before_action :set_investigation, only: %i[new create]
+  include Wicked::Wizard
 
-  def new; end
+  steps :coronavirus, :project_details
 
+  before_action :set_investigation, only: %i[show new create update]
+
+  #GET /xxx/step
```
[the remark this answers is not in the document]
scruti: Brought it from the rest of controllers using `Wicked::Wizard`. It means that the different steps display will map to the `#show` action.

psd-web/app/controllers/investigations/project_controller.rb
scruti: Wops, this was meant to be only in the `update`. On creation the full Investigation validations should be  valid. Good catch.

psd-web/app/controllers/investigations/creation_flow_controller.rb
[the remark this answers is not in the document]
StupidCodeFactory: So adding a hidden field  with  a  nil value would remove the need for this if nothing is selected. And we can remove this hack.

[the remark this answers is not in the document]
slorek: worth a try, I didn't think of this.

psd-web/app/controllers/investigations/project_controller.rb
scruti: **I guess** this has been done through all the flows so the steps can be re-ordered without needing to change the forms every time the last step changes for the flow.
In this way all the step views send an update form and it works out of the box when reordering or adding steps into the controller.

psd-web/app/controllers/investigations/creation_flow_controller.rb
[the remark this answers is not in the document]
scruti: Tested the proposed hidden field but was defaulting to "false", going to double check it in case I missed something.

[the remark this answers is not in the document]
StupidCodeFactory: I know why it does this. It because coronavirus_related is a boolean field that defaults to "false" 
you probably `explicitly` tell the field to have nothing selected as well as adding the hidden field. 
```
f.hidden_field :coronavirus_related, nil
f.input_field :coronavirus_related, nil
```
Coupling HTML form fields and database record isn't ideal.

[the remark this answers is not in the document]
scruti: From what I understand, the hidden field will set a value **on the form submission**. 
The problem is that we need this `coronavirus_related: nil` set when displaying the form before being submitted. So the hidden field does not fit for this purpose? 

An alternative approach I found to the params hack is... maybe another hack that, at least, requires not to be checked for every single request submission, but just on `Controller#new`.

```ruby
  def new
    session.delete :investigation
    redirect_to wizard_path(steps.first, investigation: { coronavirus_related: nil })
  end
```
We save the conditional (and checking it for every params setup for every step), but will only work while `coronavirus` is the first step in the flow.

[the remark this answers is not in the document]
scruti: @StupidCodeFactory  missed your last comment: The problem is we're using [a GovUK helper for the radio buttons ](https://github.com/UKGovernmentBEIS/govuk-design-system-rails/blob/master/app/views/form_components/_govuk_radios.html.slim) that does not seem easy to override as it defaults to the DB value.
I guess that is why Steve used this hack in order to enforce the default to nil.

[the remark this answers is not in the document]
StupidCodeFactory: so I you pass a options[:value] than it should not check it unless I can not read this code:
https://github.com/UKGovernmentBEIS/govuk-design-system-rails/blob/master/app/views/form_components/_govuk_radios.html.slim#L17-L18

nasirkhanbeis:
Retested this in review app 475, looking good as expected. 
also activity log records CV flag if the case is related. 
Verified project/enquiry/allegation as opss user, also regression tested TS flow, all good.

psd-web/app/controllers/concerns/flow_with_coronavirus_form.rb
```diff
@@ -0,0 +1,24 @@
...
+  included do
+    before_action :set_new_coronavirus_form, only: :show, if: -> { step == :coronavirus }
+  end
+
+  def set_new_coronavirus_form
+    @coronavirus_related_form = CoronavirusRelatedForm.new
+  end
+
+  def assigns_coronavirus_related_from_form(investigation, coronavirus_form_params)
+    form = coronavirus_related_form(coronavirus_form_params)
```
StupidCodeFactory, line 13: So re-reading this with a fresh mind today, I'd suggest we let the caller instantiate  the form object outside this method and inject it instead of the params. This also make the name of this function actually does what it says ;)

psd-web/app/controllers/investigations/project_controller.rb
```diff
@@ -1,27 +1,73 @@
...
-    @investigation = Investigation::Project.new(project_params).decorate
+    @investigation = Investigation::Project.new(investigation_params).decorate
+  end
+
+  def coronvirus_form_params
+    params.require(:investigation).permit(:coronavirus_related)
+  end
+
+  def investigation_valid?
+    if step == :coronavirus
```
StupidCodeFactory, line 67: State of mind: I don't like this `if` at all...

psd-web/app/models/investigation.rb
```diff
@@ -12,7 +12,6 @@ class Investigation < ApplicationRecord
 
   validates :description, presence: true, on: :update
   validates :assignable_id, presence: { message: "Select assignee" }, on: :update
-  validates :coronavirus_related, inclusion: { in: [true, false] }
```
StupidCodeFactory, line 15: **Note:** We decided to remove this as this is no longer used anywhere and the database already enforces a boolean type, has a sensible default and rails will on top of this cast any value to its boolean equivalent. I think there is nor harm in removing this and duplicate this in the form.

psd-web/app/controllers/investigations/project_controller.rb
scruti: Don't see escape from it while with `Wicked:Wizard` approach of re-using the controller methods for all the different steps :man_shrugging:. More steps, more `if/cases`...

psd-web/app/forms/coronavirus_related_form.rb
```diff
@@ -0,0 +1,9 @@
+class CoronavirusRelatedForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attribute :coronavirus_related, :boolean, default: nil
+
+  validates :coronavirus_related,
+            inclusion: { in: [true, false], message: "Select whether or not the case is related to the coronavirus outbreak" }
```
frankieroberto, line 8: Please change: use the message from the locale file instead. (It's already there, but under a different scope)

psd-web/app/controllers/investigations/project_controller.rb
scruti: Resolved while pairing.

psd-web/app/views/investigations/coronavirus.html.erb
```diff
@@ -0,0 +1,26 @@
...
+  <p>Coronavirus (COVID-19): OPSS is urging all market surveillance authorities to prioritise coronavirus-related cases above other work.</p>
+<% end %>
+
+<%= form_with scope: scope, model: @coronavirus_related_form, url: wizard_path, method: :put, local: true do |form| %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds-from-desktop">
+      <%= error_summary @coronavirus_related_form.errors  %>
+      <% if unsafe_product_message %>
+        <span class="govuk-caption-l">Report an unsafe or non-compliant product</span>
+      <% end %>
```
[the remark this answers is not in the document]
scruti: Way better.

psd-web/app/views/investigations/ts_investigations/coronavirus.html.erb
```diff
@@ -1,24 +1,2 @@
...
-              hint: { text: "For example, cases relating to hand sanitiser or personal protective equipment (PPE)" },
-              items: [{ text: "Yes, it is (or could be)", value: "true", id: "coronavirus_related" },
-                      { text: "No, this is business as usual", value: "false" }] %>
-
-      <%= govukButton(text: "Continue") %>
-    </div>
-  </div>
-<% end %>
+<%= render template: "investigations/coronavirus",
+    locals: { scope: :investigation, unsafe_product_message: true } %>
```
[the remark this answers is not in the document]
scruti: Double checked on this and, sadly, attempting that would introduce a new level of pain in an already complex Controller as `Wicked::Wizard` requires direct matches of the step names with the view names under its controller.
Seems could be tweaked with the `wizard_value` and `I18n` (https://github.com/zombocom/wicked#internationalization-of-urls-i18n) , but I think is not worth the trade-off.

psd-web/app/controllers/investigations/creation_flow_controller.rb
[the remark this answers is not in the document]
scruti: Solved over pairing. :raised_hands:

PR #476, opened 2020-04-09
Allow safe products to be reported

frankieroberto:
**CHANGES:**

https://trello.com/c/E0ShTNHr/464-5-add-support-for-reporting-safe-products

1. Allow to mark an investigation as safe, unsafe, non compliant and, unsafe and non compliant
2. Custom Javascript to allow sets of checkboxes to be automatically unchecked if another checkbox  in another set is checked
3. This state will be store in a rails enum backed by a Postgres enum.

Screenshot 2020-04-21 at 11 02 09


nasirkhanbeis:
@StupidCodeFactory error styling seems not correct.
image
Need to display error message only once (not to repeat) and highlight the fields when nothing selected

slorek:
Issues found during testing:

* When nothing is selected and the user clicks Continue, the error message appears three times (as per Nasir's screenshot)
* Selecting safe plus one or more other options with Javascript and clicking Continue gives three errors: Screenshot 2020-04-23 at 10 51 11

psd-web/app/forms/why_reporting_form.rb
```diff
@@ -0,0 +1,78 @@
...
+
+private
+
+  def reported_reason
+    @reported_reason ||= case checkboxes
+                         when [true,  false, false] then Investigation.reported_reasons[:unsafe]
+                         when [false, true,  false] then Investigation.reported_reasons[:non_compliant]
+                         when [true,  true,  false] then Investigation.reported_reasons[:unsafe_and_non_compliant]
+                         when [false, false, true]  then Investigation.reported_reasons[:safe_and_compliant]
+                         end
```
slorek, line 32: Please change: this is really unnecessary complication and can be much more simply expressed something like:

```suggestion
return :unsafe_and_non_compliant if reported_reason_unsafe && reported_reason_non_compliant
return :unsafe if reported_reason_unsafe
return :non_compliant if reported_reason_non_compliant
return :safe_and_compliant if reported_reason_safe_and_compliant
```
There's other ways this could be expressed but I think using the `checkboxes` in this way is quite difficult to follow.

psd-web/db/migrate/20190321122934_populate_assignee_and_description.rb
```diff
@@ -1,7 +1,7 @@
 class PopulateAssigneeAndDescription < ActiveRecord::Migration[5.2]
   def change
     Investigation.all.each do |i|
-      i.description = i.description || i.reason_created
+      i.description = i.description
```
slorek, line 4: This line is redundant now.

psd-web/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "license": "MIT",
   "engines": {
-    "node": "10.19.x",
+    "node": "10.x",
```
slorek, line 6: The Ruby CloudFoundry buildpacks need specific versions: https://github.com/cloudfoundry/ruby-buildpack/releases

psd-web/spec/features/report_product_spec.rb
```diff
@@ -195,6 +199,73 @@
       end
     end
 
+    scenario "reporting a safe product" do
```
slorek, line 202: A lot of the steps in this scenario aren't required and can be dropped. We only need one test doing all of these steps.

psd-web/app/assets/javascripts/mutually_exclusive.js
```diff
@@ -0,0 +1,34 @@
...
+
+  const unCheck = checkbox => {
+    if (checkbox.checked) {
+      checkbox.click()
+      checkbox.checked = false
+    }
+  }
+}
+
+$(document).ready(() => {
```
frankieroberto, line 29: Think we can switch to this and avoid requiring jQuery:

```suggestion
document.addEventListener('DOMContentLoaded', () => {
```

psd-web/app/assets/javascripts/mutually_exclusive.js
```diff
@@ -0,0 +1,34 @@
+import $ from 'jquery'
```
frankieroberto, line 1: ✂️  Think we should avoid using jQuery, to bring us more in line with the [GOV.UK Frontend coding standards](https://github.com/alphagov/govuk-frontend/blob/master/docs/contributing/coding-standards/js.md) so that components can be upstreamed into the Design System easier.

Will also reduce javascript weight once we can drop jquery completely.

PR #480, opened 2020-04-13
Different approach to 2FA

slawosz:
# Allows to have 2FA by controller basis

Easy to extend 2FA solution. Replaces `two_factor_authentication` gem. Already done for general 2FA and reset password. Please note, that essential 2FA feature specs are working with only cosmetic changes (eg. different way to retrieve `OTP`)

## Authentication process

When user enters action which require 2FA, `SecondaryAuthentication` record is being searched against the user. **Search is limited only to records which are known for current browser session.**

### Creating new 2FA authentication
When no valid `SecondaryAuthentication` record is found, OTP code is being send and user is being redirected to authentication page. On authentication page, user enters `OTP` code. If code is correct, authentication record is marked as `authenticated` and **record is added to session**.

## SecondaryAuthentication model

Entity which stores information required for 2FA authentication:
* `user_id`
* `operation` - operation to authenticate against, eg: `change_password`. Operation is defined by controller that need to be authenticated. Default operation in below implementation is `secondary_authentication`
* `direct_otp` - code to authenticate against
* `authenticated` - flag which indicates if given authentication was successful. Will be set only by providing correct `otp`
* `expired?` - method to examine if given authentication can be still used
Expiration is different for different operation

## SecondaryAuthenticationConcern

Module which is included in `ApplicationController`. Provides all authentication methods.

## Using with arbitrary controller

To add custom 2FA to controller, controller need to override 1 ot 2 methods:
* `user_id_for_secondary_authentication` - provides id for user to authenticate. In most cases this method doesn't need to be changed as it defaults to `current_user`, however when token is used as during reset password flow, custom implementation is required.
* `current_operation` returns string to determine operation.

# TODO for production standard

- add specs for `SecondaryAuthentication` model
- add specs for `SecondaryAuthenticationConcern` for proper workflow and session manipulation.
- handle token validation
- add login attempts limitation to `SecondaryAuthentication`

# Comments for reviewers

Please comment only on logic flaws or potential issues, as it is work in progress.


psd-web/app/controllers/users/passwords_controller.rb
```diff
@@ -129,5 +112,15 @@ def after_sending_reset_password_instructions_path_for(_resource_name)
     def after_resetting_password_path_for(_resource)
       password_changed_path
     end
+
```
slawosz, line 115: This is example, how new 2FA flow is used for custom actions. By overriding `user_id_for_secondary_authentication` and `current_operation`, controller is required to created different `SecondaryAuthentication`, when `SecondaryAuthentication`, for default `current_operation` still exists.

psd-web/app/controllers/users/passwords_controller.rb
```diff
@@ -6,33 +6,16 @@ class PasswordsController < Devise::PasswordsController
...
+    skip_before_action :require_secondary_authentication, only: [:new, :create]
+
     def edit
       return render :invalid_link, status: :not_found if reset_token_invalid?
       return render :signed_in_as_another_user, locals: { reset_password_token: params[:reset_password_token] } if wrong_user?
       return render :expired, status: :gone if reset_token_expired?
 
       @email = user_with_reset_token.email
 
-      if passed_two_factor_authentication?
```
slawosz, line 16: New approach simplifies this logic.

nasirkhanbeis:
Hi @slawosz  I tested this in review app 480, over all functionality works fine as expected. Just few minor issues to address such as
1. Resend link on after 2fa returning page not found
2.On sign-in 2fa page,I see your account link displayed(where as in stage we don't show), its not an issue but when I click on your-account link it triggers another 2fa code(not a big issue but I suggest we can hide this your account link as in master)
3. URL is displaying secondary auth information such as id=xx (not sure if we want to show this in query param or not from security perspective) and this keeps incrementing on manual reload after validation error 
image
4.Not to show two validation errors at same time, eg. when I enter too many digits, the only error message we show is "You've entered too many digits" . we will show incorrect security code only when they enter expired code but in correct format(5 digit)
image
5.2fa code sent to be 5 digits(as we do in master)

edwardhorsford:
Just a note on the url: I'd expect the page to render at `/two-factor`:
Screenshot 2020-04-22 at 10 20 58

nasirkhanbeis:
Also secondary authentication page "content" should be slightly different as specified in other ticket https://trello.com/c/2cMZbcaZ/349-5-require-2fa-on-risky-actions-potentially-inviting-users. 
I my POV this pr is dependent on the content and can't be merge until we fully implement content and error message.

psd-web/app/controllers/concerns/secondary_authentication_concern.rb
```diff
@@ -0,0 +1,62 @@
...
+      unless SecondaryAuthentication.find_by(id: auth_id)
+        return true
+      end
+
+      if SecondaryAuthentication.find(auth_id).expired?
+        SecondaryAuthentication.find(auth_id).delete
+        true
+      else
+        false
+      end
```
scruti, line 42: Suggestion: Seems we're hitting the DB 3 times here for retrieving the same information. 
Can't we just get the `SecondaryAuthentication` instance once and operate over it? Or there was a explicit intent behind it (that I'm obviously missing)

```suggestion
      secondary_authentication = SecondaryAuthentication.find_by(id: auth_id)
      return true unless secondary_authentication

      if secondary_authentication.expired?
        secondary_authentication.delete
        true
      else
        false
      end
```

psd-web/app/controllers/concerns/secondary_authentication_concern.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+  end
+
+  def secondary_authentication_present_for_operation_and_user
```
scruti, line 46:
```suggestion
  def secondary_authentication_present_for_operation_and_user?
```

nasirkhanbeis:
pending issues to resolve
1. page not found on manual 2fa  page reload 
2. lock 2fa after 10 attempts and should work after an hour

psd-web/app/controllers/secondary_authentications_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+    @secondary_authentication_form = SecondaryAuthenticationForm.new(secondary_authentication_id: params[:secondary_authentication_id])
+  end
+
+  def create
+    params.permit!
+    @secondary_authentication_form = SecondaryAuthenticationForm.new(params[:secondary_authentication_form])
+
+    if @secondary_authentication_form.valid?
+      @secondary_authentication_form.authenticate!
+
```
scruti, line 23: :scissors:

psd-web/app/controllers/secondary_authentications_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+  end
+
+  def set_secondary_authentication_cookie_for(authentication)
+    session[:secondary_authentication] << authentication.id
+  end
+
+private
+
+   def hide_nav?
```
scruti, line 44:
```suggestion
  def hide_nav?
```

psd-web/app/controllers/users/passwords_controller.rb
```diff
@@ -6,33 +6,14 @@ class PasswordsController < Devise::PasswordsController
                        :has_viewed_introduction,
                        only: %i(edit sign_out_before_resetting_password)
 
-    def edit
-      return render :invalid_link, status: :not_found if reset_token_invalid?
-      return render :signed_in_as_another_user, locals: { reset_password_token: params[:reset_password_token] } if wrong_user?
-      return render :expired, status: :gone if reset_token_expired?
+    skip_before_action :require_secondary_authentication
+    before_action :secondary_authentication_before_edit, only: :edit
+    before_action :secondary_authentication_before_update, only: :update
```
scruti, line 11: Suggestion: I don't see why`require_secondary_authentication_before_update` has any reason to exist as adds 0 value over this call:
```suggestion
    before_action :require_secondary_authentication, only: :update
```

psd-web/app/controllers/users/passwords_controller.rb
```diff
@@ -73,12 +54,24 @@ def hide_nav?
 
   private
 
+    def secondary_authentication_before_edit
+      return render :signed_in_as_another_user, locals: { reset_password_token: params[:reset_password_token] } if wrong_user?
+      return render :invalid_link, status: :not_found if reset_token_invalid?
+      return render :expired, status: :gone if reset_token_expired?
+
+      require_secondary_authentication
+    end
```
scruti, line 63: Suggestion:
This method contains logic unrelated with second authentication and core to the controller edit action.

IMO this method should not exists and the inner calls should be embedded inside `edit`.

```ruby
def edit
  return render :signed_in_as_another_user, locals: { reset_password_token: params[:reset_password_token] } if wrong_user?
  return render :invalid_link, status: :not_found if reset_token_invalid?
  return render :expired, status: :gone if reset_token_expired?

  require_secondary_authentication

  @email = user_with_reset_token.email
  super
end
```

psd-web/app/controllers/users/sessions_controller.rb
```diff
@@ -5,6 +5,7 @@ class SessionsController < Devise::SessionsController
     # These methods trigger Warden authentication.
     # We don't want this to happen until we explicitly attempt to authenticate the user.
     skip_before_action :set_current_user, :set_raven_context, :authorize_user, only: :create
+    skip_before_action :require_secondary_authentication
```
scruti, line 8: :question: 
I do not understand what will happen now with the sessions. We do not require 2FA anymore?
We will require the mobile to be verified even when no requiring secondary authentication?

I'm clearly missing something here  :sweat_smile:

slawosz: For this action, we don't check 2FA. So you can access all actions without entering 2FA. Hope this answers your question.

psd-web/app/controllers/users/passwords_controller.rb
slawosz: It won't work. By default, its being checked before action, so we need to do all validations first.

scruti: Not sure we are in the same page about what I am proposing or why would not work. 
I changed it locally and passed tests, so opened a PR to make it clearer (and double check that effectively passes tests): https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/519

psd-web/app/controllers/users/sessions_controller.rb
scruti: Oh, I get it now. This is the submission of the sign-in. They will be required 2FA (or not) once redirected after sign-in.

psd-web/app/controllers/users/sessions_controller.rb
```diff
@@ -75,7 +76,7 @@ def add_wrong_credentials_errors(resource)
     end
 
     def mobile_not_verified?(user)
-      Rails.configuration.two_factor_authentication_enabled && user && !user.mobile_number_verified
+      user && !user.mobile_number_verified
```
scruti, line 79: If I understood this right this change will enforce users having their phone number verified even when 2FA is disabled.

In an environment where 2FA is not enabled this will block the user from signing in unless the `secondary_authentication#authenticate!` is called even when `Rails.configuration.secondary_authentication_enabled` is set to `false`

psd-web/app/forms/secondary_authentication_form.rb
```diff
@@ -0,0 +1,66 @@
+class SecondaryAuthenticationForm
```
scruti, line 1: Love this bit <3

psd-web/app/forms/secondary_authentication_form.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def authenticate!
+    secondary_authentication.authenticate!
+  end
+
+  def correct_otp_validation
+    return if errors.present?
+
+    unless secondary_authentication.valid_otp? self.otp_code
+      errors.add(:otp_code, "Incorrect security code")
```
scruti, line 39: Please change: Add these validation error messages to the `.otp_code` under `locales` file.

psd-web/app/models/secondary_authentication.rb
```diff
@@ -0,0 +1,76 @@
...
+  UNLOCK_OPERATION = "unlock_operation".freeze
+
+  TIMEOUTS = {
+    DEFAULT_OPERATION => 7 * 24 * 3600, # 7 days
+    RESET_PASSWORD_OPERATION => 300, # 5 minutes
+    INVITE_USER => 3600, # 1 hour
+    UNLOCK_OPERATION => 300, # 5 minutes
+  }.freeze
+
+  OTP_LENGTH = 6
```
scruti, line 14: Has been any change of requirements regarding the 2FA code length? Last time I worked on this the length was requested to be `5`.

psd-web/app/models/secondary_authentication.rb
```diff
@@ -0,0 +1,76 @@
+class SecondaryAuthentication < ApplicationRecord
```
scruti, line 1: Nice to see all this configuration and logic wrapped in a class :clap:

psd-web/app/models/user.rb
```diff
@@ -5,16 +5,17 @@ class User < ApplicationRecord
 
   # Include default devise modules. Others available are:
   # :confirmable, :lockable, :timeoutable, :registerable, :trackable and :omniauthable
-  devise :two_factor_authenticatable, :database_authenticatable, :timeoutable, :trackable, :rememberable, :validatable, :recoverable, :encryptable, :lockable
+  devise :database_authenticatable, :timeoutable, :trackable, :rememberable, :validatable, :recoverable, :encryptable, :lockable
 
   belongs_to :organisation
 
-  has_one_time_password(encrypted: true)
+  # has_one_time_password(encrypted: true)
```
scruti, line 12: :scissors:

psd-web/app/models/user.rb
```diff
@@ -148,7 +149,7 @@ def invitation_expired?
     invited_at <= INVITATION_EXPIRATION_DAYS.days.ago
   end
 
-  def two_factor_authentication_code_expired?
+  def secondary_authentication_code_expired?
```
scruti, line 152: :raising_hand:  Please change:
This has been moved to `SecondaryAuthentication`

psd-web/app/models/user.rb
```diff
@@ -158,25 +159,25 @@ def two_factor_locked?
     second_factor_attempts_locked_at && !two_factor_lock_expired?
   end
 
-  def fail_two_factor_authentication!
+  def fail_secondary_authentication!
```
scruti, line 162: :raising_hand:  Please change:
This has been moved to `SecondaryAuthentication`

psd-web/app/models/user.rb
```diff
@@ -158,25 +159,25 @@ def two_factor_locked?
...
     return if max_login_attempts?
 
     self.increment!(:second_factor_attempts_count, 1)
     lock_two_factor! if max_login_attempts?
   end
 
-  def pass_two_factor_authentication!
+  def pass_secondary_authentication!
```
scruti, line 169: :raising_hand:  Please change:
This seems to be moved to `SecondaryAuthentication`

psd-web/app/models/user.rb
```diff
@@ -158,25 +159,25 @@ def two_factor_locked?
...
     unlock_two_factor! if max_login_attempts?
     update_column(:second_factor_attempts_count, 0)
   end
 
   # BEGIN: place devise overriden method calls bellow
-  def send_two_factor_authentication_code(code)
-    SendTwoFactorAuthenticationJob.perform_later(self, code)
+  def send_secondary_authentication_code(code)
```
scruti, line 175: :raising_hand:  Please change:
This seems to be moved to `SecondaryAuthentication`

psd-web/app/models/user.rb
```diff
@@ -158,25 +159,25 @@ def two_factor_locked?
...
+    SendSecondaryAuthenticationJob.perform_later(self, code)
   end
 
-  def need_two_factor_authentication?(_request)
-    Rails.configuration.two_factor_authentication_enabled
+  def need_secondary_authentication?(_request)
```
scruti, line 179: :raising_hand:  Please change:
This is not globally checked anymore for the user right?

psd-web/config/initializers/devise.rb
```diff
@@ -302,13 +302,17 @@
...
-  # Devise two_factor_authentication gem
-  config.max_login_attempts = 10  # Maximum second factor attempts count.
-  config.direct_otp_valid_for = 5.minutes  # Time before direct OTP becomes invalid
-  config.direct_otp_length = 5  # Direct OTP code length
-  config.remember_otp_session_for_seconds = 7.days  # Time before browser has to perform 2fA again. Default is 0.
-  config.second_factor_resource_id = 'id' # Field or method name used to set value for 2fA remember cookie
-  config.delete_cookie_on_logout = false # Delete cookie when user signs out, to force 2fA again on login
+
+  # to prevent devise to destroying entire session object
+  # need to be false as secondary authentication relies on session
```
scruti, line 307: :clap: very informative

psd-web/config/routes.rb
```diff
@@ -30,6 +30,7 @@
   end
 
   resource :password_changed, controller: "users/password_changed", only: :show, path: "password-changed"
+  resource :secondary_authentications, only: %i[new create], path: 'two-factor'
```
scruti, line 33:
```suggestion
  resource :secondary_authentication, only: %i[new create], path: 'two-factor'
```

psd-web/db/schema.rb
```diff
@@ -167,6 +167,7 @@
     t.string "pretty_id"
     t.string "product_category"
     t.string "received_type"
+    t.tsvector "search_index"
```
scruti, line 170: A wild postgres extension appeared!

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -58,6 +58,9 @@
 
       click_link "Continue"
 
+
+      # bug or feature? requires second otp for login, as only 2FA was for reset password
```
scruti, line 62: Guess we need to clarify this with @edwardhorsford /Laurence.

psd-web/spec/features/registration_spec.rb
```diff
@@ -51,16 +54,20 @@ def fill_in_registration_form
...
     click_on "Continue"
   end
 
   def expect_to_be_on_declaration_page
     expect(page).to have_title("Declaration - Product safety database - GOV.UK")
   end
+
+  def otp_code
+    SecondaryAuthentication.last.direct_otp
+  end
```
scruti, line 72: Suggestion: 
Extract this to a feature helper? (as seems used everywhere where we test 2FA)

psd-web/spec/models/secondary_authentication_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+  let(:secondary_authentication) { create(:secondary_authentication, attempts: attempts, direct_otp_sent_at: direct_otp_sent_at) }
+
+  # rubocop:disable Style/MethodCalledOnDoEndBlock
+  it "increase attempts when checking code" do
+    expect do
+      secondary_authentication.valid_otp? "123"
+    end.to change { secondary_authentication.reload.attempts }.from(0).to(1)
+  end
+  # rubocop:enable Style/MethodCalledOnDoEndBlock
+end
```
scruti, line 15: Not sure what is the policy in this team about Unit testing and if it is ok to disregard them in favour of extensive feature testing, but this class seems to be missing a lot of unit tests.

psd-web/spec/requests/secondary_authentication_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+RSpec.describe "Secondary Authentication submit", :with_stubbed_notify, type: :request do
+  include ActiveSupport::Testing::TimeHelpers
+
+  let(:attempts) { 0 }
+  let(:direct_otp_sent_at) { Time.new.utc }
+  let(:secondary_authentication) { create(:secondary_authentication, user: user, attempts: attempts, direct_otp_sent_at: direct_otp_sent_at) }
+  let(:submitted_code) { secondary_authentication.direct_otp }
+  let(:max_attempts) { SecondaryAuthentication::MAX_ATTEMPTS }
+
+  describe "the form" do
```
scruti, line 12: :cut_of_meat: ? Adds an extra nested block and does not give any info over the parent block "Secondary Authentication submit"

psd-web/spec/requests/secondary_authentication_spec.rb
```diff
@@ -0,0 +1,94 @@
+require "rails_helper"
+
+RSpec.describe "Secondary Authentication submit", :with_stubbed_notify, type: :request do
+  include ActiveSupport::Testing::TimeHelpers
```
scruti, line 4: :scissors: ? We do not use any of the time helpers right?

psd-web/spec/support/sms.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class TestOTPSMS
```
scruti, line 3: Nice helper :clap:

psd-web/test/fixtures/users.yml
```diff
@@ -33,15 +36,17 @@ southampton_frankie:
...
   password_salt: 123
   encrypted_password: <%= Devise::Encryptable::Encryptors::PBKDF2.digest("2538fhdkvuULE36f", 27500, "123", nil) %>
+  mobile_number_verified: true
 
 opss:
   email: opss@example.com
   has_accepted_declaration: true
   has_viewed_introduction: true
-  name: Slavosh
+  name: Slawosz
```
scruti, line 45: :rofl:

slorek:
Some feedback from testing:

* Code should be 5 digits
* When the user is logged in and clicks 'Invite a team member' they seem to have unlimited attempts to enter the 2FA code. I tried 15 times. I expect the user to be signed out and the account to be locked as it would be if I tried to do the same thing when logging in, but we should agree what happens in this case.
* If the user has signed in and and entered a FA code within the last X minutes that should be remembered when clicking 'Invite a team member'. Having a single `last_time_otp_verified` timestamp against the session or user objects would allow us to check this and share this across different actions. Having separate authentications against each type of action means that the user could be asked repeatedly to do 2FA. I do think it's valid and useful to have different time thresholds for different actions.

slorek:
I think this is very nice work and a massive improvement over the existing Devise 2FA plugin code, thank you! 👏 

For me I think there is just a question of whether we actually need the extra `SecondaryAuthentication` model. I think we can probably get away with simply having a `otp_code` and `last_otp_verified_time` column on `User`. I don't think need to track different codes for different actions.

The action would then simply need to check that the `current_user.last_otp_verified_time` was within the threshold that it needs.

This would reduce the number of times users need to enter codes, and also remove confusion about different codes being tied to different actions. For example, if the user had two tabs open and had to provide an OTP code for each tab, which code belongs to which? With one code per account it would just always be the most recent code. I think this is sufficient for our needs.

I'm happy to open a PR against this to simplify it if that helps explain things.

psd-web/db/schema.rb
slawosz: Its from other branch.

slawosz:
@slorek Stupid github does not allow direct comment replies.
I am happy to accept PR (I was actually waiting for on).  Last OTP verification should be per session, not per user, as in odd cases we might have one user using 2 devices, or, more important, as security consideration. I designed this around multiple `SecondaryAuthentication` records, as it gives little complication, and code fits domain. 
So in general, to get rid of `SecondaryAuthentication`, we need to have following columns in `User`: `direct_otp`, `direct_otp_expires_at`, `direct_otp_attempts`, `direct_otp_requested_for`. Last column is needed because we sometimes customise code entry page. In session we need to have a key `:secondary_authentication_performed_at` which will store last time secondary authentication is performed. Should work. Its in `review` for long time, but if we want to simplify, lets do this first thing tomorrow. 
Regarding other comments, changing to simple model will invalidate some of them and I will fix other issues.

psd-web/spec/models/secondary_authentication_spec.rb
slawosz: Most of stuff is tested in different places, and I was still expecting some PR for storing authentication info. You know, :balance_scale:

psd-web/app/controllers/concerns/secondary_authentication_concern.rb
slawosz: @scruti When doing suggestion, please also change usages.

psd-web/app/forms/secondary_authentication_form.rb
slawosz: Thanks :)

psd-web/app/models/secondary_authentication.rb
```diff
@@ -0,0 +1,89 @@
+class SecondaryAuthentication
```
scruti, line 1: Suggestion? (more a thought).
Given that now this model is basically a `user.x` collection of calls... that may indicate the logic  behind it effectively belongs to `User`. I personally would consider to turn it into a concern and include the concern in the User.

But not a biggie at all.

psd-web/app/models/secondary_authentication.rb
```diff
@@ -0,0 +1,89 @@
...
+  end
+
+  def send_secondary_authentication_code
+    SendSecondaryAuthenticationJob.perform_later(user, user.direct_otp)
+  end
+
+  # def expiry_seconds
+  #   TIMEOUTS[user.secondary_authentication_operation]
+  # end
+
```
scruti, line 65:
```suggestion
```

psd-web/app/controllers/application_controller.rb
```diff
@@ -3,10 +3,14 @@ class ApplicationController < ActionController::Base
   include CacheConcern
   include HttpAuthConcern
   include RavenConfigurationConcern
+  include SecondaryAuthenticationConcern
 
   protect_from_forgery with: :exception
   before_action :authenticate_user!
   before_action :set_current_user
+  before_action :ensure_secondary_authentication
+  before_action :cleanup_secondary_authentication
```
scruti, line 12:
```suggestion
```

psd-web/app/controllers/concerns/secondary_authentication_concern.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  # Use as `before_filter` in application_controller controller
+  def ensure_secondary_authentication
+    session[:last_secondary_authentication_performed_at] = {}
+  end
+
+  def cleanup_secondary_authentication
+    # TODO: delete
+  end
+
```
scruti, line 35:
```suggestion
```

psd-web/db/migrate/20200412181146_create_secondary_authentications.rb
```diff
@@ -0,0 +1,14 @@
+class CreateSecondaryAuthentications < ActiveRecord::Migration[5.2]
```
scruti, line 1: Please change: Remove this migration. Not used anymore.

psd-web/db/schema.rb
```diff
@@ -215,6 +215,18 @@
...
+    t.boolean "authenticated"
+    t.datetime "authentication_expires_at"
+    t.datetime "created_at", null: false
+    t.string "direct_otp"
+    t.datetime "direct_otp_sent_at"
+    t.string "operation"
+    t.datetime "updated_at", null: false
+    t.string "user_id"
+  end
+
```
scruti, line 229: Please change: Needs to be removed.

psd-web/app/models/secondary_authentication.rb
slawosz: It prevents us from having overbloated user model. Concern method is still technically a user method and IHMO does not give distinction. Sure, `SecondaryAuthentication` is very connected to user, but it is separate thing and does not meet test `is_a? User`

psd-web/app/models/secondary_authentication.rb
```diff
@@ -0,0 +1,85 @@
...
+      secondary_authentication_operation: operation
+    )
+  end
+
+  def send_secondary_authentication_code
+    SendSecondaryAuthenticationJob.perform_later(user, user.direct_otp)
+  end
+
+  def try_to_verify_user_mobile_number
+    user.update(mobile_number_verified: true) unless user.mobile_number_verified
```
StupidCodeFactory, line 63: you'd want to either:
* handle the case where this isn't save and return false
or 
use `update!` so it raises and error and we know something wring happened.
or 
you can use 
`update_column(mobile_number_verified: true)` which will directly update the field in the DB not running any validation, hence safer. I'd go with this option.

psd-web/app/models/secondary_authentication.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  def try_to_unlock_secondary_authentication
+    if user.second_factor_attempts_locked_at && (user.second_factor_attempts_locked_at + MAX_ATTEMPTS_COOLDOWN.seconds) < Time.now.utc
+      user.update(second_factor_attempts_locked_at: nil)
```
StupidCodeFactory, line 68: same comment as above

psd-web/app/models/secondary_authentication.rb
StupidCodeFactory: If this fails silently it will be a nightmare to debug.

psd-web/app/models/secondary_authentication.rb
```diff
@@ -0,0 +1,85 @@
...
+    user.increment!(:second_factor_attempts_count) unless otp_locked?
+
+    if user.second_factor_attempts_count > MAX_ATTEMPTS
+      user.update(second_factor_attempts_locked_at: Time.now.utc, second_factor_attempts_count: 0)
+    end
+    user.reload.second_factor_attempts_locked_at.nil? && otp == user.direct_otp
+  end
+
+  def generate_code(operation)
+    user.update(
```
StupidCodeFactory, line 50: here too if it fails what happens?

psd-web/app/models/secondary_authentication.rb
```diff
@@ -0,0 +1,85 @@
...
+    user.second_factor_attempts_locked_at.present? && (user.second_factor_attempts_locked_at + MAX_ATTEMPTS_COOLDOWN.seconds) > Time.now.utc
+  end
+
+  def valid_otp?(otp)
+    try_to_unlock_secondary_authentication
+
+    user.increment!(:second_factor_attempts_count) unless otp_locked?
+
+    if user.second_factor_attempts_count > MAX_ATTEMPTS
+      user.update(second_factor_attempts_locked_at: Time.now.utc, second_factor_attempts_count: 0)
```
StupidCodeFactory, line 44: here as well. What happens when this fails to save?

psd-web/app/models/secondary_authentication.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  def otp_expired?
+    user.direct_otp_sent_at && (user.direct_otp_sent_at + OTP_EXPIRY_SECONDS) < Time.now.utc
+  end
+
+  def otp_locked?
+    user.second_factor_attempts_locked_at.present? && (user.second_factor_attempts_locked_at + MAX_ATTEMPTS_COOLDOWN.seconds) > Time.now.utc
+  end
+
+  def valid_otp?(otp)
```
StupidCodeFactory, line 38: you might want to wrap this in a transaction or/and use [pessimistic locking](https://api.rubyonrails.org/classes/ActiveRecord/Locking/Pessimistic.html). 

line 41 or  43 could fail independently (user not being valid due to a bug elsewhere in the code), leaving the user in a non desirable state.

psd-web/app/models/secondary_authentication.rb
```diff
@@ -0,0 +1,87 @@
...
+    try_to_unlock_secondary_authentication
+
+    user.with_lock do
+      user.increment!(:second_factor_attempts_count) unless otp_locked?
+
+      if user.second_factor_attempts_count > MAX_ATTEMPTS
+        user.update!(second_factor_attempts_locked_at: Time.now.utc, second_factor_attempts_count: 0)
+      end
+    end
```
StupidCodeFactory, line 47: You re-write this as:

```ruby
def increment_or_lock
  user.increment!(:second_factor_attempts_count) unless otp_locked?

  if user.second_factor_attempts_count > MAX_ATTEMPTS
    user.update!(second_factor_attempts_locked_at: Time.now.utc, second_factor_attempts_count: 0)
  end
end

# and in this method
user.with_lock(&:increment_or_lock)
```

StupidCodeFactory: that should quiet code climate and make this method nicer to read

PR #483, opened 2020-04-14
Add healthcheck for Investigations in DB

scruti:
We want the health-check to fail when the database does not contain
any investigation.
This may alert us when, for example, after a deploy to production the 
database connection is pointing to the wrong instance.


scruti:
[Add healthcheck count for investigations](https://trello.com/c/w3s0u7rl/403-add-healthcheck-count-for-investigations)

PR #487, opened 2020-04-15
Allow users to change their names

frankieroberto:
This adds a "Your account" screen, and allows users to change their name.

Trello card: https://trello.com/c/FGYrtWv4/332-3-your-account-pages

## Implementation notes

This was mostly pretty trivial. I've added two controllers, `AccountController` (which renders the account page put doesn't do anything else) and `AccountNameController`, which has `show` and `update` actions. These are named to match the URL paths.

There's one validation on the `name` attribute (a simple `presence` validation), which is also shared by the account setup flow. It seemed simpler to duplicate it though, rather than extract this into a shared concern/method/whatever.

The account page uses a new [`govukSummaryList`](https://github.com/UKGovernmentBEIS/govuk-design-system-rails/blob/master/app/helpers/govuk_design_system/summary_list_helper.rb) helper from the shared `govuk-design-system-rails` gem, which is designed to use the same arguments as the [nunjucks version](https://design-system.service.gov.uk/components/summary-list/). This outputs a definition list (`<dl>`) rather than a table, as I think this has proved to be more accessible.

### Your account page

Screenshot 2020-04-15 at 16 01 02

### Change name page

Screenshot 2020-04-15 at 16 01 07

### Change name page (with validation error)

Screenshot 2020-04-15 at 16 01 12

## Checklist:

### DIFF
--- a/psd-web/app/controllers/account_name_controller.rb
+++ b/psd-web/app/controllers/account_name_controller.rb
@@ -1,0 +1,18 @@
+class AccountNameController < ApplicationController
+  def show
+    @user = current_user
+    render "account/name"
+  end
+
+  def update
+    @user = current_user
+
+    @user.name = params[:user][:name]
+
+    if @user.save(context: :change_name)
+      redirect_to account_path
+    else
+      render "account/name"
+    end
+  end
+end
--- a/psd-web/app/controllers/account_name_controller.rb
+++ b/psd-web/app/controllers/account_name_controller.rb
@@ -7,7 +7,7 @@ class AccountNameController < ApplicationController
   def update
     @user = current_user
 
-    @user.name = params[:user][:name]
+    @user.name = params.dig(:user, :name)
 
     if @user.save(context: :change_name)
       redirect_to account_path
--- a/psd-web/app/views/account/name.html.erb
+++ b/psd-web/app/views/account/name.html.erb
@@ -1,0 +1,31 @@
+<% title = "Change your name" %>
+<% page_title title, errors: false %>
+
+<% content_for(:after_header) do %>
+  <%= govukBackLink(text: "Back", href: account_path) %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+
+    <%= error_summary(@user.errors) %>
+
+    <h1 class="govuk-heading-l"><%= title %></h1>
+
+    <%= form_with model: @user, url: account_name_path, method: :patch do |form| %>
+
+      <%= render "form_components/govuk_input", key: :name, form: form,
+        classes: "app-!-max-width-two-thirds",
+        label: {
+          text: "Full name"
+        },
+        value: @user.name,
+        autocomplete: "name",
+        id: "name"
+        %>
+
+      <%= govukButton(text: "Save") %>
+
+    <% end %>
+  </div>
+</div>
--- a/psd-web/app/views/account/name.html.erb
+++ b/psd-web/app/views/account/name.html.erb
@@ -7,14 +7,15 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds-from-desktop">
-
     <%= error_summary(@user.errors) %>
 
     <h1 class="govuk-heading-l"><%= title %></h1>
 
     <%= form_with model: @user, url: account_name_path, method: :patch do |form| %>
 
-      <%= render "form_components/govuk_input", key: :name, form: form,
+      <%= render "form_components/govuk_input",
+        key: :name,
+        form: form,
         classes: "app-!-max-width-two-thirds",
         label: {
           text: "Full name"
@@ -25,7 +26,6 @@
         %>
 
       <%= govukButton(text: "Save") %>
-
     <% end %>
   </div>
 </div>
--- a/psd-web/app/views/account/name.html.erb
+++ b/psd-web/app/views/account/name.html.erb
@@ -1,5 +1,5 @@
 <% title = "Change your name" %>
-<% page_title title, errors: false %>
+<% page_title title, errors: @user.errors.any? %>
 
 <% content_for(:after_header) do %>
   <%= govukBackLink(text: "Back", href: account_path) %>
--- a/psd-web/app/views/account/show.html.erb
+++ b/psd-web/app/views/account/show.html.erb
@@ -5,7 +5,7 @@
 
     <h1 class="govuk-heading-l"><%= title %></h1>
 
-    <%= govukSummaryList(rows: [
+    <%= govukSummaryList(classes: "app-summary-list--key-weight-regular", rows: [
       {
         key: {text: "Name"},
         value: {text: current_user.name },
--- a/psd-web/app/views/account/show.html.erb
+++ b/psd-web/app/views/account/show.html.erb
@@ -45,5 +45,9 @@
         }
       ]
     ) %>
+
+    <div class="govuk-inset-text">
+      If you need to change your details, email <%= mail_to t(:enquiries_email), nil, class: "govuk-link" %>
+    </div>
   </div>
 </div>
--- a/psd-web/app/views/account/show.html.erb
+++ b/psd-web/app/views/account/show.html.erb
@@ -36,7 +36,7 @@
         {
           # TODO: simplify this once we’ve removed the ability for users to belong
           # to multiple teams.
-          key: { text: "Your #{current_user.teams.size == 1 ? 'team' : 'teams'}" },
+          key: { text: "Your #{"team".pluralize(current_user.teams.size)}" },
           value: {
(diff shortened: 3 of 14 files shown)


psd-web/app/views/account/show.html.erb
```diff
@@ -0,0 +1,49 @@
...
+          key: { text: "Email address" },
+          value: { text: current_user.email }
+        },
+        {
+          key: { text: "Mobile number" },
+          value: { text: current_user.mobile_number }
+        },
+        {
+          key: { text: "Password" },
+          value: { text: "********" }
```
frankieroberto, line 34: @edwardhorsford any thoughts what this should say, given that we can’t currently say when the password was last changed? Alternatively, we could drop this row for now (until we add the "Change password" feature).

psd-web/app/controllers/account_name_controller.rb
```diff
@@ -0,0 +1,18 @@
+class AccountNameController < ApplicationController
```
frankieroberto, line 1: Did consider calling this `NameController`, but thought that might be a bit ambiguous ("who’s name?")

psd-web/app/views/account/name.html.erb
```diff
@@ -0,0 +1,31 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+    <%= error_summary(@user.errors) %>
+
+    <h1 class="govuk-heading-l"><%= title %></h1>
+
+    <%= form_with model: @user, url: account_name_path, method: :patch do |form| %>
+
+      <%= render "form_components/govuk_input",
```
slawosz, line 16: I wonder if we could start use more compact form builders, like `simple_form` gem for example.

psd-web/spec/support/matchers/capybara_matchers.rb
```diff
@@ -1,19 +1,29 @@
...
     def matches?(page)
-      page.find("th", text: @key, exact_text: true).sibling("td", text: @value, exact_text: true)
+      key_element = @table ? "th" : "dt"
+      value_element = @table ? "td" : "dd"
+
+      page.find(key_element, text: @key, exact_text: true).sibling(value_element, text: @value, exact_text: true)
     end
   end
 
+  # Deprecated: remove once all summary lists have been switched to definition list markup
```
StupidCodeFactory, line 22: Can you remove the comment, and `Rails.logger.warn "my comment that is more useful if displayed in the logs so that developer can address it in due time?"`

psd-web/spec/features/your_account_spec.rb
```diff
@@ -0,0 +1,54 @@
+require "rails_helper"
+
+RSpec.feature "Your Account", :with_stubbed_elasticsearch, :with_stubbed_mailer, type: :feature do
+  let(:user) {
+    create(:user,
+           :activated,
+           :opss_user,
+           name: "Joe Jones",
+           email: "joe.jones@testing.gov.uk",
+           mobile_number: "07700 900000",
```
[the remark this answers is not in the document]
StupidCodeFactory: I think @edwardhorsford has a point but better than a comment no one ever pays attention to is to use https://github.com/faker-ruby/faker/blob/master/doc/default/phone_number.md

psd-web/app/views/account/show.html.erb
```diff
@@ -0,0 +1,49 @@
...
+        },
+        {
+          # TODO: simplify this once we’ve removed the ability for users to belong
+          # to multiple teams.
+          key: { text: "Your #{current_user.teams.size == 1 ? 'team' : 'teams'}" },
```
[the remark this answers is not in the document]
frankieroberto: @StupidCodeFactory this doesn't do the same thing, as it'd output "Your 1 team" or "Your 1 teams".

We could use this, but I don't think it’s particularly cleaner:

```ruby
key: { text: "Your #{current_user.teams.size == 1 ? 'team' : 'teams'}" },

File: psd-web/app/controllers/account_name_controller.rb (hunk 2 above)
Comment (frankieroberto) [reply psd-web/app/controllers/account_name_controller.rb:1]: I'm not convinced by module namespacing, as there isn't any shared behaviour at the account level (yet), but maybe it's something we could consider in future. 👍

File: psd-web/app/controllers/account_name_controller.rb [hunk 7]
```diff
@@ -0,0 +1,18 @@
...
+  def show
+    @user = current_user
+    render "account/name"
+  end
+
+  def update
+    @user = current_user
+
+    @user.name = params[:user][:name]
```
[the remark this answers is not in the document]
frankieroberto: I've switched this to `params.dig(:user, :name)` which will render a validation error (complaining about missing name) if this ever happens (which it shouldn’t).

Although thinking about it, perhaps it _should_ render a 500 error, as if the param was missing, that'd indicate a problem with the form... 🤔

psd-web/app/views/account/show.html.erb
frankieroberto: 👍

psd-web/app/views/account/name.html.erb
frankieroberto: We should have a separate chat about this. Right now we've inherited all these partials, but I'd like to find a way to wrap these within a more Rails-y API. I think we'd still want to support the same params and markup output as the Nunjucks macros (eg [`govukInput`](https://design-system.service.gov.uk/components/text-input/) though (as the Design System team have done a lot of work on improving accessibility of the HTML).

eg it'd be nice to be able to do something like (specifying `label` and `hint` together with the attribute):

```ruby
<%= form.input :name, label: "Full name", hint: "blah", autocomplete: "name" %>
```

psd-web/spec/support/matchers/capybara_matchers.rb
frankieroberto: Done, although I've left a brief `TODO` comment in there too.

psd-web/spec/features/your_account_spec.rb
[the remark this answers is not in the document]
frankieroberto: I quite like setting an explicit phone number string, so that the test can repeat that later in the expectation, rather than looking it up from the model.

psd-web/app/views/account/show.html.erb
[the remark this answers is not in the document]
StupidCodeFactory: sorry I mean to suggests `"teams".pluralize(current_user.teams.size)`

psd-web/app/controllers/account_name_controller.rb
```diff
@@ -0,0 +1,18 @@
+class AccountNameController < ApplicationController
+  def show
+    @user = current_user
+    render "account/name"
+  end
+
+  def update
+    @user = current_user
+
+    @user.name = params.dig(:user, :name)
```
StupidCodeFactory, line 10: I still think this is wrong this controller needs a `user[name]` params to work correctly. This now will fail silently. 
what's wrong with strong parameter that's been conceived for this purpose.

psd-web/app/views/account/name.html.erb
StupidCodeFactory: that would be ideal and the best compromise between both world. But I'm afraid we'll never get to it unless @slawosz slave's during the bank holiday weekend :)

psd-web/app/controllers/account_name_controller.rb
```diff
@@ -0,0 +1,18 @@
+class AccountNameController < ApplicationController
+  def show
+    @user = current_user
+    render "account/name"
```
StupidCodeFactory, line 4: If not name-spacing this controller than please follow rails convention: this view should then in the `account_name` folder to stick with rails convention and then you can just do `render :show`

psd-web/app/views/account/show.html.erb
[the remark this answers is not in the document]
frankieroberto: 👍

psd-web/app/views/account/name.html.erb
frankieroberto: I'm trying to edge my way there, but I think it's easiest to start by emulating the nunjucks macros as helpers (as it's then easy to copy the nunjucks logic and output), and then wrap those up in Rails form helpers. Will take a while though.

psd-web/app/controllers/account_name_controller.rb
frankieroberto: strong parameter is useful when you're bulk updating from params (eg `.update_attributes(params)`) and want to whitelist which ones are ok. As it's only one attribute here, I think it's ok to pluck it straight out the params hash (if present).

psd-web/app/controllers/account_name_controller.rb
frankieroberto: I'm going to leave as-is for now. I renamed the view to group it with the account page and make it match the URL path.

PR #488, opened 2020-04-15
add tweaks

StupidCodeFactory:
Add some tweaks in the coronavirus form so we don't depend in the model instance to show errors and can simplify the controller code.

### DIFF
--- a/psd-web/app/controllers/investigations/project_controller.rb
+++ b/psd-web/app/controllers/investigations/project_controller.rb
@@ -4,6 +4,7 @@ class Investigations::ProjectController < ApplicationController
   steps :coronavirus, :project_details
 
   before_action :set_investigation, only: %i[show new create update]
+  before_action :set_new_coronavirus_form, only: :show, if: -> { step == :coronavirus }
 
   #GET /xxx/step
   def show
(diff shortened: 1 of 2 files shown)


psd-web/app/views/investigations/coronavirus.html.erb
```diff
@@ -9,11 +9,11 @@
 <%= form_with scope: scope, model: @coronavirus_related_form, url: wizard_path, method: :put, local: true do |form| %>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds-from-desktop">
-      <%= error_summary @investigation.errors  %>
+      <%= error_summary @coronavirus_related_form.errors  %>
       <% if unsafe_product_message %>
         <span class="govuk-caption-l">Report an unsafe or non-compliant product</span>
       <% end %>
-
+      <%= form.hidden_field :coronavirus_related %>
```
StupidCodeFactory, line 16: this is important: it ensures the params is always there when the user submits the for without choosing any option!

psd-web/app/controllers/investigations/project_controller.rb
```diff
@@ -60,12 +61,17 @@ def set_investigation
     @investigation = Investigation::Project.new(investigation_params).decorate
   end
 
+  def set_new_coronavirus_form
+    @coronavirus_related_form = CoronavirusRelatedForm.new
+  end
+
   def coronavirus_related_form
-    @coronavirus_related_form ||= CoronavirusRelatedForm.new(investigation_params)
+    @coronavirus_related_form ||= CoronavirusRelatedForm.new(params.require(:investigation).permit(:coronavirus_related))
```
StupidCodeFactory, line 69: this form does not care about the other investigation attributes!!

scruti: Oh yeah, got an error on this when moving it to other controllers :+1:

PR #490, opened 2020-04-16
Strip whitespace from security code that user enters

frankieroberto:
This is helpful, as some users may accidentally add a space after the security code (eg from a mobile keyboard).

See https://trello.com/c/RWJ7pENv/492-1-strip-whitespace-on-2fa-codes-before-validating

### DIFF
--- a/psd-web/app/controllers/users/two_factor_authentication_controller.rb
+++ b/psd-web/app/controllers/users/two_factor_authentication_controller.rb
@@ -46,7 +46,7 @@ module Users
   private
 
     def otp_code_param
-      @otp_code_param ||= resource_params.permit(:otp_code)[:otp_code]
+      @otp_code_param ||= resource_params.permit(:otp_code)[:otp_code].strip
     end
 
     # Suppress unwanted Devise alert.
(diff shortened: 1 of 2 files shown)


psd-web/app/controllers/users/two_factor_authentication_controller.rb
```diff
@@ -46,7 +46,7 @@ def secondary_nav_items
   private
 
     def otp_code_param
-      @otp_code_param ||= resource_params.permit(:otp_code)[:otp_code]
+      @otp_code_param ||= resource_params.permit(:otp_code)[:otp_code].strip
```
scruti, line 49: We will get an ` { otp_code: ""}` even when not filling the form, right? 
Just trying to figure out if there is any chance of the parameter being nil and raising an exception on the strip.

frankieroberto: Hmm, I guess this would happen if we change the input name in the form, or if someone 'fakes' the page and sends an HTTP request without the param. But just not filling in the input field should work, as an empty string will be sent.

PR #500, opened 2020-04-17
Add teams to a case

frankieroberto:
Adds the ability for members of the team assigned to the case to add other teams to the case. 

Members of all teams added to the case can view all details of the case, even if it is otherwise restricted.

https://trello.com/c/aYLARnFZ/431-add-teams-to-a-case

## Implementation notes

This adds a new model and database table which joins cases (investigations) to teams, which I've called "collaborators" (to match the UI). For the time being, this is additional to the existing "assignee" relationship, which uses a polymorphic association from the Investigation model, allowing the assignee to be either a team or a user.

The existing Pundit-based permission for "can a user see this case" is updated to also check for whether the user is a member of one of the teams added as a collaborator, in addition to the existing checks based upon assignee organisation.

As well as adding a team to the case via the collaborator model, the code also needs to send the added team an email, and add an audit entry. To encapsulate these additional actions, I abstracted these out into an `AddTeamToAnInvestigation` service object.  This is a deviation, but hopefully an improvement, on the existing pattern of adding audit entries via activerecord callbacks on the models.

## Checklist:

### General testing (author)
- [x] Test without JavaScript
- [x] Test on small screen

- [x] Works keyboard only


edwardhorsford:
Had a look, looking good so far!

Some things I've spotted:
1. When there's an error on the teams page it loads at `/teams` rather than `/teams/add`. It bothers me that we're loading the same page at two entirely different urls. I'm aware this is a "rails way" / strict "RESTful" way thing, but it feels weird that the url changes and no longer matches what you're doing.
2. Content on teams page missing some bits - see story for full content
3. Can we limit the length of the teams autocomplete? suggest 2/3 to start.
4. As we don't have a teams list we go back to, can we show a success flash message on the overview  page - I'll write you some content. (would possibly remove this in the future, but it matches what inviting people does)
5. Error messages not quite right.
6. Can turn on showAllValues on the autocomplete?

frankieroberto:
@edwardhorsford I've added a flash message, the missing content, and the "show all values" option for the autocomplete.

What do you think the error messages should all say? See [existing values here](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/6a5e0c2af21f14b18ad4e90025690a83a2ebdfb2/psd-web/config/locales/en.yml#L138-L144).

psd-web/app/models/collaborator.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  attr_reader :include_message
+
+  def include_message=value
+    @include_message = if value.is_a? String
+                         (value == "true")
+                       else
+                         value
+                       end
+  end
```
frankieroberto, line 19: Rails coerces strings to booleans automatically for attribute setters backed by a boolean database column, but as this is a custom setter it needs to be done manually.

psd-web/app/models/investigation.rb
```diff
@@ -60,6 +63,18 @@ def assignee
     end
   end
 
+  def assignee_team
+    if assignable_type == "Team"
+      assignee
+    elsif assignable_type == "User"
+      assignee.teams.first
```
frankieroberto, line 70: Returning the 'first' team is a bit random, but in practice users _should_ only belong to one team, and we’re planning on enforcing this soon.

psd-web/app/policies/investigation_policy.rb
```diff
@@ -37,6 +37,9 @@ def visible_to(user:, private: @record.is_private)
     return true if @record.assignee.present? && (@record.assignee&.organisation == user.organisation)
     return true if @record.source&.user_has_gdpr_access?(user: user)
 
+    # Have any of the user’s teams been added to the case as a collaborator?
+    return true if (@record.teams & user.teams).any?
```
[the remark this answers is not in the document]
frankieroberto: Codeclimate is complaining about the number of return statements here, but I don’t think this is worth refactoring _yet_ as we'll likely review this very soon in separate stories.

psd-web/app/controllers/collaborators_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def create
+    @investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
+
+    authorize @investigation, :add_collaborators?
+
+    message = params.dig(:collaborator, :message)
+
+    result = AddTeamToAnInvestigation.call(
+      team_id: params.dig(:collaborator, :team_id),
```
slawosz, line 20: Please change: use `strong parameters`

psd-web/app/services/add_team_to_an_investigation.rb
```diff
@@ -0,0 +1,32 @@
+class AddTeamToAnInvestigation
```
slawosz, line 1: I really don't see any sense with pretending all values here with `context`. Its ugly.
We still can agree on having unified interface by using `call` in services and implementing `success?` method, nothing stop us from doing this. Please look at the screen below and see how many times we are using context... @slorek 

Screenshot from 2020-04-23 10-15-02

psd-web/app/views/collaborators/new.html.erb
```diff
@@ -0,0 +1,84 @@
...
+
+    <%= form_with model: @collaborator, url: investigation_collaborators_path(@investigation), method: :post do |form| %>
+
+      <% team_select_items = [{text: "", value: ""}] + @teams.map { |team| { text: team.name, value: team.id, selected: (team.id == @collaborator.team_id) } } %>
+
+      <% if @collaborator.errors.include?(:team) %>
+        <% error_message = { text: @collaborator.errors.full_messages_for(:team).first } %>
+      <% end %>
+
+      <%= render "components/govuk_select",
```
slawosz, line 25: really, we should start thinking about using rails friendly forms. Hopefully, we could end up with this:
```
<%= form_for @collaborator do |f| %>
  <%= f.input :team_id, as: govuk_select %>
  <%= f.input :etc
<% end %>

File: psd-web/app/views/investigations/tabs/_overview.html.erb [hunk 7]
```diff
@@ -27,6 +27,48 @@
       <%= @investigation.source_details_summary_list %>
     <% end %>
 
+    <%= govuk_hr %>
+    <h2 class="govuk-heading-m">Sharing and collaboration</h2>
+
+    <% if policy(@investigation).add_collaborators? %>
```
slawosz, line 33: There is the potential of using decorators to make this view thinner. Consider this next time.

psd-web/db/migrate/20200417130015_create_collaborators.rb
```diff
@@ -0,0 +1,20 @@
...
+        t.text :message
+
+        t.timestamps
+      end
+
+      add_foreign_key :collaborators, :investigations
+      add_foreign_key :collaborators, :teams
+      add_foreign_key :collaborators, :users, column: :added_by_user_id
+
+      add_index :collaborators, %i[investigation_id team_id], unique: true
```
slawosz, line 17: Index order is important. If you will ever search by `team_id` alone, index won't work.

frankieroberto: Hard to know which order is more useful at this point... The index is mainly here to stop us adding the same team to a case more than once...

psd-web/config/locales/en.yml
```diff
@@ -134,6 +134,14 @@ en:
               blank: "Enter date of the test request"
               invalid: "Enter a real date of the test request"
               date_missing_component: "Enter date of the test request and include a %{missing_components}"
+        collaborator:
+          attributes:
+            team:
+              required: "Select a team"
```
edwardhorsford, line 140:
```suggestion
              required: "Select a team to add to the case"
```

psd-web/config/locales/en.yml
```diff
@@ -134,6 +134,14 @@ en:
...
+            message:
+              blank: Enter a message to the team
```
edwardhorsford, line 142:
```suggestion
              blank: Enter a message
```

psd-web/app/controllers/collaborators_controller.rb
frankieroberto: I don't see any compelling reason to use strong parameters here (we don't need them for security reasons as I'm already whitelisting the three parameters needed).

The main difference would be in the call to `AddTeamToAnInvestigation` which would become:

```ruby
AddTeamToAnInvestigation.call(
  params: params.require(:collaborator).permit(:team_id, :message, :include_mesage),
  investigation: @investigation,
  current_user: current_user
)
```

...which is fine, and slightly shorter, but seems to me to be making the service object _more specific_ to the controller (ie requiring some of the user-provided params in a hash), rather than more abstract, with all of the params named in the same way.

psd-web/app/views/collaborators/new.html.erb
frankieroberto: Yeah, I'd like us to get to something like:

```ruby
<%= form_for @collaborator do |form| %>
  <%= form.govuk_select :team_id, items: @teams, label: "Choose team" %>
<% end %>
```

...but that's for another day.

psd-web/app/views/investigations/tabs/_overview.html.erb
frankieroberto: 🤷‍♂️ I personally find the decorator pattern unhelpful, vs the relatively simple condition here.

psd-web/app/services/add_team_to_an_investigation.rb
frankieroberto: I'm not a huge fan of the `context` calls either, but we’ve agreed to use the `interactor` gem for the time being, and that’s the convention it uses. Let's review this again in a few months?

psd-web/spec/features/case_permissions_management_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    click_button "Add team to this case"
+
+    expect_to_be_on_case_page(case_id: investigation.pretty_id)
+
+    expect(page).to have_summary_item(key: "Teams added to case", value: "Southampton Trading Standards Portsmouth Trading Standards")
+
+
+    notification_email = delivered_emails.last
+
+    expect(notification_email.recipient).to eql("enquiries@southampton.gov.uk")
```
StupidCodeFactory, line 60: Neat pick:  we have been using `eq` everywhere since were started writing RSpec. The ruby community also vastly uses `eq`. You only really want to use `eql` to explicitly check two variable are equal without type conversion so that it is obvious to the reader.

psd-web/spec/models/investigation_spec.rb
```diff
@@ -0,0 +1,76 @@
+require "rails_helper"
+
+RSpec.describe Investigation, :with_stubbed_elasticsearch, :with_stubbed_mailer, :with_stubbed_notify do
+  describe "#teams_with_access" do
+    context "when there is no-one assigned" do
+      let(:investigation) { create(:investigation, assignee: nil) }
+
+      it "is an empty list" do
+        expect(investigation.teams_with_access).to eql([])
```
StupidCodeFactory, line 9:
```suggestion
        expect(investigation.teams_with_access).to be_empty
```

psd-web/spec/services/add_team_to_an_investigation_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        )
+      end
+      # rubocop:enable RSpec/ExampleLength
+
+      # rubocop:disable RSpec/MultipleExpectations
+      it "queues a job to notify the team", :with_test_queue_adapter do
+        expect { result }.to have_enqueued_job(NotifyTeamAddedToCaseJob).with do |collaborator|
+          expect(collaborator.team_id).to eql(team.id)
+        end
+      end
```
StupidCodeFactory, line 41: Opinion: I would merge all these together, at least the first two.

psd-web/spec/services/add_team_to_an_investigation_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      # rubocop:enable RSpec/MultipleExpectations
+
+      # rubocop:disable RSpec/MultipleExpectations
+      it "adds an activity audit record" do
+        result
+        last_added_activity = investigation.activities.order(:id).first
```
StupidCodeFactory, line 47: `order(:id)` is not doing what you thing it does. UUIDs are not sortable only alphabetically but it it not very useful

psd-web/spec/services/add_team_to_an_investigation_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+        expect(last_added_activity).to be_a(AuditActivity::Investigation::TeamAdded)
+        expect(last_added_activity.title).to eql("Testing team added to allegation")
+        expect(last_added_activity.source.user_id).to eql(user.id)
```
StupidCodeFactory, line 51: this could be written better:

`expect(last_added_activity).to have_attributes(type: "AuditActivity::Investigation::TeamAdded", title: "Testing team added to allegation", user_id: user.id)`

psd-web/spec/services/add_team_to_an_investigation_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        expect(result.collaborator).to have_attributes(
+          team: team,
+          added_by_user: user,
+          investigation: investigation,
+          message: message
+        )
+      end
+      # rubocop:enable RSpec/ExampleLength
+
+      # rubocop:disable RSpec/MultipleExpectations
```
StupidCodeFactory, line 36: Please change: add :agregate_failures metatag when you want multiple expectations per spec

psd-web/spec/support/mailer.rb
```diff
@@ -6,6 +6,8 @@ def initialize(recipient, personalization)
     @personalization = personalization
   end
 
+  attr_reader :personalization
```
StupidCodeFactory, line 9: please remove: this will not be use in production.

If you want to test the correct personalisation have been sent to notify use a webmock expectation

psd-web/spec/services/add_team_to_an_investigation_spec.rb
frankieroberto: activities don't use UUIDs, they use incrementing serial ids. That said, I think `activities.order(:created_at).last` would be better.

psd-web/app/views/collaborators/new.html.erb
slawosz: I would advise for using `simple_form` which would result with `f.input :field, as: :govuk_select`

psd-web/spec/services/add_team_to_an_investigation_spec.rb
StupidCodeFactory: ah true theses where are from before we added uuid! I haven't worked on a project with integer ids in probably more than 6 or 7 years...

psd-web/app/views/collaborators/new.html.erb
frankieroberto: One to investigate for sure.

psd-web/spec/support/mailer.rb
frankieroberto: I thought we'd decided not to use webmock for asserting Notify api calls (as we don't control that API or the gem which calls it), but instead to use this `TestNotifyEmail` class in `spec/support/mailer`?

psd-web/app/controllers/collaborators_controller.rb
slorek: I'm with @frankieroberto - I don't think there's much point moving this out of the controller if the service requires a `params` object. It should be agnostic as to where the input came from.

slorek: We can sanitise the `params` first before using `dig` though to give us the proper HTTP status codes.

psd-web/app/controllers/collaborators_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      include_message: params.dig(:collaborator, :include_message),
+      message: message,
+      investigation: @investigation,
+      current_user: current_user
+    )
+
+    if result.success?
+      flash[:success] = "#{result.collaborator.team.name} added to the case"
```
slorek, line 28: Suggestion: Use I18n for text strings.

psd-web/app/controllers/collaborators_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+private
+
+  def teams_without_access
+    Team.where.not(id: team_ids_with_access).order(:name)
+  end
+
+  def team_ids_with_access
+    @investigation.collaborators.pluck(:team_id) + [@investigation.assignee_team.try(:id)]
```
slorek, line 44: Shouldn't the `assignee_team` be part of the `collaborators` set? Assignee to me is just a type of collaborator.

psd-web/app/models/collaborator.rb
slorek: Yes this looks like a good case for a form object.

psd-web/app/controllers/collaborators_controller.rb
StupidCodeFactory: That's exactly what strong parameter does, it sanitises the input (#permit) and raise (#require) the appropriate error code that `ActionController` knows how to catch and return the appropriate HTTP status in this case 400. 
```ruby
def collaborator_params
  params.require(:collaborator).permit(:team_id, :message, :include_mesage)
end

AddTeamToAnInvestigation.call(collaborator_params: collaborator_params, ...)
```
This vastly simplify the code in Interactor as you can just as this to

```ruby
investigation.collaborators.new(context.collaborator_params)
```

Instead of digging individual values out the strong params to pass them individually to the interactor and then reconstruct yet another time the same structure to pass to the ActiveRecord constructor

psd-web/app/models/investigation.rb
slorek: Agree team is a property of `assignee` so it should be `assignee.team`.

psd-web/app/controllers/collaborators_controller.rb
slorek: Agreed, but I would splat the sanitised params to the service object so the are separate variables within the context of the service object. Then you can avoid the `dig` stuff.

StupidCodeFactory: It's just not very dry to dig things out, like this to go build the same on the other side.
I feel very strongly about this, We pass hash as argument everywhere we don't deconstruct them to reconstruct them after.

slorek:
```ruby
AddTeamToAnInvestigation.call(
  params.require(:collaborator).permit(:team_id, :message, :include_mesage).merge {
    investigation: @investigation,
    current_user: current_user
  }
)
```

psd-web/app/controllers/collaborators_controller.rb
frankieroberto: @slorek I think we should consider this as a future task. Currently investigations can have a single optional assignee which can be either a Team or a User (or be nil). I think the intention is to possibly move towards a case have a single "owner" which is always a team, and can't be nil - but let’s confirm this as a separate story.

psd-web/app/controllers/collaborators_controller.rb
frankieroberto: Good call, forgot this. 👍

psd-web/spec/features/case_permissions_management_spec.rb
frankieroberto: Done, thanks. Keep forgetting this.

psd-web/app/controllers/collaborators_controller.rb
frankieroberto: On the strong parameters front, I personally don’t think the the Rails default behaviour of throwing a 400 error if required parameters are missing is helpful – much better to render a page with some validation messages. But this is a bit academic as this should never happen in practice.

I like @slorek’s suggestion. It felt weird to me to separate be calling the service object with separate `collaborator_params` alongside other arguments not originating from the params hash, but this way they all get sent together. 👍 

The other potential inconsistency of this service object’s API, were it to be used from another context (eg a rake task or a job), is that the team is sent as an id (`team_id`) but the investigation is sent as an object reference (`investigation`). This was mainly because controller needed to raise a `RecordNotFound` exception early so that a 404 page can be rendered if you were to POST to a URL which doesn't exist.  I think we can probably live with this for now though, but as we add additional service objects, it might be worth trying to be consistent about passing objects vs ids...

psd-web/app/models/collaborator.rb
frankieroberto: `attribute :include_message, :boolean` doesn't seem to work.

It's definitely a bit of a hack adding this boolean attribute to the model. I did consider a form object, but it seemed like a fair bit of work (and an extra abstraction) for the sake of simply adding one extra checkbox as a guard question. If we feel strongly I can do it though.

I'm also curious as to how this'd work alongside the service object. It'd feel a bit weird for me to have the service object construct a form object, see if that's valid, and if so re-assign the collaborator to the investigation and save it, as this then really tightly couples it to the controller & form flow.  Alternatively, maybe the controller initialised the form, sees if it it's valid, if so re-assigns the collaborator to the investigation, and passes the (unsaved) collaborator to the service object.  That'd reduce the service object to this (in which case it should probably be renamed to something like `SaveCollaboratorAndNotify` which isn't _that_ different from an activerecord callback... (apart from being opt-in):

```ruby
class AddTeamToAnInvestigation
  delete :collaborator, to: :context
  def call
    begin
      if collaborator.save
        NotifyTeamAddedToCaseJob.perform_later(collaborator)

        AuditActivity::Investigation::TeamAdded.create!(
          source: UserSource.new(user: collaborator.added_by_user),
          investigation: collaborator.investigation,
          title: "#{collaborator.team.name} added to #{collaborator.investigation.case_type.downcase}",
          body: collaborator.message.to_s
        )
      else
        context.fail!
      end
    rescue ActiveRecord::RecordNotUnique
      # Collaborator already added, so return successful but without notfiying the team
      # or creating an audit log.
    end
  end
end
```

psd-web/app/models/investigation.rb
frankieroberto: But the assignee can _be_ a team, so it means adding a property of `Team` which just calls `self`, purely for the convenience of another method being able to delegate to it. Also users can currently have multiple teams, but we've agreed that in this context it’s safe to naively use the first one, as we know that multiple teams are rare/non-existent and we want to revisit this soon anyway.

The `assignee_team` method is definitely not ideal, but it’s intended to be short term, and it follows the same pattern as the [assignee getters and setters](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/models/investigation.rb#L54-L61) (no idea why these are overriding the rails defaults - perhaps just to allow for nils?).

Suggest we leave this for now and revisit as part of https://trello.com/c/uQgGqQDn/510-remove-ability-to-be-in-multiple-teams

nasirkhanbeis:
Tested this in review app, looking good as expected.

PR #510, opened 2020-04-21
Create organisations and seeds

StupidCodeFactory:
We need better seeds on PR review app that are a bit more realistic and works well with 2FA

This stores in a PaaS service a blurb of JSON that's ran whenever a new database is seeded.


psd-web/deploy-review.sh
```diff
@@ -23,6 +23,8 @@ until cf7 service $DB_NAME > /tmp/db_exists && grep -E "create succeeded|update
 
 cp -a ${PWD-.}/infrastructure/env/. ${PWD-.}/psd-web/env/
 
+cf bind-service $APP_NAME psd-seeds --binding-name psd-seeds
```
StupidCodeFactory, line 26: this is done here as opposed to inside the manifest because I need to provide a binding name to reliably find "credentials"  in the seed file https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/510/files#diff-ae78a7ee787b598a69bc8b08686388c3R418

psd-web/deploy-review.sh
```diff
@@ -23,6 +23,8 @@ until cf7 service $DB_NAME > /tmp/db_exists && grep -E "create succeeded|update
 
 cp -a ${PWD-.}/infrastructure/env/. ${PWD-.}/psd-web/env/
 
+cf7 bind-service $APP_NAME psd-seeds --binding-name psd-seeds
```
StupidCodeFactory, line 26: this is done here as opposed to inside the manifest because I need to provide a binding name to reliably find "credentials" in the seed file https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/510/files#diff-ae78a7ee787b598a69bc8b08686388c3R418

psd-web/db/seeds.rb
```diff
@@ -413,36 +413,39 @@ def create_blob(filename, title: nil, description: nil)
 
   investigation.products << product
 
-  organisation = Organisation.create!(name: "Office for Product Safety and Standards")
+  if ENV["SEED_USERS"]
+    # rubocop:disable Rails/DynamicFindBy
+    organisations = Rails.env.production? ? CF::App::Credentials.find_by_service_name("psd-seeds")["organisations"] : Rails.application.credentials.organisations
```
slorek, line 418: Please change: document how `Rails.application.credentials.organisations` is structured and where it is defined. Add example code if possible.

psd-web/db/seeds.rb
```diff
@@ -413,36 +413,73 @@ def create_blob(filename, title: nil, description: nil)
...
-
-  Team.create!(name: "OPSS Science and Tech", team_recipient_email: nil, "organisation": organisation)
-  Team.create!(name: "OPSS Trading Standards Co-ordination", team_recipient_email: nil, "organisation": organisation)
-  Team.create!(name: "OPSS Incident Management",  team_recipient_email: nil, "organisation": organisation)
-  Team.create!(name: "OPSS Testing", team_recipient_email: nil, "organisation": organisation)
-
-  # On review apps environment (int) this will be set using `psd-seed-users-env`
-  # example of variable:
-  # SEED_USERS='JohannMuster:muster@example.org:07777888999;JohnDoe:example@example.org:07777888999'
   if ENV["SEED_USERS"]
```
slorek, line 416: Do we still need `ENV["SEED_USERS"]` if users are defined in the JSON? What would this contain? Perhaps we should be checking if we are in `production` environment and `CF::App::Credentials.find_by_service_name("psd-seeds")` returns true instead.  What happens if this service is not defined (i.e. on staging and production)?

StupidCodeFactory: Good point we probably do not need `ENV["SEED_USERS"]` 
This is probably a sufficient to check:
```ruby
Rails.env.production? && CF::App::Credentials.find_by_service_name("psd-seeds")&.key?("organisations")
```

psd-web/db/seeds.rb
```diff
@@ -413,36 +413,68 @@ def create_blob(filename, title: nil, description: nil)
...
-        password: "testpassword",
-        password_confirmation: "testpassword",
-        organisation: organisation,
-        mobile_number_verified: true,
-        teams: [enforcement],
-        mobile_number: phone
-      )
-      %i[superuser team_admin opss_user psd_user user].each do |role|
-        UserRole.create!(user: user, name: role)
+  if Rails.env.production? && (organisations = CF::App::Credentials.find_by_service_name("psd-seeds")["organisations"]) # rubocop:disable Rails/DynamicFindBy
```
slorek, line 416: Please change: This will throw an exception when the service isn't defined (on staging/prod): ```NoMethodError (undefined method `[]' for nil:NilClass)```
```suggestion
  if Rails.env.production? && (organisations = CF::App::Credentials.find_by_service_name("psd-seeds").try(:[], "organisations")) # rubocop:disable Rails/DynamicFindBy
```

PR #513, opened 2020-04-22
Remove titleize from nearly everywhere

edwardhorsford:
The app was using titleize all over the place - mostly incorrectly.

I've removed titelize:
* Where we should not be recasing user input (eg people's names or team names)
* And replaced with `upcase_first` where what we actually want to do is just uppercase the first letter.

Previously we'd have things like this, where the use of `titleize` has incorrectly changed the case of the team names.
Screenshot 2020-04-22 at 12 31 24

It will now look like this:
Screenshot 2020-04-22 at 13 16 12

I searched across the whole app, so this should be removed from all places we use it - including emails.

---

Where we simply want the first letter to be uppercase, I've swapped for `upcase_first`. If those cases eventually have multiword strings, we would not want to uppercase each word. A good example would be business relationship - where we accept a user provided string, and should not be recasing things (though it probably won't go wrong).

## Downsides

A downside of this is that some user's names might now be lowercase / poorly cased - this will be because they originally typed the name like that.

## One remaining `titleize`

There's one remaining `titleize` [here](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/helpers/application_helper.rb#L19). I wasn't sure what it was doing. If a dev can explain, we can decide what to do with it.

## Checklist:
- [?] Has the CHANGELOG been updated? (If change is worth telling users about.)

### DIFF
--- a/psd-web/app/controllers/investigations/assign_controller.rb
+++ b/psd-web/app/controllers/investigations/assign_controller.rb
@@ -28,7 +28,7 @@ class Investigations::AssignController < ApplicationController
     @investigation.assignee = potential_assignee
     @investigation.assignee_rationale = params[:investigation][:assignee_rationale]
     @investigation.save
-    redirect_to investigation_path(@investigation), notice: "#{@investigation.case_type.titleize} was successfully updated."
+    redirect_to investigation_path(@investigation), notice: "#{@investigation.case_type.upcase_first} was successfully updated."
   end
 
 private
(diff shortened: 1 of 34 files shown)


slorek:
@edwardhorsford if you've reviewed all the usage of this helper in the review app then I'm happy. It looks like the `sortable` method is redundant code originally used for column headers.

nasirkhanbeis:
Tested this in review app, verified Name and team name on various activities in audit log and emails, looking all good.

PR #514, opened 2020-04-22
Prevent default on hide cookie link

edwardhorsford:
The link in our cookie banner is used to hide the message - however because the default href is `#`, this means the url is changed when the link is clicked. As the link is exclusively for js users and not for navigation, we can prevent the default link behaviour from happening.

Before, clicking `hide this message`:
Screenshot 2020-04-22 at 14 51 06
Would result in:
Screenshot 2020-04-22 at 14 52 09

Now it looks like:
Screenshot 2020-04-22 at 14 53 48

### DIFF
--- a/psd-web/app/assets/javascripts/cookie_banner.js
+++ b/psd-web/app/assets/javascripts/cookie_banner.js
@@ -9,7 +9,7 @@ function cookieBanner () {
     const expires = `expires=${date.toUTCString()}`
     document.cookie = `seen_cookie_message = true;${expires};path=/`
   }
-  function hideCookieBanner () {
+  function hideCookieBanner (event) {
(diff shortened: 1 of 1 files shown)


psd-web/app/assets/javascripts/cookie_banner.js
```diff
@@ -10,6 +10,7 @@ function cookieBanner () {
     document.cookie = `seen_cookie_message = true;${expires};path=/`
   }
   function hideCookieBanner () {
```
frankieroberto, line 12: Not sure why this isn't required, but I think it'd be good practice, to make the `event` reference more explicit:
```suggestion
  function hideCookieBanner (event) {
```

PR #516, opened 2020-04-23
Convert 2 system tests to feature specs

frankieroberto:
This is a bit of an experiment to see a) how long it'd take and b) how much time it’d save to naively convert our existing minitest system tests (which use Chromium) to rspec feature specs.

I picked two relatively simple tests to start with, and it took less than 10 mins to do the conversion for each one.

I ran each test as a system test and a feature spec a few times to benchmark the time taken:

Introduction test
System test: ~9 seconds
Rspec feature: ~4 seconds

Add activity entry test
System test: ~8 seconds
Rspec feature: ~3 seconds

Based on this, it’d take less than half the time to run the system tests if they were converted to feature specs. Once they were all done, we'd save the time taken to install and initialise chromium too. We’d lose some of this time back though, as feature specs would run after the existing ones, rather than in parallel, unless we can find some other way of parallelising them.

Right now, the rspec feature specs run in around 7 mins, and the system tests in 10 mins. So if we were to convert all the system tests to feature specs it might push the feature specs up to around ~11 mins.

(We could also review the system tests and decide to drop some or find some way of speeding some of them up, but that might take more development effort)


psd-web/spec/features/viewing_introduction_spec.rb
```diff
@@ -0,0 +1,49 @@
+require "rails_helper"
+
+RSpec.feature "Viewing the introduction", :with_stubbed_elasticsearch, :with_stubbed_antivirus, :with_stubbed_mailer, type: :feature do
+
+  let(:user) { create(:user, :activated, has_viewed_introduction: false) }
+
+  scenario "Vieiwng the steps in order" do
```
scruti, line 7:
```suggestion
  scenario "Viewing the steps in order" do
```

psd-web/spec/features/viewing_introduction_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  scenario "Not being shown the introduction twice" do
+    sign_in user
+
+    visit "/introduction/overview"
+    click_on "Continue"
+    click_on "Continue"
+    click_on "Continue"
+    click_on "Get started"
+    visit "/"
+    expect(page).to have_current_path("/")
```
slorek, line 44: Nitpick: would be more readable and consistent with an `expect_to_be_on_home_page` helper or similar. Shows the intent. The system tests aren't the best in this regard.

psd-web/spec/features/viewing_introduction_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  scenario "Skipping the introduction" do
+    sign_in user
+
+    visit "/introduction/overview"
+    click_on "Skip introduction"
+
+    expect(page).to have_current_path("/")
+  end
+
+  scenario "Not being shown the introduction twice" do
```
slorek, line 35: We can add this to the end of the first scenario as it's just repeating the same steps initially.

psd-web/spec/features/viewing_introduction_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    sign_in user
+
+    visit "/introduction/overview"
+    click_on "Continue"
+    click_on "Continue"
+    click_on "Continue"
+    click_on "Get started"
+    visit "/"
+    expect(page).to have_current_path("/")
+  end
```
scruti, line 45: Suggestion: As this seems to reproduce the steps of the "Viewing the steps in order" scenario to extend it with a further visit, I would just append this last visit & assertion over that scenario.

I've been doing this in some of the system tests --> feature tests conversions in order to make the feature tests a little bit faster when multiple checks can be chained instead of reproducing the full setup for each.

Also the assertion in this scenario is a little bit cryptic. I would try to assert over something more obvious like "The introductions is not displayed" kind of assertion.

frankieroberto: I had the same thought about combining this with the previous scenario (or even just deleting it). I kept the tests exactly the same mainly to save on thinking time and to be able to get a better benchmark of how much time it'd save to run the exact same tests without using Chrome.

psd-web/spec/features/viewing_introduction_spec.rb
frankieroberto: Agree. And I think we should put all the `expect_to_be_on_xxxx_page` helpers in a single place.  But for the sake of quick conversion, I wanted to try and avoid any refactoring initially.

psd-web/spec/features/viewing_introduction_spec.rb
frankieroberto: Agree, see [comment below](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/516#discussion_r413662764).

PR #517, opened 2020-04-23
Cancel in-flight deployments when deploying review apps

frankieroberto:
This aims to solve the problem of review apps failing to deploy updates from a new push to a feature branch if the previous deployment is still in progress.


slorek:
Can you do something like https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/deploy-review.sh#L22 where it checks if there's any running deployments and waits until it's clear?

frankieroberto:
@slorek could do, but I think it'd be quicker to just cancel any pre-existing deploys, rather than wait for them? Should be no need to let previous deploys complete, I think, for review apps?

slorek:
@frankieroberto I was thinking to replace the 6 second timeout waiting for the deploy to be cancelled.

frankieroberto:
@slorek yeah I'm not super happy about that `sleep`, but the docs do say "The cancel command is designed to revert the app to its original state as quickly as possible" so hopefully 6 seconds is enough... 🤷‍♂️

PR #519, opened 2020-04-23
Proposal over PR: Inline before action and remove unneeded method

scruti:
Propose changes to https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/480
`UserPasswordsController`

- One of the before filters calls a method that contains just another call to a different method. I believe the second method could be directly called in the filter without the proxy method. 
- Other call mixes logic from the `edit` action with a `require_secondary_authentication` call. It seems to me this should be inline so `edit` logic is wrapped together.


slawosz:
Just to make sure - you assume that 'edit' action does not require 2FA?

scruti:
Mmm... it requires it, but the requirement happens inside the action. Calling it from the filter or inline makes no difference in this case. Right?

PR #522, opened 2020-04-24
Allow safe products to be reported

StupidCodeFactory:
## CHANGES:

[trello](https://trello.com/c/E0ShTNHr/464-5-add-support-for-reporting-safe-products)
This allows to mark a product as safe.

## Checklist:
- [X] Has the CHANGELOG been updated? (If change is worth telling users about.)

### General testing (author)
- [X] Test without JavaScript


psd-web/app/helpers/investigations/ts_instigations_helper.rb
```diff
@@ -0,0 +1,67 @@
+module Investigations::TsInstigationsHelper
```
slorek, line 1:
```suggestion
module Investigations::TsInvestigationsHelper
```
(also rename the file).

PR #529, opened 2020-04-27
Use 'case owner' instead of 'assign'/'assignee' across the interface

frankieroberto:
Following a review as part of the permissions work, we have decided to rename the "assignee" to "case owner".

See https://trello.com/c/tNaCt0xt/436-change-language-of-assignee-to-case-owner

## Implementation notes

This branch _just_ changes the interface, for now, rather than any of the variable names or underlying implementation. Updating these will be part of a follow-up task, currently underway in #549.

## Screenshots

Screenshot 2020-04-29 at 13 42 00

Screenshot 2020-04-29 at 13 42 15

Screenshot 2020-04-29 at 13 43 32


frankieroberto:
@slorek agree that would be useful! But I think slightly out-of-scope for this PR.

I'll merge #549 in to this PR before merging, so that they can get deployed together.

psd-web/app/models/audit_activity/investigation/update_assignee.rb
```diff
@@ -35,7 +35,7 @@ def email_update_text
   end
 
   def email_subject_text
-    "#{investigation.case_type.upcase_first} was reassigned"
+    "Case owner changed for #{investigation.case_type}"
```
edwardhorsford, line 38:
```suggestion
    "Case owner changed on #{investigation.case_type}"
```

edwardhorsford:
> It would be very useful to move the language to I18n so we only have to update it in one place in future. I'd also personally prefer if the underlying code changes were done together with the UI, although I understand the intent is to do it immediately after things don't always go to plan and we could end up with a disconnect for some period of time.

I don't think there would be 'one' place anyway - as the usage often changes depending on the sentence / purpose.

PR #531, opened 2020-04-27
Fix invitations

slorek:
https://trello.com/c/RwQ63f7o/520-bugs-with-invitations-for-users-invited-prior-to-keycloak-migration

## Description
Also includes fixes for https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/527 - I've included them here because it's affecting the deployment due to a new migration being added, and in anticipation of it being merged fairly quickly.

This fixes some bugs around the invitation/registration process, particularly for legacy users invited prior to the removal of Keycloak, but not yet registered.

These users were able to attempt to reset their password, and would then receive the invitation email with no `invitation` param in the link in the email. This was because they have no `invitation_token` set. And the form would render because there was no check for this combination of conditions, but would show an error message only after filling in the form.

We would instead expect that when this type of user attempts to reset their password they would see an invitation expired error message. This was not happening because the `invited_at` timestamp was also `nil`, and our code treated this as _not expired_ and therefore send the invitation email again.

It appears we wanted to delay setting the `invited_at` until the email was actually sent, presumably in the event of a delay in email delivery. But I think given the problem it has introduced and the threshold for expiry, we can remove this, and I think this more accurately captures the time of the event.

To combat this I have set the users `invited_at` column to be `NOT NULL` and migrated any existing data to set this to the `created_at` time (this has already been done in production). In order to keep things DRY and to avoid forgetting to set this on user creation I've added default values to the database and the model. The reason for doing so in the model as well was because database defaults would only be accessible to the app after `reload` which wasn't convenient.

This should not introduce any issues because even when we create a new organisation and team admin they will receive an invitation email.

I have not set the `invitation_token` column to be `NOT NULL` because we agreed not to change this for existing users. When their team admin resends their invite the token will be reset. But I have made the default behaviour of the model to generate the token so that we don't have to remember to keep doing this.

I've also added the edge case scenario to the request spec.

## Checklist:

### DIFF
--- a/psd-web/db/migrate/20200427131227_user_invited_at_not_null.rb
+++ b/psd-web/db/migrate/20200427131227_user_invited_at_not_null.rb
@@ -1,10 +1,14 @@
 class UserInvitedAtNotNull < ActiveRecord::Migration[5.2]
   def up
-    User.where(invited_at: nil).update_all("invited_at=created_at")
-    change_column :users, :invited_at, :datetime, null: false, default: -> { "CURRENT_TIMESTAMP" }
+    safety_assured do
+      User.where(invited_at: nil).update_all("invited_at=created_at")
+      change_column :users, :invited_at, :datetime, null: false, default: -> { "CURRENT_TIMESTAMP" }
+    end
   end
 
   def down
-    change_column :users, :invited_at, :datetime, null: true, default: nil
+    safety_assured do
+      change_column :users, :invited_at, :datetime, null: true, default: nil
+    end
   end
 end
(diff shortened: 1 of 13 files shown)


psd-web/db/migrate/20200427131227_user_invited_at_not_null.rb
```diff
@@ -0,0 +1,10 @@
+class UserInvitedAtNotNull < ActiveRecord::Migration[5.2]
+  def up
+    User.where(invited_at: nil).update_all("invited_at=created_at")
+    change_column :users, :invited_at, :datetime, null: false, default: -> { "CURRENT_TIMESTAMP" }
```
StupidCodeFactory, line 4: are you sure it's not "CURRENT_TIMESTAMP()"?

slorek: I tested it works.

slorek:
@frankieroberto I think we still need that code because we have a few users who still do not have an `invitation_token`.

PR #532, opened 2020-04-27
Allow user deletion

scruti:
[Trello ticket](https://trello.com/c/Yt7uEDsr/338-user-deletion-deactivation)

## Description
Introduce the possibility of deleting users via rake task.

Users won't be effectively deleted from the system but flagged with a `deleted_at` timestamp. 

When an user is marked as deleted:
- Their cases will be assigned to their team.
- Won't be able to sign in.
- Won't be able to receive a password reset.
- Won't be able to be invited to any team.
- Won't receive any email notifications.
- Won't be displayed in the team member list.
- Won't appear in the user list for filtering cases.
- Won't appear in case list/assignment autocomplete.
- Their name will be displayed as `User Name [user deleted]` when possible.

To delete a user:
```
 EMAIL=example@example.com rake user:delete  
```
or
```
 ID=id-sequence rake user:delete 
```

## Checklist:


scruti:
[User deletion/deactivation](https://trello.com/c/Yt7uEDsr/338-user-deletion-deactivation)

edwardhorsford:
Had a discussion with @scruti.

I reckon two different functions to decorators for names across two different files isn't ideal. We've suggested leaving as-is for now, but following up with a story to clean up the `display_name` function, and probably move it out of the user model.

psd-web/app/models/user.rb
```diff
@@ -91,13 +91,12 @@ def name
...
 
-    suffix = if deleted?
-               "[user deleted]"
-             elsif (ignore_visibility_restrictions || (organisation_id == other_user&.organisation_id)) && teams.any?
+    suffix = if (ignore_visibility_restrictions || (organisation_id == other_user&.organisation_id)) && teams.any?
                "(#{team_names})"
              else
                "(#{organisation.name})"
              end
+    suffix << " [user deleted]" if deleted?
```
StupidCodeFactory, line 99: probably a good occasion to refactor/move this to the decortator. This really doesn't belong to the database layer.

StupidCodeFactory:
@scruti did have a look into https://github.com/rubysherpas/paranoia ?

psd-web/app/models/user.rb
scruti: Yeah definitely. Originally we were thinking about doing a follow up story but does not seem to going to happen. So will go for it here.

scruti:
@StupidCodeFactory I am trying to avoid to rely on default scopes as much as possible (and Paranoia seems to be relying on them) + seems their own maintainers are discouraging to adopt it nowadays. But was nice to peek around the project :+1:

psd-web/app/decorators/user_decorator.rb
```diff
@@ -8,11 +8,24 @@ def assignee_short_name(viewing_user:)
...
   end
 
-  def name
-    if deleted?
-      "#{object.name} [user deleted]"
-    else
-      object.name
-    end
+  def full_name
+    object.name + deleted_suffix
```
edwardhorsford, line 12:
```suggestion
    name + deleted_suffix
```
Should this just be `name`?

scruti: good catch. `object.name` was useful when this method was called `name` but now is unneeded.

psd-web/spec/features/sign_in_spec.rb
```diff
@@ -172,6 +172,15 @@ def expect_incorrect_email_or_password
     expect_incorrect_email_or_password
   end
 
+  scenario "user tries to sign in with email address belonging to a deleted account" do
+    user.update_column(:deleted, true)
```
slorek, line 176: Suggestion: Prefer to create a new `context` block and override `let(:user)` with the correct properties. This sets up the scenario fully before execution.

psd-web/app/services/delete_user.rb
```diff
@@ -0,0 +1,25 @@
+class DeleteUser
```
StupidCodeFactory, line 1: 🤟

psd-web/app/services/delete_user.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+private
+
+  def assign_user_investigations_to_their_team
+    context.team = context.user.teams.first
+
+    context.user.investigations.each do |investigation|
+      investigation.assignable = context.team
+      investigation.save
```
StupidCodeFactory, line 21: what happens when this returns false?

scruti:
@slorek 
> I am not sure but I think we might need to implement display_name in the TeamDecorator too for instances where the assignee or source was a Team. I think this will bubble the call up to the Team class but that is inconsistent.

Yes I relied in the method call being handled by the model delegation. Got tired of replacing things around and left it like it was :sweat_smile:  Will change it to keep the consistency.

psd-web/app/views/investigations/assign.html.slim
```diff
@@ -10,7 +10,7 @@
       - if @investigation.assignable.present?
         p
           | Currently assigned to:
-          strong = " #{@investigation.assignable.display_name}"
+          strong = " #{@investigation.assignable.decorate.display_name}"
```
StupidCodeFactory, line 13: decorate the investigation in the controllers, not in the views

psd-web/app/views/investigations/assign/_assignee_form.html.slim
```diff
@@ -4,7 +4,7 @@
 
   ruby:
     items = investigation.important_assignable_people.map do |user|
-      { text: user.display_name, value: user.id, checked: investigation.assignable == user }
+      { text: user.decorate.display_name, value: user.id, checked: investigation.assignable == user }
```
StupidCodeFactory, line 7: Same decorate in controller not in views

psd-web/app/models/user.rb
```diff
@@ -37,8 +37,12 @@ class User < ApplicationRecord
   attribute :invitation_token, :string, default: -> { SecureRandom.hex(15) }
   attribute :invited_at, :datetime, default: -> { Time.current }
 
-  def self.activated
-    where(account_activated: true)
+  def self.active
```
frankieroberto, line 40: Might be worth adding a comment to remind our future-selves what `active` means? Something like "Active users are those with current access to the service (ie have set up an account and haven't been deleted) and who have accepted the user declaration"?

psd-web/app/models/user.rb
```diff
@@ -37,8 +37,12 @@ class User < ApplicationRecord
...
+    where(account_activated: true, deleted: false)
```
frankieroberto, line 41: Could chain this with the other scope, ie:

```suggestion
    where(account_activated: true).not_deleted
```

Possibly slightly easier to read?

psd-web/lib/tasks/delete_user.rake
```diff
@@ -0,0 +1,28 @@
+namespace :user do
+  desc "Marks the given user as deleted, assigning their investigations to their team"
+  task delete: :environment do
+    user_id = ENV.fetch("ID", nil)
+    user_email = ENV.fetch("EMAIL", nil)
+
+    if !user_id && !user_email
+      puts "Error: Need to provide user ID or EMAIL with the call."
+      puts "Eg: EMAIL=example@example.com rake user:delete"
```
StupidCodeFactory, line 9: Rake task will execute successfully. i.e return 0. Instead be a nice shell citizen and raise an error so the error  message is written to STDERR:
```ruby
raise RuntimeError, "Error: Need to provide user ID or EMAIL with the call.\n\nEg: EMAIL=example@example.com rake user:delete"
```

psd-web/app/models/user.rb
```diff
@@ -154,7 +146,14 @@ def invitation_expired?
     invited_at <= INVITATION_EXPIRATION_DAYS.days.ago
   end
 
+  def mark_as_deleted!
```
frankieroberto, line 149: I was tempted to suggest renaming this to `delete!` but on second thoughts your method name is better, as it'd be easy to assume that `delete!` actually deletes the record.

psd-web/spec/decorators/user_decorator_spec.rb
```diff
@@ -21,4 +21,125 @@
...
+    let(:result) { decorated_user.display_name(other_user: other_user, ignore_visibility_restrictions: ignore_visibility_restrictions) }
+
+    context "when the user is a member of the same organisation" do
+      context "when the user has no teams" do
+        it "returns their name and organisation name" do
+          expect(result).to eq("#{user_name} (#{organisation_name})")
+        end
+
+        it "deleted users get a 'user deleted' flag" do
+          user.update_column(:deleted, true)
```
StupidCodeFactory, line 70: why not use your `mark_as_deleted!` method?

psd-web/app/services/delete_user.rb
```diff
@@ -0,0 +1,25 @@
...
+    ActiveRecord::Base.transaction do
+      context.user.mark_as_deleted!
+      assign_user_investigations_to_their_team
+    end
+  end
+
+private
+
+  def assign_user_investigations_to_their_team
+    context.team = context.user.teams.first
```
frankieroberto, line 17: We should probably guard against a user not belonging to any teams here, even though hopefully it'll never happen for real.

Might also be worth adding a note to say why we’re using `teams.first` – ie that users should only belong to one team, and that we plan on enforcing this in future?

psd-web/spec/decorators/user_decorator_spec.rb
```diff
@@ -21,4 +21,125 @@
       it { expect(decorated_user.assignee_short_name(viewing_user: viewing_user)).to eq(user.organisation.name) }
     end
   end
+
```
StupidCodeFactory, line 24: ❤️   Thank you for being consistent with the rest of the code by using `eq` and not `eql` ;)

psd-web/app/services/delete_user.rb
frankieroberto, line 21: I think the job ought to be raising an exception if the save fails?

```suggestion
      investigation.save!
```

psd-web/spec/decorators/user_decorator_spec.rb
```diff
@@ -21,4 +21,125 @@
...
+        expect(decorated_user.full_name).to eq("Foo Bar [user deleted]")
+      end
+    end
+
+    it "non deleted users show their name" do
+      expect(decorated_user.full_name).to eq(user.name)
+    end
+  end
+
+  describe "#display_name" do
```
StupidCodeFactory, line 39: I wonder whether this is not a few minitests that we could delete?

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -13,12 +13,28 @@
...
       fill_in "Email address", with: Faker::Internet.safe_email
       click_on "Send email"
 
+      expect(delivered_emails).to be_empty
+      expect_to_be_on_check_your_email_page
+    end
+  end
+
+  scenario "entering an email from a deleted user does not send a notification but shows the confirmation page" do
+    user.update!(deleted: true)
```
StupidCodeFactory, line 26: use your method! or even better use the interactor!!!
Or even even create a factory bot `trait :deleted` 

```ruby
trait :deleted { after(:create) { |user| DeteteUser.call(user: user) } }
```
This way we don't have to maintain the business logic in 2 places the factories and the production code and another advantage is that we'll be testing against more realistic data. 🧙‍♂️

psd-web/db/migrate/20200417161318_add_email_unique_index_to_users.rb
```diff
@@ -0,0 +1,13 @@
+class AddEmailUniqueIndexToUsers < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def up
+    remove_index :users, column: :email, algorithm: :concurrently
+    add_index :users, :email, unique: true, algorithm: :concurrently
```
frankieroberto, line 6: Hooray! Finally emails are guarantees to be unique. 👏

psd-web/spec/features/password_reset_spec.rb
```diff
@@ -165,26 +181,22 @@ def request_password_reset
...
       click_on "Send email"
 
+      expect(delivered_emails.size).to eq 1
+      email = delivered_emails.first
+
+      expect(email.recipient).to eq user.email
+      expect(email.reference).to eq "Password reset"
+      expect(email.template).to eq NotifyMailer::TEMPLATES[:reset_password_instruction]
+      expect(email.personalization[:name]).to eq user.name
+      expect(email.personalization[:edit_user_password_url_token]).to eq edit_user_password_url_with_token
```
StupidCodeFactory, line 198:
```suggestion
      expect(email.personalization).to eq(edit_user_password_url_token: edit_user_password_url_with_token, name: user.name)
```

StupidCodeFactory: and remove line 197

psd-web/spec/features/report_product_spec.rb
```diff
@@ -413,14 +413,14 @@ def expect_case_attachments_page_to_show(file_description:)
 
   def expect_case_activity_page_to_show_allegation_logged
     item = page.find("h3", text: "Allegation logged: #{product_details[:name]}, #{product_details[:type]}").find(:xpath, "..")
-    expect(item).to have_text("Assigned to #{user.display_name}")
+    expect(item).to have_text("Assigned to #{user.decorate.display_name}")
```
StupidCodeFactory, line 416: decorate the factory ;)

psd-web/spec/features/sign_in_spec.rb
StupidCodeFactory, line 176: use your method!

psd-web/spec/models/audit_activity/investigation/automatically_reassign_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  let(:team) { create(:team) }
+  let(:investigation) { create(:enquiry, assignable: team, assignee_rationale: "Test assign") }
+
+  describe ".from" do
+    it "creates an audit activity without information for email notification", :aggregate_failures do
+      expect(audit_activity.title).to eq("Enquiry automatically reassigned to #{team.display_name}")
+      expect(audit_activity.body).to eq(nil)
+      expect(audit_activity.email_update_text).to eq(nil)
+      expect(audit_activity.email_subject_text).to eq(nil)
```
StupidCodeFactory, line 14:
```ruby
      expect(audit_activity).to have_attributes(title: "....", body: nil, email_update_text: nil, email_subject_text: nil)
```

psd-web/spec/models/audit_activity/investigation/update_assignee_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    ex.run
+    User.current = nil
+  end
+
+  describe ".from" do
+    it "creates an audit activity with information for email notification", :aggregate_failures do
+      expect(audit_activity.title).to eq("Assigned to #{user.decorate.display_name}")
+      expect(audit_activity.body).to eq(investigation.assignee_rationale)
+      expect(audit_activity.email_update_text).to start_with("Enquiry was assigned to #{user.decorate.display_name} by #{user.name}")
+      expect(audit_activity.email_subject_text).to eq("Enquiry was reassigned")
```
StupidCodeFactory, line 20: same here use the `have_attributes` matcher

psd-web/spec/models/audit_activity/investigation/update_coronavirus_status_spec.rb
```diff
@@ -12,7 +12,7 @@
     it "creates an audit activity", :aggregate_failures do
       expect(audit_activity.title).to eq("Status updated: not coronavirus related")
       expect(audit_activity.body).to eq("The case is not related to the coronavirus outbreak.")
-      expect(audit_activity.email_update_text).to eq("#{investigation.case_type.capitalize} #{investigation.pretty_id} is not related to the coronavirus outbreak. This status was updated by #{user.display_name}.")
+      expect(audit_activity.email_update_text).to eq("#{investigation.case_type.capitalize} #{investigation.pretty_id} is not related to the coronavirus outbreak. This status was updated by #{user.decorate.display_name}.")
       expect(audit_activity.email_subject_text).to eq("Coronavirus status updated on #{investigation.case_type.downcase}")
```
StupidCodeFactory, line 16: `have_attributes`

psd-web/app/decorators/user_decorator.rb
```diff
@@ -7,4 +7,25 @@ def assignee_short_name(viewing_user:)
 
     name
   end
+
+  def full_name
+    name + deleted_suffix
+  end
```
frankieroberto, line 13: Any reason to do this in the decorator rather than within the existing `def name` method on the user model? I can't think of any context where we _wouldn’t_ want to add the deleted suffix? This will require us to remember to always use the decorator, and the different method name...

StupidCodeFactory:
@slorek has a good point we need to add `#display_name` to the team decorator as these are polymorphic model we need to maintain the same interfaces! I wish ruby had interfaces like Java that would be amazing.

psd-web/app/decorators/user_decorator.rb
edwardhorsford: I'd like to soon introduce several decorators:

`full_name`
`full_name_with_team`
`full_name_with_conditional_team`
`conditional_team`

I suspect it would be helpful to keep it consistent and use `full_name` rather than just `name` in a rare occasion?

psd-web/app/decorators/user_decorator.rb
```diff
@@ -7,4 +7,25 @@ def assignee_short_name(viewing_user:)
 
     name
   end
+
+  def full_name
```
slawosz, line 11: Regarding whole decorator: how about having separate decorator only for deleted users?

psd-web/app/models/user.rb
```diff
@@ -201,7 +205,7 @@ def two_factor_lock_expired?
   end
 
   def send_reset_password_instructions_notification(token)
-    NotifyMailer.reset_password_instructions(self, token).deliver_later
+    NotifyMailer.reset_password_instructions(self, token).deliver_later unless deleted?
```
slawosz, line 208: My preference would be `return unless deleted?` before sending email.

psd-web/app/views/teams/show.html.slim
```diff
@@ -7,7 +7,7 @@
       a.govuk-button role="button" href=invite_to_team_url Invite a team member
     h2.govuk-heading-m Team members
     ul.teams--list
-      - @team.users.sort_by(&:name).each do |user|
+      - @team.users.not_deleted.sort_by(&:name).each do |user|
```
slawosz, line 10: I think `team.users` should not return deleted users anyway.

slawosz:
One general thing is calling `decorate` inside view. I would prefer to pass already decorated users to the view.

psd-web/app/decorators/user_decorator.rb
frankieroberto: I think `full_name` is a bit unexpected, as usually that's used to describe someone's first, middle and last names.

`name_with_status` maybe? Would allow other types of suffix in future...

psd-web/app/models/audit_activity/investigation/automatically_reassign.rb
```diff
@@ -0,0 +1,31 @@
...
+    super(investigation, title)
+  end
+
+  def assignable_id
+    # We store assignable_id in title field, this is getting it back
+    # Using alias for accessing parent method causes errors elsewhere :(
+    AuditActivity::Investigation::Base.instance_method(:title).bind(self).call
+  end
+
+  # We store assignable_id in title field, this is computing title based on that
```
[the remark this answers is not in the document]
StupidCodeFactory: yes me neither, this file was copied over from another file.

psd-web/app/services/delete_user.rb
StupidCodeFactory: This was my implicit suggestion ;)

I think we need to have a conversation about this `record.save` or `record.update` it's something I see very often in pull requests and can potentially have disastrous side effects/hard to track bugs down the line.

psd-web/app/views/teams/show.html.slim
StupidCodeFactory: I was thinking the same as @slawosz. 
To avoid making so many changes, potentially bug prone, I would have added the `type` column and used STI with `class DeletedUser < AnonymousUser` and `class User < AnonymousUser`
then all associations referencing `belongs_to :user` or `has_many :users` would automatically not include deleted user and there would not been any need to change scopes anywhere.

psd-web/spec/features/sign_in_spec.rb
scruti: If I had multiple scenarios to test under the same conditions I would take that route, but when is a single scenario and the context is not going to be shared I prefer to keep the test structure as flat as possible.

psd-web/app/models/audit_activity/investigation/automatically_reassign.rb
```diff
@@ -0,0 +1,31 @@
+class AuditActivity::Investigation::AutomaticallyReassign < AuditActivity::Investigation::Base
+  def self.from(investigation)
+    title = investigation.assignable.id.to_s
+    super(investigation, title)
+  end
+
+  def assignable_id
+    # We store assignable_id in title field, this is getting it back
+    # Using alias for accessing parent method causes errors elsewhere :(
+    AuditActivity::Investigation::Base.instance_method(:title).bind(self).call
```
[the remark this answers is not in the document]
scruti: Oh is worse than you think. We need to keep an history of assignee in order to be able to display the historical update activities:
```
Case assigned to UserA on X date
Case assigned to UserB on Y date
```
By grabbing the `assignable_id` from the investigation we only get the current assignee, and that wouldn't work when having to show a previous assignee update activity. 

Instead of adding a field for it in `activities`, seems the decision was to hack the title field to store the id there. 
I think we should just add the field to the activities table and populate it only for the needed subclass of activities.  Then back-fill it for all the "update assignee" activities subtype.

I just didn't want to get into solving that in a piece of work/PR that is already big enough.

psd-web/app/models/audit_activity/investigation/automatically_reassign.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  # We store assignable_id in title field, this is computing title based on that
+  def title
+    type = investigation.case_type.capitalize
+    new_assignable = (User.find_by(id: assignable_id) || Team.find_by(id: assignable_id))&.display_name
```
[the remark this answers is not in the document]
scruti: For the last one yes. Problem is when having to display previous ones in the activity history.

psd-web/spec/decorators/user_decorator_spec.rb
scruti: :man_facepalming:  true

psd-web/app/views/investigations/assign/_assignee_form.html.slim
scruti: In this case the users/teams get retrieved in the view, and the view display is conditional.  So would have to refactor the assignee form and maybe the assignee in order to be able to decorate this in the controller. Not keen to keep digging into the rabbit hole.

psd-web/app/models/audit_activity/investigation/automatically_reassign.rb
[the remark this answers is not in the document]
scruti: Yes, copied from the [Update Assignee activity](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/2f3225d5cbcef11c6c9bd30ff989157fd3886a65/psd-web/app/models/audit_activity/investigation/update_assignee.rb).

What is happening is that we need to know who was the assigned user when the activity was recorded (not only the current assigned one) in order to show reassign history in the case activities.

The `activities` table does not have an `assignable` field for this so... someone decided that was a good idea to use `title` for storing the assigned user id.

They could have just hardcode the title. But seems the plus of being able to compute over the names and change the title display on the spot was a requirement (I am wild guessing here).

When I originally wrote this new file, it was a subclass of the `update_assignee.rb` hence copying the same code (but ended moving it back to side-by-side) and ended forgetting about this bit of implementation.

I think the field should be just added into the DB and backfilled with a migration for all the assignee relying activities, but thought I was going waay out of scope for the task. Maybe not so much if I am adding a new type of activity doing the same nasty hack. But it may be worth to attack this independently?

psd-web/spec/decorators/user_decorator_spec.rb
scruti: Re-thinking this I don't want the test setup to have dependencies in other User method behaviour unless strictly necessary (if the interface changes, would break tests).
Going to extract this to the user setup though to avoid the update and rely in the factory creation.

psd-web/app/models/audit_activity/investigation/automatically_reassign.rb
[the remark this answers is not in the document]
frankieroberto: Yikes, that really is a hack. Thanks for the explanation!

Agree that tackling this is way out of scope for the PR though. Perhaps we need a way of saving arbitrary metadata to the activity logs - eg as a JSON blob? One for the future...

psd-web/spec/decorators/user_decorator_spec.rb
scruti: Good one. We can delete the `UserTest` file now :raised_hands:

psd-web/spec/features/password_reset_spec.rb
scruti: Not really keen on make a factory trait dependant on a Service Object. 
I see the test setup responsibility as setting the minimum necessary in order to test some behaviour. If we start adding calls to services when not strictly necessary we trigger things we not relevant in the tested scenario and also make the tests dependant on code that may change. 
So in these cases I prefer to stick to flagging the user as `deleted`. Going to extract it to the test context setup though.

psd-web/app/models/audit_activity/investigation/automatically_reassign.rb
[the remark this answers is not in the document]
scruti: Just going back to explain it caused me nightmares :/. Thanks for being understanding on this!

psd-web/app/decorators/user_decorator.rb
scruti: mmm... what I imagine on this is a little bit of an overkill. But if I see a code proposal on this it may be simpler than what I think!

psd-web/app/services/delete_user.rb
scruti: Good point. Yeah the `teams.first` is a little bit "just because" without context.

psd-web/app/services/delete_user.rb
scruti: You guys are right! Sloppy Marc!

psd-web/app/views/teams/show.html.slim
scruti: Well... I can imagine at least a situation where we could want that (admins checking all the users/deleted or not). But yeah, I see the point of, 99% of the time, not wanting deleted users to be included.
I just avoided default scopes like the plague (or in modern terms, like Covid-19). But didn't consider the STI approach to be honest. 

With STI my only concern would be the same as with inheritance in general. That in the moment where the conditions get more complicated or new combinations or types are needed/discovered, the inheritance taken structure may not work anymore.

scruti:
@frankieroberto about `deleted_at` & `deleted_by` I just took the simplest approach that works with the current requirements. 
`deleted_at` could replace `deleted` field as flag field. But... do we need to know this info? Also the `user.deleted?` would require to have logic to wrap the timestamp check instead of being automatically provided by AR. But I recognise the cost of this is marginal and I know there are some recommendations for using `null/timestamp` as alternative for `false/true` flag attributes.

For the `deleted_by` I expect to be introduced whenever we develop the possibility of marking users as deleted from the user interface. At this stage of development, triggered by developers on the terminal, this concept didn't make sense to me. So left it out.

psd-web/lib/tasks/delete_user.rake
```diff
@@ -0,0 +1,31 @@
...
+
+      result = DeleteUser.call(user: user)
+      if result.success?
+        puts "User #{id.presence || email} successfully marked as deleted."
+        puts "User investigations assigned to #{result.team.name}"
+      else
+        puts result.error
+      end
+    rescue ActiveRecord::RecordNotFound
+      puts "Error: User #{id.presence || email} not found"
```
StupidCodeFactory, line 18:
```suggestion
      STDERR.puts "Error: User #{id.presence || email} not found"
```

psd-web/lib/tasks/delete_user.rake
```diff
@@ -0,0 +1,31 @@
...
+    # rubocop:disable Rake/MethodDefinitionInTask
+    def delete_user(id, email)
+      user = id ? User.find(id) : User.find_by!(email: email)
+
+      result = DeleteUser.call(user: user)
+      if result.success?
+        puts "User #{id.presence || email} successfully marked as deleted."
+        puts "User investigations assigned to #{result.team.name}"
+      else
+        puts result.error
```
StupidCodeFactory, line 15:
```suggestion
        STDERR.puts result.error
```

psd-web/spec/models/user_spec.rb
```diff
@@ -419,4 +352,30 @@
...
+    it "does not change the flag if was already enabled" do
+      user = create(:user, :deleted)
+      expect { user.mark_as_deleted! }.not_to change(user, :deleted_at)
+    end
+  end
+
+  describe "#deleted?" do
+    it "returns true for users with deleted timestamp" do
+      user = create(:user, :deleted)
+      expect(user.deleted?).to be true
```
StupidCodeFactory, line 373:
```suggestion
      expect(user).to be_deleted
```

psd-web/spec/models/user_spec.rb
```diff
@@ -419,4 +352,30 @@
...
+    end
+
+    it "returns false for users without deleted timestamp" do
+      user = create(:user)
+      expect(user.deleted?).to be false
```
StupidCodeFactory, line 378:
```suggestion
      expect(user).to be_deleted
```

psd-web/app/models/user.rb
```diff
@@ -150,11 +144,24 @@ def has_viewed_introduction!
     update has_viewed_introduction: true
   end
 
+  def deleted?
+    deleted_at.present?
+  end
```
StupidCodeFactory, line 149: No biggy but this is superfluous and using active record generated method `#deleted_at?` would have been more  enough

PR #535, opened 2020-04-28
Fix invite

slorek:
https://trello.com/c/RwQ63f7o/520-bugs-with-invitations-for-users-invited-prior-to-keycloak-migration

## Description
Due to a lack of tests around resending user invitations, https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/531 introduced a bug whereby a user whose invitation was expired could not have their invitation resent. They would instead receive the "invitation expired" email.

The fix is to reset their `invited_at` timestamp when the team admin resends their invitation. This means the job will send the correct invitation email.

I've taken the opportunity to add a spec around this scenario and refactor some of the existing tests, and removed a few redundant Minitest tests.

Thanks to @nasirkhanbeis for spotting this bug.

## Checklist:

### DIFF
--- a/psd-web/spec/features/invite_user_spec.rb
+++ b/psd-web/spec/features/invite_user_spec.rb
@@ -1,0 +1,24 @@
+require "rails_helper"
+
+RSpec.feature "Inviting a user", :with_stubbed_mailer, :with_stubbed_elasticsearch, type: :feature do
+  let(:team) { create(:team) }
+  let(:user) { create(:user, :activated, :team_admin, teams: [team], has_viewed_introduction: true) }
+
+  before do
+    sign_in(user)
+    visit invite_to_team_url(team)
+    expect_to_be_on_invite_a_team_member_page
+  end
+
+  context "when there is already a user with that email" do
+    scenario "shows an error message" do
+      fill_in "new_user_email_address", with: user.email
+      click_button "Send invitation email"
+      expect(page).to have_css(".govuk-error-summary__list", text: "You cannot invite this person to join your team because they are already a member of another team from a different organisation.")
+    end
+  end
+
+  def expect_to_be_on_invite_a_team_member_page
+    expect(page).to have_css("h1", text: "Invite a team member")
+  end
+end
--- a/psd-web/spec/features/team_spec.rb
+++ b/psd-web/spec/features/team_spec.rb
@@ -5,7 +5,7 @@ RSpec.feature "Your team page", :with_stubbed_mailer, :with_stubbed_elasticsearc
   let(:user) { create(:user, :activated, :psd_user, teams: [team], has_viewed_introduction: true) }
 
   let!(:another_active_user) { create(:user, :activated, email: "active.sameteam@example.com", organisation: user.organisation, teams: [team], has_viewed_introduction: true) }
-  let!(:another_inactive_user) { create(:user, email: "inactive.sameteam@example.com", organisation: user.organisation, teams: [team]) }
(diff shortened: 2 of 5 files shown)


psd-web/spec/features/invite_user_spec.rb
```diff
@@ -0,0 +1,24 @@
+require "rails_helper"
+
+RSpec.feature "Inviting a user", :with_stubbed_mailer, :with_stubbed_elasticsearch, type: :feature do
```
slorek, line 3: This is an existing test moved out of `team_spec.rb`.

psd-web/spec/features/team_spec.rb
```diff
@@ -2,58 +2,60 @@
 
 RSpec.feature "Your team page", :with_stubbed_mailer, :with_stubbed_elasticsearch, type: :feature do
```
slorek, line 3: I've refactored the existing tests to improve the code and reduce the number of scenarios to improve speed.

psd-web/spec/features/team_spec.rb
```diff
@@ -2,58 +2,60 @@
...
+    scenario "displays the invite a team member link and only displays the resend invite link for inactive users" do
+      expect(page).to have_link("Invite a team member")
+      expect(page).to have_resend_invite_link_for(another_inactive_user)
+      expect(page).not_to have_resend_invite_link_for(another_active_user)
     end
 
-    scenario "does not display the invite a team member link" do
-      expect(page).to have_css("h1", text: "test organisation")
-      expect(page).not_to have_button("Invite a team member")
+    scenario "resending an invitation sends an email to the user and shows a confirmation message" do
```
slorek, line 43: This is the new test introduced for the bug.

psd-web/spec/features/team_spec.rb
scruti: Ouch! I'm going to get some ugly conflicts in my "delete user" PR (that adds a bunch of tests here too). Nice improvement!

psd-web/spec/features/team_spec.rb
```diff
@@ -2,58 +2,60 @@
...
+      click_link "Resend invitation to #{another_inactive_user.email}"
+      expect(message_delivery_instance).to have_received(:deliver_now)
```
scruti, line 45: Suggestion: 
@slawosz introduced `delivered_emails` in the `with_delivered_emails` spec helper. Allowing us a closer to end-to-end approach by just asserting over the delivered emails list & recipients. 
I personally prefer to use it on feature tests over asserting on code calls.

psd-web/spec/models/user_spec.rb
```diff
@@ -179,13 +179,11 @@
 
       before { allow(SendUserInvitationJob).to receive(:perform_later) }
 
-      # rubocop:disable RSpec/MultipleExpectations
-      it "raises an exception and does not send the invitation" do
+      it "raises an exception and does not send the invitation", :aggregate_failures do
```
scruti, line 182: :clap:

psd-web/spec/features/team_spec.rb
slorek: `delivered_emails` doesn't tell us which email was sent - it only provides the personalisation arguments. In this test I need to know that the invitation email was sent instead of the expired invitation email. It also appears that the stub is not preventing emails being sent as it is merely wrapping the implementation to add to the `delivered_emails` array, so it's probably talking to the Notify API, which we should not be doing in tests. I'll take a look at improving that as it might be related to the currently flaky tests which are making calls to the Notify API.

slorek: I improved the `delivered_emails` so we can use that now instead. Thanks!

PR #538, opened 2020-04-28
Show team name rather than organisation in creator section

edwardhorsford:
Updates the `case creator` field on the case overview to show a user's team rather than their org.

Before:
Screenshot 2020-04-28 at 15 16 00

After:
Screenshot 2020-04-28 at 15 15 32

Notes:
I wasn't sure if we still needed to test for presence so I left that in. Perhaps a dev can advise.


psd-web/app/decorators/investigation_decorator.rb
```diff
@@ -119,7 +119,7 @@ def created_by
            else
              h.escape_once(source.user.name.to_s)
            end
-    out << h.escape_once(source.user&.organisation&.name) if source&.user&.organisation.present?
+    out << h.escape_once(source.user&.team_names) if source&.user&.team_names.present?
```
scruti, line 122: Regarding your question in the PR description, the `present?` seems still necessary, yes. As the Source may or may not have an user. And we only want to write this line if the source has an user and that user has `team_names`.

PR #539, opened 2020-04-28
Remove assignee methods

StupidCodeFactory:
While investigatin random test failure around 
`Investigation#assignee_team` it became clear that `Investigation#assignee=` was swallowing  exceptions in the test.
I've removed it in favour of the real polymorphic associations that does not swallow exception `belongs_to :assignable, polymorphic: true, optional: true` (we could probably remove the `optional: true` but I'll keep this for a separate PR.
I've also address gross misuse of polymorphism and extract a couple method to the the `User` and `Team` polymorphic models. to remove redundant check to `assignable.is_a?(User||Team)` and let duck typing do the job.


psd-web/app/models/investigation.rb
```diff
@@ -154,10 +131,7 @@ def add_pretty_id
   def child_should_be_displayed?
     # This method is responsible for white-list access for assignee and their team, as described in
     # https://regulatorydelivery.atlassian.net/wiki/spaces/PSD/pages/598933517/Approach+to+case+sensitivity
-    return true if (self.assignee.is_a? Team) && self.assignee.users.include?(User.current)
-    return true if (self.assignee.is_a? User) && (self.assignee.teams & User.current.teams).any?
-
-    false
+    assignable.display_child?(User.current)
```
StupidCodeFactory, line 134: behaviour has been extracted to where to each of its respective classes. see below

frankieroberto:
@StupidCodeFactory we’re planning on renaming this to `owner` in #529 to match the term used in the UI.

StupidCodeFactory:
> @StupidCodeFactory we’re planning on renaming this to `owner` in #529 to match the term used in the UI.

Then that's a first step, you'll only have to rename the database fields `assignable_type` and `assignable_id` to `owner_type` and `owner_id`.

PR #547, opened 2020-04-29
Don't profile rspec tests by default

frankieroberto:
This stops rspec from profiling the 10 slowest tests by default, as this can be annoying when you're only running the specs in a specific file or line, as it means you have to scroll further up in the output to see the result, or remember to add the `--no-profile` flag.

You can still opt-in to see the slowest specs, by running:

```ruby
bundle exec rspec --profile -- spec/
```

to show the slowest 10, or

```ruby
bundle exec rspec --profile 5 spec/
```

to show a different number of the slowest specs.


StupidCodeFactory:
can you not added this to your ~/.rspec ?

StupidCodeFactory:
https://relishapp.com/rspec/rspec-core/v/3-6/docs/configuration/read-command-line-configuration-options-from-files

PR #548, opened 2020-04-29
Fix healthcheck

slawosz:
Missing spec


psd-web/spec/requests/health_spec.rb
```diff
@@ -0,0 +1,17 @@
+require "rails_helper"
+
+# rubocop:disable RSpec/DescribeClass
```
StupidCodeFactory, line 3: do you really need that?

PR #554, opened 2020-04-29
Don't output team and org names if they're the same

edwardhorsford:
Our assignee section currently shows the assignee and their org. For many users, the org name is the same as the team name.

Where the team is the assignee, this means we show the same text twice.

This pr cleans up the markup a little (no unnecessary divs) and uniques the names.

Before:
Screenshot 2020-04-29 at 18 19 01

After:
Screenshot 2020-04-29 at 18 18 52

Because I'm uniquing the names this means that for opss we'll still show the team name and the org - the same for other orgs where they've got multiple teams.

Screenshot 2020-04-29 at 18 21 18

### DIFF
--- a/psd-web/app/helpers/investigations/display_text_helper.rb
+++ b/psd-web/app/helpers/investigations/display_text_helper.rb
@@ -104,10 +104,10 @@ module Investigations::DisplayTextHelper
 
   # rubocop:disable Rails/OutputSafety
   def investigation_assignee(investigation, classes = "")
-    out = [investigation.assignable ? investigation.assignable.name.to_s : tag.div("Unassigned", class: classes)]
-    out << investigation.assignable.organisation.name if investigation.assignable&.organisation.present?
-    out.uniq!
-    out.compact!
+    out = [investigation.assignable ? h(investigation.assignable.name.to_s) : h("Unassigned")]
+    if investigation&.assignable&.organisation&.name != investigation&.assignable&.name
+      out << h(investigation.assignable.organisation.name)
+    end
     out.join("<br />").html_safe
   end
   # rubocop:enable Rails/OutputSafety
--- a/psd-web/app/helpers/investigations/display_text_helper.rb
+++ b/psd-web/app/helpers/investigations/display_text_helper.rb
@@ -105,8 +105,10 @@ module Investigations::DisplayTextHelper
   # rubocop:disable Rails/OutputSafety
   def investigation_assignee(investigation, classes = "")
     out = [investigation.assignable ? investigation.assignable.name.to_s : tag.div("Unassigned", class: classes)]
-    out << tag.div(investigation.assignable.organisation.name, class: classes) if investigation.assignable&.organisation.present?
-    out.join.html_safe
+    out << investigation.assignable.organisation.name if investigation.assignable&.organisation.present?
+    out.uniq!
+    out.compact!
+    out.join("<br />").html_safe
   end
   # rubocop:enable Rails/OutputSafety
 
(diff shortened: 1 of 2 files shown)


psd-web/app/helpers/investigations/display_text_helper.rb
```diff
@@ -105,8 +105,10 @@ def correspondence_should_be_hidden(result, source, investigation)
   # rubocop:disable Rails/OutputSafety
   def investigation_assignee(investigation, classes = "")
     out = [investigation.assignable ? investigation.assignable.name.to_s : tag.div("Unassigned", class: classes)]
-    out << tag.div(investigation.assignable.organisation.name, class: classes) if investigation.assignable&.organisation.present?
-    out.join.html_safe
+    out << investigation.assignable.organisation.name if investigation.assignable&.organisation.present?
+    out.uniq!
```
frankieroberto, line 109: Only adding the org name if it's not the same as the user name might make this a bit easier to follow?

eg

```ruby
if investigation&.assignable&.organisation&.name != investigation&.assignable&.name
  out << investigation.assignable.organisation.name
end
```

psd-web/app/helpers/investigations/display_text_helper.rb
```diff
@@ -105,8 +105,10 @@ def correspondence_should_be_hidden(result, source, investigation)
...
+    out.compact!
+    out.join("<br />").html_safe
```
frankieroberto, line 111: Removing the call to `tag.div` and outputting the result as `.html_safe` introduces a cross-site scripting vulnerability ([see example](https://psd-pr-554.london.cloudapps.digital/cases/2004-0049).

We could either wrap the names in an `h()` method call (which escapes the HTML), or use an existing helper like `simple_format()` which will sanitize the input by stripping tags.

psd-web/app/helpers/investigations/display_text_helper.rb
```diff
@@ -103,10 +103,12 @@ def correspondence_should_be_hidden(result, source, investigation)
   end
 
   # rubocop:disable Rails/OutputSafety
-  def investigation_assignee(investigation, classes = "")
-    out = [investigation.assignable ? investigation.assignable.name.to_s : tag.div("Unassigned", class: classes)]
-    out << tag.div(investigation.assignable.organisation.name, class: classes) if investigation.assignable&.organisation.present?
-    out.join.html_safe
+  def investigation_assignee(investigation)
+    out = [investigation.assignable ? h(investigation.assignable.name.to_s) : h("Unassigned")]
```
StupidCodeFactory, line 107:
```suggestion
    out = [investigation.assignable ? h(investigation.assignable.name.to_s) : "Unassigned"]
```
no need to escape "Unassigned" it doesn't contain malicious markup

psd-web/app/helpers/investigations/display_text_helper.rb
```diff
@@ -103,10 +103,12 @@ def correspondence_should_be_hidden(result, source, investigation)
...
-  def investigation_assignee(investigation, classes = "")
-    out = [investigation.assignable ? investigation.assignable.name.to_s : tag.div("Unassigned", class: classes)]
-    out << tag.div(investigation.assignable.organisation.name, class: classes) if investigation.assignable&.organisation.present?
-    out.join.html_safe
+  def investigation_assignee(investigation)
+    out = [investigation.assignable ? h(investigation.assignable.name.to_s) : "Unassigned"]
+    if investigation&.assignable&.organisation&.name != investigation&.assignable&.name
+      out << h(investigation.assignable.organisation.name)
+    end
+    out.join("<br>").html_safe
```
frankieroberto, line 111:
```suggestion
    out = [investigation.assignable ? h(investigation.assignable.name.to_s) : "Unassigned".html_safe]
    if investigation&.assignable&.organisation&.name != investigation&.assignable&.name
      out << h(investigation.assignable.organisation.name)
    end
    safe_join(out, "<br>".html_safe)
```

PR #558, opened 2020-04-30
Proposal: Speed up CI tests by using parallel jobs

scruti:
Coming from today's retro where concerns about our CI deploy & test speed were raised.

### For RSpec
Divided Feature and Unit tests so each group runs in parallel. Done by rspec tags.

### For System tests
Divided system tests files into different slices (configured to 3, can be changed by config). Run each slice of tests in parallel.

### For all the ruby workflows
Unified the gems cache between Minitest, System and RSpec workflows. There is no point on them generating/using different caches when the output is going to be the same independently from witch workflow generated it.

Also seems the System tests workflow cache usage was buggy and each system tests run was re-installing the gems without using the cache, taking more than 3m on it. This has been resolved.

## Output
With the current configuration and having cached gems, the full test suit takes **5-6 minutes.** 
That's around a **40-50% cut** over our current master test runs :raised_hands: 
Have to recognise that a considerable bit of the time saving has to do with solving the issue with the gems cache on System Tests, but parallel tests are definitely shortening the runs too.

## Thinks to consider:
- There is  that much we can achieve by adding parallelism. At some point the bottleneck is the workflow setup time more than the test run time.
- If we have big disparity between different parallel runs... we could consider solutions like Knapsackpro that offer automated balancing between the different jobs (and it is free for Open Source projects)

## Documentation
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix
 
## To Do:
- [x] Try "matrix" approach instead of duplicating workflows.
- [ ] If the proposal is approved: Modify Github required checks for our project to be merged.


frankieroberto:
@scruti I like this idea! Seems to be a good 50:50 split in the rspec tests too.

What are the pros/cons between having 2 jobs in the same workflow vs 2 different workflows (with one job each)?

scruti:
@frankieroberto didn't take much time to deep diving and probably there are considerations that I am not aware of, but from what I noticed:

### Pros for parallel jobs for a single workflow
- Single place for shared tests setup. Not having to maintain 2 files with 99% shared code and one small tweak different in each file. 
- Having different subsets of test jobs running for the same workflow seems more intuitive with the parallelisation as the jobs become "namespaced".
- Knapsack parallel test balancing tool is based on [Github Actions using this approach.](https://docs.knapsackpro.com/2019/github-actions-ci-config-for-ruby-on-rails-project-with-mysql-redis-elasticsearch-how-to-run-parallel-tests). So in the event of deciding to use the tool, would require minimal changes from this proposal. 

### Pros for different workflows
- I noticed that running in different Workflows for Rspec was taking a few seconds less than 2 jobs in the same workflow. But wasn't a big difference.
- Running different workflows advantage would be to allow tweaking the steps in some tests. Eg: Unit tests not needing to install Chrome driver before run.  But this maybe could be achieved by introducing a conditional clause in the step for the single workflow approach.

frankieroberto:
@scruti I'm not too familiar on the differences between jobs and workflows either...

My reading of https://help.github.com/en/actions/getting-started-with-github-actions/core-concepts-for-github-actions#job is that maybe jobs within the same workflow are run on the same virtual machine (even if run in parallel), so that maybe explains why it's a bit slower?

slorek:
Thanks for doing this. We could also consider splitting the RSpec suite into feature specs and everything else, to give it some logical separation.

scruti:
@slorek you have that approach in this commit: https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/558/commits/b2696e296b0cd84e397aad1eff5ce42ea9e68ca4 
I moved it back just because 99% of the workflow was exactly the same so felt an overkill. 
But not very strong opinionated on this, just went to with what it felt easier to maintain.

slorek:
@scruti I would lean to having a single workflow with multiple jobs (to reduce duplication in the def files and to organise the checks better), ideally in the first instance split with feature specs in one job and the rest in another, as then it is divided logically. We could consider splitting it up further as the suite grows.

PR #560, opened 2020-05-03
Bugfix filtering cases assigned to me

frankieroberto:
This fixes the bug whereby if you click the "Your cases" link on the homepage, you also see cases assigned to your team.

See https://trello.com/c/5LUydUFA/199-your-cases-link-filter-doesnt-filter-to-users-cases

This was caused because the filter builds the query by assuming that the team ID of your team should be added to the query unless the `assigned_to_team_0` param has the value of `unchecked`. But because the link from the homepage doesn't set this param (it just sets `/cases?assigned_to_me=checked`), the param had the value of nil instead of `unchecked`.

A private `set_default_assignee_filter` sets the params values to `unchecked` if they are missing, but the `assigned_to_team_0` was missing. Adding this fixes the bug.

I've added a feature spec which tests for this bug (if you checkout just this commit you can [see the test failing](https://github.com/UKGovernmentBEIS/beis-opss-psd/runs/641142391), with the second commit fixing the bug).

Also fixed the params in an unrelated spec which was relying upon the bug.

The code which builds the elasticsearch query is pretty hard to follow, and the resulting query is more complex than it often needs to be – however refactoring this is a task for the future. 😄

### DIFF
--- a/.github/workflows/overrides/bugfix-filtering-cases-assigned-to-me.env
+++ b/.github/workflows/overrides/bugfix-filtering-cases-assigned-to-me.env
@@ -1,1 +1,1 @@
-DB_VERSION=filtering-cases-bugfix
+DB_NAME=filtering-cases-bugfix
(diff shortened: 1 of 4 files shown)


frankieroberto:
PS I'm ignoring the comment from the codeclimate bot, as fixing that is much bigger job (which we should do at some point, but not now).

PR #567, opened 2020-05-04
Remove ability to be in multiple teams

slorek:
https://trello.com/c/uQgGqQDn/510-remove-ability-to-be-in-multiple-teams
https://trello.com/c/eYtohraA/544-refactor-display-names

## Description
This implements a change requiring users to belong to only one team, as opposed to none or many currently.

## Suggestion for reviewers
I suggest reviewing the changes as one set as there were a lot of bugs and merge conflicts I had to tackle. There are a lot of files changed, but only a few key changes to bear in mind:

* Moved invitations to a new REST controller and all the code from the model and controllers to form/service objects.
* Users must belong to only a single team and can never have zero or more than one.
* I have changed the behaviour of `display_name`, and there were existing bugs where `display_name` was called incorrectly - the fix was to pass the `viewer` context through various methods so that the caller can pass this through to `display_name`.

I have added comments throughout to help you understand the changes.

## Changes
* Refactored the messy controller/model/mailer code around invitations. `invitations` is now a resource on `teams` instead of member actions on the teams controller, in order to provide a cleaner RESTful interface and ensure the resend action is via `PUT` rather than `GET` (this means the URLs have changed for these actions). All business logic code has been moved out of the model and controller into form and service objects.
* Removed the `NewUser` model which was only used to validate email addresses on invitations, and replaced with a form object and moved most of the validations previously implemented in the controller into the form
* Converted the invitation view templates from Slim to ERb and fixed some dodgy inline styles and error message handling in the view
* Removed the redundant email template for inviting another user to your team
* Removed duplicate routes definitions
* Removed the "your teams" page
* Remove redundant functionality from Your account page
* Enforce one team per user in the database model and validations
* Fixed the `display_name` issues with displaying the org. name on `Team`/`User` (see [Trello ticket](https://trello.com/c/eYtohraA/544-refactor-display-names)) since these methods contained bugs and fixing them was not the desired behaviour (note: I did not rename `display_name` because it is used on both `Team` and `User` models, and `name_with_conditional_team` or similar makes no sense on a `Team`).
* Fixed the fixtures in the legacy test suite so that all users belong to a team (previously these had to be manually assigned in a test)
* Fixed a deprecation warning in the Rspec suite
* Changed existing specs to use the test queue adapter for consistency following feedback from @frankieroberto 

Note:

I added a `with_exception_errors` shared context to ensure error pages render instead of raising exceptions. For some reason the CI environment was rendering the pages _without_ this configuration, but locally it was working as expected. So there is some issue with the CI environment setting these configuration options unexpectedly, but I haven't figured out what that is yet.

## Important: Before merging
This change includes a migration which requires maintenance mode to be enabled as it will otherwise result in some user-visible errors for about 10 minutes. A backup of the database should also be taken prior to deployment.

## Post deploy steps
* Remove maintenance mode.
* Delete the email template ID `e3b2bbf5-3002-49fb-adb5-ad18e483c7e4` from Notify when merged

## Checklist:


psd-web/app/controllers/invitations_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  def create
+    @invite_user_to_team_form = InviteUserToTeamForm.new(params.require(:invite_user_to_team_form).permit(:email))
+    return render :new unless @invite_user_to_team_form.valid?
+
+    existing_user = User.find_by email: @invite_user_to_team_form.email
+
+    if existing_user
+      if existing_user.deleted?
+        @invite_user_to_team_form.errors.add(:email, :user_deleted)
```
slawosz, line 17: Is there any reason why these errors are outside form? If not, please change.

slorek: @slawosz it's still WIP, these will probably be in a service

psd-web/spec/features/registration_spec.rb
```diff
@@ -6,22 +6,21 @@
...
+    set_whitelisting_enabled(false)
     allow(Rails.application.config)
       .to receive(:secondary_authentication_enabled).and_return(true)
   end
 
   it "sending an invitation and registering" do
     sign_in(admin)
 
-    visit "/teams/#{team.id}/invite"
+    visit "/teams/#{team.id}/invitations/new"
```
frankieroberto, line 17: Could we avoid changing the URLs somehow, especially as `/teams/#{team.id}/invitations` doesn't do anything?

Maybe by using a singular resource, eg `resource :invite, only: [:show, :create, :update]`, with `update` being the resend method?

slorek: I think we should stick to standard conventions unless there is a compelling user need to retain the existing URL slug, i.e. lots of users have bookmarked that page.

psd-web/app/controllers/invitations_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+    @team = Team.find(params[:team_id])
+  end
+
+  def authorize_team_invite
+    authorize @team, :invite_user?
+  end
+
+  def current_operation
+    SecondaryAuthentication::INVITE_USER
+  end
```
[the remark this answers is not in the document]
slorek: The name `current_operation` is a bit ambiguous, we could consider changing it to be more explanatory. It is used by the secondary authentication `before_action`.

psd-web/app/controllers/invitations_controller.rb
```diff
@@ -0,0 +1,44 @@
+class InvitationsController < ApplicationController
+  before_action :get_team
```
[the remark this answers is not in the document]
slorek: I really have no opinion about this, it was a design decision made previously in the teams controller. But we should be consistent so I'll check which pattern is used most commonly.

psd-web/spec/support/exceptions.rb
```diff
@@ -0,0 +1,17 @@
...
+  before do
+    allow(Rails.application).to receive(:env_config).with(no_args).and_wrap_original do |m|
+      m.call.merge(
+        "consider_all_requests_local" => false,
+        "action_dispatch.show_exceptions" => true,
+        "action_dispatch.show_detailed_exceptions" => false
+      )
+    end
+  end
+end
```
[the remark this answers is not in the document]
slorek: 🤦 I didn't see this. I'l merge them.

psd-web/spec/services/invite_user_to_team_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+
+        expect(result.user.user_roles.length).to eq(1)
+        expect(result.user).to be_is_psd_user
+      end
+      # rubocop:enable RSpec/ExampleLength
+
+      it "enqueues the SendUserInvitationJob with the new user ID" do
+        result
+        expect(SendUserInvitationJob).to have_received(:perform_later).with(result.user.id, nil)
+      end
```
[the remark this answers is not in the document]
slorek: We should be consistent, I'll check which pattern is most commonly used.

psd-web/spec/requests/health_spec.rb
```diff
@@ -10,7 +10,7 @@
     it "checks health" do
       auth_headers = { "Authorization" => "Basic #{Base64.encode64('health:check')}" }
       get health_all_path, headers: auth_headers
-      expect(response).to be_success
+      expect(response).to be_successful
```
slorek, line 13: This fixes a deprecation warning.

psd-web/spec/services/invite_user_to_team_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+    before do
+      allow(SendUserInvitationJob).to receive(:perform_later)
+    end
+
+    context "when there is no existing user" do
+      let(:params) { { email: email, team: team } }
+      let(:new_token) { "new_token" }
+
+      before do
+        allow(SecureRandom).to receive(:hex).with(15).and_return(:new_token)
```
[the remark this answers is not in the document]
slorek: I considered using `kind_of(String)` (even checking length is tying it to the implementation) but I need to be sure it has changed and is also a sufficiently random token. I don't mind the test failing if we change the implementation in this case.

psd-web/spec/features/registration_spec.rb
edwardhorsford: I have a weak preference to keeping the existing url.

Like Frankie i'd prefer not to add stuff to our urls, especially where they don't actually do anything. `/invitations/` is not a thing we have. I'm of the view that path stems further up a url should generally work.

With that said, since this is inviting a specific team to a thing, I don't think it's something that would likely be bookmarked by people.

slorek: The existing URL structure is `/teams/<long ID>`, but since the user should now only ever be in one team it would make sense for the URLs to be `/team, `/team/invite`, `/team/resend-invite` or similar.

edwardhorsford: I think the existing structure should be kept - we originally intended that these pages could act as a contact / team index so other teams could know how to reach each other. We never got around to building that out, but `teams/foo` allows for that.

psd-web/app/models/audit_activity/document/destroy.rb
```diff
@@ -4,15 +4,17 @@ def self.from(document, investigation)
     super(document, investigation, title)
   end
 
-  def subtitle_slug
-    "#{attachment_type} deleted"
+  def restricted_title
+    "Document deleted"
```
edwardhorsford, line 8: Is this changing the text? I think we support more than just `document` don't we?

slorek: I just moved things around a bit, the code is unchanged.

psd-web/app/models/activity.rb
```diff
@@ -37,20 +35,35 @@ def restricted_title
...
+  def email_update_text(viewing_user = nil); end
+
+  def email_subject_text
+    "#{investigation.case_type.upcase_first} updated"
+  end
+
+private
+
+  def pretty_date_stamp
+    created_at.strftime("%d %B %Y")
```
StupidCodeFactory, line 47:
```suggestion
     created_at.to_s(:govuk)
```

slorek: I've not changed this from the old code, just moved around a bit. Suggest doing this in a separate PR if we want to change it.

psd-web/test/models/notification_test.rb
```diff
@@ -114,12 +114,10 @@ class NotificationTest < ActiveSupport::TestCase
   end
 
   test "should notify everyone in team that is the case owner" do
-    users(:southampton).teams << teams(:southampton)
-    users(:southampton_steve).teams << teams(:southampton)
     User.current = users(:southampton)
     mock_investigation_updated(who_will_be_notified: teams(:southampton).users.map(&:email))
     @investigation.update(owner: teams(:southampton))
-    assert_equal @number_of_notifications, 2
+    assert_equal @number_of_notifications, 4
```
slorek, line 120: This number changed as a result of correcting the fixtures so that all users belong to a team.

psd-web/test/models/notification_test.rb
```diff
@@ -94,7 +94,7 @@ class NotificationTest < ActiveSupport::TestCase
 
     mock_investigation_updated(who_will_be_notified: expected_recipients)
     @investigation.update!(owner: users(:southampton))
-    assert_equal @number_of_notifications, 2
+    assert_equal @number_of_notifications, 3
```
slorek, line 97: This was actually a bug (or a deliberate oversight) in the test because it should always have been three - this is even stated 2 lines above. Correcting the fixtures exposed the error

psd-web/spec/requests/user_requests_password_reset_spec.rb
```diff
@@ -26,9 +22,11 @@
       expect(response).to redirect_to(check_your_email_path)
     end
 
-    it "queues the resend invitation job" do
-      request_password_reset
-      expect(SendUserInvitationJob).to have_received(:perform_later).with(user.id, nil)
+    it "queues the resend invitation job", :with_test_queue_adapter, :aggregate_failures do
```
slorek, line 25: This change is not directly related, but changed for consistency following feedback from @frankieroberto.

psd-web/spec/models/user_spec.rb
```diff
@@ -125,144 +125,15 @@
     end
   end
 
-  describe ".create_and_send_invite!" do
```
slorek, line 128: All of this functionality was moved to the service object.

psd-web/spec/models/audit_activity/investigation/update_owner_spec.rb
```diff
@@ -6,18 +6,14 @@
   let(:user) { create(:user).decorate }
   let(:investigation) { create(:enquiry, owner: user, owner_rationale: "Test owner") }
 
-  around do |ex|
-    User.current = user
-    ex.run
-    User.current = nil
-  end
+  before { User.current = user }
```
slorek, line 9: `User.current` is reset after each test already.

psd-web/spec/models/audit_activity/investigation/update_coronavirus_status_spec.rb
```diff
@@ -9,11 +9,11 @@
   before { User.current = user }
 
   describe ".from" do
-    it "creates an audit activity" do
+    it "creates an audit activity", :aggregate_failures do
       expect(audit_activity).to have_attributes(title: "Status updated: not coronavirus related",
                                                 body: "The case is not related to the coronavirus outbreak.",
-                                                email_update_text: "#{investigation.case_type.capitalize} #{investigation.pretty_id} is not related to the coronavirus outbreak. This status was updated by #{user.display_name}.",
                                                 email_subject_text: "Coronavirus status updated on #{investigation.case_type.downcase}")
+      expect(audit_activity.email_update_text(user)).to eq("#{investigation.case_type.capitalize} #{investigation.pretty_id} is not related to the coronavirus outbreak. This status was updated by #{user.name}.")
```
slorek, line 16: `email_update_text` now requires a parameter to ensure the correct name is displayed, so can't be checked using `have_attributes`.

psd-web/spec/models/alert_spec.rb
```diff
@@ -4,15 +4,16 @@
...
 
-      alert.send_alert_email
-
-      expect(SendAlertJob).to have_received(:perform_later).with([activated_user.email], subject_text: "test", body_text: "test")
+    it "sends an email to activated users only", :with_test_queue_adapter, :aggregate_failures do
+      expect { alert.send_alert_email }.to have_enqueued_job(SendAlertJob).at(:no_wait).on_queue("psd").with do |recipients, subject_text, body_text|
+        expect(recipients).to eq [activated_user.email]
+        expect(subject_text).to eq "test"
+        expect(body_text).to eq "test"
+      end
```
slorek, line 16: These changes are not directly related, but changed for consistency following feedback from @frankieroberto.

psd-web/app/services/create_organisation_with_team_and_admin_user.rb
```diff
@@ -12,7 +12,8 @@ def call
       context.user = context.team.users.create!(
         email: context.admin_email,
         organisation: context.org,
-        skip_password_validation: true
+        skip_password_validation: true,
+        team: context.team
```
slorek, line 16: I had to add this to avoid an  exception despite already using `context.team.users.create`.

psd-web/app/models/user_source.rb
```diff
@@ -1,8 +1,8 @@
 class UserSource < Source
   belongs_to :user, optional: true
 
-  def show
-    user.present? ? user.decorate.display_name : "anonymous"
+  def show(viewer = nil)
```
slorek, line 4: We need to pass the context of the `viewer` through here so that `display_name` works properly. We could do with refactoring this whole `source` model but that's for another time.

psd-web/app/models/user.rb
```diff
@@ -46,36 +47,6 @@ def self.not_deleted
...
-    SendUserInvitationJob.perform_later(user.id, inviting_user.id)
-  end
-
-  def self.resend_invite(email_address, inviting_user)
-    user = find_user_within_teams_with_email!(email: email_address, teams: inviting_user.teams)
-
-    user.update! invitation_token: user.invitation_token || SecureRandom.hex(15), invited_at: Time.current
-
-    SendUserInvitationJob.perform_later(user.id, inviting_user.id)
-  end
```
slorek, line 77: All this code was moved to a service.

psd-web/app/models/team.rb
```diff
@@ -11,10 +9,8 @@ def self.all_with_organisation
     all.includes(:organisation)
   end
 
-  def display_name(ignore_visibility_restrictions: false, current_user: User.current)
-    return name if (current_user && (current_user.organisation_id == organisation_id)) || ignore_visibility_restrictions
-
-    organisation.name
+  def display_name(*)
```
slorek, line 12: We no longer want to hide the team name, even if the viewer is in another organisation. We still need to accept arguments because the corresponding `UserDecorator` methods require one.

psd-web/app/models/activity.rb
```diff
@@ -37,20 +35,35 @@ def restricted_title
     # where necessary should be implemented by subclasses
   end
 
+  def email_update_text(viewer = nil); end
```
slorek, line 38: We need to add the `viewer` parameter so that subclasses can correctly call `display_name` when relevant.

psd-web/app/decorators/user_decorator.rb
```diff
@@ -12,15 +12,10 @@ def full_name
...
-    suffix = if (ignore_visibility_restrictions || (organisation_id == other_user&.organisation_id)) && teams.any?
-               "(#{team_names})"
-             else
-               "(#{organisation.name})"
-             end
-    suffix << deleted_suffix
-
-    "#{name} #{suffix}"
+  # viewer could be a Team or User
+  def display_name(viewer: User.current)
```
slorek, line 16: We no longer want to show the organisation name, only the team name if the `viewer` is in another team. There were a few bugs where `display_name` was called without a `viewer` and so the wrong name was returned - these are fixed by ensuring `viewer` can be passed in through various methods seen in this PR.

psd-web/app/decorators/team_decorator.rb
```diff
@@ -5,9 +5,7 @@ def owner_short_name(*)
     display_name
   end
 
-  def display_name(ignore_visibility_restrictions: false, current_user: User.current)
-    return name if (current_user && (current_user.organisation_id == organisation_id)) || ignore_visibility_restrictions
-
-    organisation.name
+  def display_name(*)
```
slorek, line 8: We only ever want to return the team name now. We need to accept arguments because the caller doesn't always know if it is a `Team` or a `User`.

psd-web/app/services/invite_user_to_team.rb
```diff
@@ -0,0 +1,42 @@
...
+    context.user ||= find_or_create_user
+
+    send_invite
+  end
+
+private
+
+  def find_or_create_user
+    existing_user = User.find_by email: email
+    existing_user&.team == team ? existing_user : create_user
```
slorek, line 19: This is designed to throw an exception if there is an existing user on another team.

nasirkhanbeis:
can we please change this in url to singular as team?  
image

psd-web/test/system/investigation_assignee_test.rb
```diff
@@ -1,49 +0,0 @@
-require "application_system_test_case"
-
-class InvestigationAssigneeTest < ApplicationSystemTestCase
```
slorek, line 3: These tests were faulty depending on whether they were run in isolation or as part of the full suite, and so moved to the corresponding feature spec.

slorek:
> can we please change this in url to singular as team?
> image

I discussed this with @edwardhorsford and also bearing in mind the forthcoming platform admin feature which could require a user to be able to view multiple teams I decided to keep the URL structure the same for the team page.

psd-web/db/migrate/20200507145004_drop_teams_users.rb
```diff
@@ -0,0 +1,37 @@
+class DropTeamsUsers < ActiveRecord::Migration[5.2]
+  def up
+    # Check that all users have one team
+    safety_assured do
+      query = execute "SELECT count(*) FROM users WHERE (SELECT count(*) AS teams_count FROM teams_users WHERE teams_users.user_id=users.id) != 1"
+      raise "Users with zero or more than one team exist - clean the data first" if query.first["count"] != 0
+    end
+
+    add_column :users, :team_id, :uuid, index: true
```
frankieroberto, line 9: Add a foreign key here too?

```suggestion
    add_column :users, :team_id, :uuid, index: true
    add_foreign_key :users, :teams
```

psd-web/app/services/invite_user_to_team.rb
```diff
@@ -0,0 +1,42 @@
...
+      team: team
+    )
+
+    user.user_roles.create!(name: "psd_user") # TODO: remove this once we’ve updated the application to no longer depend upon this role.
+    user.user_roles.create!(name: "opss_user") if inviting_user&.is_opss?
+    user
+  end
+
+  def send_invite
+    if !user.invitation_token || (user.invited_at < 1.hour.ago)
```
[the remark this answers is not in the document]
slorek: I think I added this to save an additional unnecessary database query. The way I structured this the query would be executed on every call even if it wasn't necessary. So this at least assures that the query is worthwhile, I felt that one hour wouldn't make a material difference to the expiry.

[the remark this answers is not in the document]
frankieroberto: Oh, I misread the condition as "less than one hour ago" when it's actually "more than one hour ago". 🤦

nasirkhanbeis:
@slorek 
Can we display team names along on changing the owner on the case pls in below template? 
image

slorek:
> @slorek
> Can we display team names along on changing the owner on the case pls in below template?
> image

This will now only happen if the user is in another team, as per the Trello card.

psd-web/app/decorators/team_decorator.rb
edwardhorsford: Shouldn't this just return `name`? The team name would never be a user.

nasirkhanbeis:
@slorek  here display name should also show team name when adding a team of diff org
image

nasirkhanbeis:
@slorek 
Tested pretty much all possible scenarios in review app and all places for display name, looks good as expected. 
Once we fix above issue should be good.

slorek:
> @slorek here display name should also show team/org name when adding a team of diff org
> image

This appears to be an existing bug unconnected with these changes, see https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/mailers/notify_mailer.rb#L140 - the `name` method is called instead of `display_name`. Suggest fixing separately to keep this PR small.

psd-web/app/decorators/team_decorator.rb
slorek: That's exactly what it does :)

PR #571, opened 2020-05-05
Add ability to remove team from case

slawosz:
Adds ability to remove access from a case, as per https://trello.com/c/CdqAARmz/433-add-ability-to-remove-a-team-from-a-case
Currently removes `Collaborator` record from DB.

Link `Change` on access list page.
Screenshot from 2020-05-07 11-27-40
Form with validations.
Screenshot from 2020-05-07 11-28-03


psd-web/spec/forms/edit_investigation_collaborator_form_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+          it "sends email to users" do
+            form.save
+
+            email = delivered_emails.last
+            expect(email.recipient).to eq(team_user.email)
+          end
+        end
+
+        # rubocop:disable RSpec/ExampleLength
+        it "creates activity entry" do
```
scruti, line 73: you can add `, :aggregate_failures` here if you want to save some lines in the spec.

psd-web/app/models/audit_activity/investigation/team_deleted.rb
```diff
@@ -0,0 +1,5 @@
+class AuditActivity::Investigation::TeamDeleted < AuditActivity::Investigation::Base
+  def subtitle
+    "Team deleted by #{source&.show}, #{pretty_date_stamp}"
```
edwardhorsford, line 3:
```suggestion
    "Team removed by #{source&.show}, #{pretty_date_stamp}"
```

psd-web/app/views/collaborators/edit.html.erb
```diff
@@ -0,0 +1,90 @@
...
+      <p class="govuk-body">We’ll email <%= @team.name %> to let them know that their permission level has been changed.
+</p>
+
+      <% message_html = capture do %>
+        <%= render "form_components/govuk_textarea",
+          form: form,
+          key: :message,
+          id: "message",
+          label: {
+            text: "Message to the team"
```
edwardhorsford, line 53: please include team name here:
`Message to Birmingham Council`, etc

psd-web/app/views/collaborators/edit.html.erb
```diff
@@ -0,0 +1,90 @@
...
+        %>
+      <% end %>
+
+      <%= render "form_components/govuk_radios",
+        form: form,
+        key: :include_message,
+        classes: "",
+        fieldset: {
+          legend: {
+            text: "Do you want to include information?"
```
edwardhorsford, line 67:
```suggestion
            text: "Do you want to include more information?"
```

psd-web/app/views/collaborators/index.html.erb
```diff
@@ -23,19 +23,27 @@
...
           <th scope="col" class="govuk-table__header">Permission level</th>
+          <th scope="col" class="govuk-table__header"></th>
         </tr>
       </thead>
       <tbody class="govuk-table__body">
         <% if @investigation.assignee_team %>
           <tr class="govuk-table__row">
             <th scope="row" class="govuk-table__header govuk-!-font-weight-regular"><%= @investigation.assignee_team.name %></th>
             <td class="govuk-table__cell"><span class="hmcts-badge hmcts-badge--grey">Case owner</span></td>
+            <td class="govuk-table__cell"><span class="hmcts-badge hmcts-badge--grey">Assignee</span></td>
```
edwardhorsford, line 34: Please remove.

psd-web/app/views/collaborators/index.html.erb
```diff
@@ -23,19 +23,27 @@
...
+            <td class="govuk-table__cell"></td>
           </tr>
         <% end %>
         <% @teams.each do |team| %>
           <tr class="govuk-table__row">
             <th scope="row" class="govuk-table__header govuk-!-font-weight-regular"><%= team.name %></th>
             <td class="govuk-table__cell">Edit full case</td>
+            <td class="govuk-table__cell">
+              <% if policy(@investigation).add_collaborators? %>
+                <%= link_to "Change", edit_investigation_collaborator_path(@investigation, team) %>
```
edwardhorsford, line 44: Please change: should have hidden text with team name - see `link_helper.rb`

nasirkhanbeis:
email notification shows some classname variables" investigation::allegation" and also ID format 
is not correct
Should read as "Your team has been removed from allegation"{case title} 
image

edwardhorsford:
Spotted by @nasirkhanbeis, the link for the error for the first radio does not anchor to them. @frankieroberto can probably advise on how to do this.

Screenshot 2020-05-06 at 10 06 19

psd-web/app/controllers/collaborators_controller.rb
```diff
@@ -35,11 +34,36 @@ def create
...
+    @collaborator = @investigation.collaborators.find_by! team_id: @team.id
+    @edit_form = EditInvestigationCollaboratorForm.new(permission_level: EditInvestigationCollaboratorForm::PERMISSION_LEVEL_EDIT)
+  end
+
+  def update
+    authorize @investigation, :add_collaborators?
+
+    params.permit!
+    @team = Team.find(params[:id])
+    @edit_form = EditInvestigationCollaboratorForm.new(params[:edit_investigation_collaborator_form]
```
StupidCodeFactory, line 50: Please change: `params.require(:edit_investigation_collaborator_form)`

psd-web/app/controllers/collaborators_controller.rb
```diff
@@ -35,11 +34,36 @@ def create
...
+
+    @team = Team.find(params[:id])
+    @collaborator = @investigation.collaborators.find_by! team_id: @team.id
+    @edit_form = EditInvestigationCollaboratorForm.new(permission_level: EditInvestigationCollaboratorForm::PERMISSION_LEVEL_EDIT)
+  end
+
+  def update
+    authorize @investigation, :add_collaborators?
+
+    params.permit!
```
StupidCodeFactory, line 48: Please change: you do not need to allow all of the params. Just need the necessary one.

psd-web/app/forms/edit_investigation_collaborator_form.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  validates_presence_of :permission_level
+  validate :select_different_permission_level
+  validates :include_message, inclusion: { in: [true, false] }
+  validates_presence_of :message,
+                        if: -> { include_message }
+
+  def save
+    if valid?
+      collaborator.delete
```
StupidCodeFactory, line 17: what happens if this fails?

psd-web/app/forms/edit_investigation_collaborator_form.rb
```diff
@@ -0,0 +1,73 @@
+class EditInvestigationCollaboratorForm
+  PERMISSION_LEVEL_EDIT = "edit".freeze
+  PERMISSION_LEVEL_DELETE = "delete".freeze
+  include ActiveModel::Model
+
+  attr_accessor :permission_level, :message, :investigation, :team, :user
+  attr_reader :include_message
```
StupidCodeFactory, line 7: Please change: so in all other forms we added 
```
include ActiveModel::Model
include ActiveModel::Attributes
```

you can then do 
```
attribute : permission_level
attribute  :message
attribute :include_message, :boolean
```

and remove this odd accessor method line 26. Let rails do that casting for you, it does it better ;)

psd-web/app/forms/edit_investigation_collaborator_form.rb
```diff
@@ -0,0 +1,73 @@
...
+  def collaborator
+    investigation.collaborators.find_by!(team_id: team.id)
+  end
+
+  def schedule_delete_emails
+    emails = if team.team_recipient_email.present?
+               [team.team_recipient_email]
+             else
+               team.users.active.pluck(:email)
+             end
```
StupidCodeFactory, line 45: Suggestion: you could extract this bit to its own method, it would read much better and you could remove the `if/else` with an early return

psd-web/app/forms/edit_investigation_collaborator_form.rb
```diff
@@ -0,0 +1,73 @@
...
+  validate :select_different_permission_level
+  validates :include_message, inclusion: { in: [true, false] }
+  validates_presence_of :message,
+                        if: -> { include_message }
+
+  def save
+    if valid?
+      collaborator.delete
+      schedule_delete_emails
+      add_deletion_activity
```
StupidCodeFactory, line 19: 1. this method raise an exception, it customary to add and `!` at the end of methods that raise exception (and/or have side effects on a state)
2. I wonder about the order of this because of potential failure, I'd would probably schedule the email last, because if the audit is not created. you might want to rollback the collaborator deletion as well.

This class could leave the database in a weird state. It's worth addressing that, is it not?

psd-web/app/forms/edit_investigation_collaborator_form.rb
```diff
@@ -0,0 +1,73 @@
...
+  def schedule_delete_email(emails)
+    emails.each do |email|
+      NotifyMailer.team_deleted_from_case_email(email_payload, to_email: email).deliver_later
+    end
+  end
+
+  def email_payload
+    { permission_level: permission_level, include_message: include_message, message: message, investigation: investigation, team: team, user: user }
+  end
+end
```
StupidCodeFactory, line 73: Opinion: This is a bit of funny class half a form object and half a service object. I would have preferred a two classes, the form handling validation and a service class doing the db stuff and sending email once everything is saved ok.

psd-web/app/views/collaborators/edit.html.erb
```diff
@@ -0,0 +1,95 @@
...
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+
+    <%= error_summary @edit_form.errors %>
+
+    <h1 class="govuk-heading-l"><%= title %></h1>
+
+    <%= form_with model: @edit_form, url: investigation_collaborator_path(@investigation, @team), method: :put do |form| %>
```
StupidCodeFactory, line 15: I think you are missing the errors on the each fields of the form. @frankieroberto am I right?

psd-web/spec/forms/edit_investigation_collaborator_form_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+
+  describe "#save" do
+    context "when deleting" do
+      context "when successful" do
+        it "removes collaborator record" do
+          expect { form.save }.to change(Collaborator, :count).from(1).to(0)
+        end
+
+        it "returns true" do
+          expect(form.save).to be_truthy
```
StupidCodeFactory, line 44:
```suggestion
          expect(form.save).to be true
```
We return either true or false

psd-web/spec/forms/edit_investigation_collaborator_form_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+        end
+
+        # rubocop:disable RSpec/ExampleLength
+        it "sends email" do
```
StupidCodeFactory, line 48: Suggestion: I think if you do 
```ruby
it "sends email", :aggregate_failures do
```
you can remove the block, which will reduce the spec by 2 lines and then you can remove the rubocop:disable comment

psd-web/spec/forms/edit_investigation_collaborator_form_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+          end
+        end
+
+        # rubocop:disable RSpec/ExampleLength
+        it "creates activity entry" do
+          form.save
+
+          last_added_activity = investigation.activities.reload.order(:created_at).first
+          aggregate_failures do
+            expect(last_added_activity).to be_a(AuditActivity::Investigation::TeamDeleted)
```
StupidCodeFactory, line 78: you should remove this and line 76:
```ruby
investigation.activities.order(created_at: :desc).find_by!(type: "AuditActivity::Investigation::TeamDeleted")
```
you would not have to test for this, and also you forgot to order by `DESC` as well

psd-web/app/forms/edit_investigation_collaborator_form.rb
slawosz: Well, I think its perfectly fine to have form object that has a heavy lifting, I am quite used to that, adding second class here doesn't make any sense now, as the sole purpose of this class would be organizing/orchiestrating some code.

psd-web/app/forms/edit_investigation_collaborator_form.rb
slawosz: 1. I decided not to add `!` as an exception here is not something we want to catch. In normal circumstances, it won't be thrown, as we are filtering params in caller (controller). It might be gray are, but in our case, its good enough.
2. I will rearrange, thanks.

psd-web/app/controllers/collaborators_controller.rb
```diff
@@ -35,11 +34,36 @@ def create
     end
   end
 
+  def edit
+    authorize @investigation, :add_collaborators?
+
+    @team = Team.find(params[:id])
+    @collaborator = @investigation.collaborators.find_by! team_id: @team.id
```
frankieroberto, line 41: suggestion: use the collaborator id in the path, and then look it up using:

```suggestion
    @collaborator = Collaborator.find(params[:id])
    @team = @collaborator.team
```

psd-web/app/controllers/collaborators_controller.rb
```diff
@@ -35,11 +34,36 @@ def create
...
+    @edit_form = EditInvestigationCollaboratorForm.new(permission_level: EditInvestigationCollaboratorForm::PERMISSION_LEVEL_EDIT)
+  end
+
+  def update
+    authorize @investigation, :add_collaborators?
```
frankieroberto, line 46: rename `add_collaborators` to `manage_collaborators` if it now covers update/deletion?

psd-web/app/forms/edit_investigation_collaborator_form.rb
slawosz: Throws exception, as below.

psd-web/spec/forms/edit_investigation_collaborator_form_spec.rb
slawosz: Its unclear, could you make a suggestion please?

psd-web/app/controllers/collaborators_controller.rb
```diff
@@ -35,11 +34,36 @@ def create
...
+      render "edit"
+    end
+  end
+
 private
 
+  # rubocop:disable Naming/MemoizedInstanceVariableName
   def find_investigation_from_params
-    Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
+    @investigation ||= Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
```
frankieroberto, line 64: ha! I deliberately didn't set the instance variable in the private method as I find that doing that from a method called from a `before_action` hooks can make the code harder to follow.  Debatable though!

psd-web/app/forms/edit_investigation_collaborator_form.rb
```diff
@@ -0,0 +1,73 @@
...
+               team.users.active.pluck(:email)
+             end
+    schedule_delete_email(emails)
+  end
+
+  def add_deletion_activity
+    AuditActivity::Investigation::TeamDeleted.create!(
+      source: UserSource.new(user: user),
+      investigation: investigation,
+      title: "#{team.name} removed from #{investigation.case_type.downcase}",
```
frankieroberto, line 53: suggestion: move string to locale file?

psd-web/app/forms/edit_investigation_collaborator_form.rb
StupidCodeFactory: 1. `delete` does not throw and exception
2. delete bypasses any before/after filter.

Please change for: `destroy!`

psd-web/app/controllers/collaborators_controller.rb
StupidCodeFactory: I very much agree with @frankieroberto

psd-web/app/forms/edit_investigation_collaborator_form.rb
StupidCodeFactory: it still is a useful convention to add `!` to method that raises exception whether you want to catch them or not
Please change: `add_deletion_activity!`

psd-web/spec/forms/edit_investigation_collaborator_form_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+            email = delivered_emails.last
+            expect(email.recipient).to eq(team_user.email)
+          end
+        end
+
+        # rubocop:disable RSpec/ExampleLength
+        it "creates activity entry" do
+          form.save
+
+          last_added_activity = investigation.activities.reload.order(:created_at).first
```
StupidCodeFactory, line 76:
```suggestion
          last_added_activity = investigation.activities.reload.order(created_at: :desc).find_by!(type: "AuditActivity::Investigation::TeamDeleted")
```
and then remove line 78

psd-web/spec/forms/edit_investigation_collaborator_form_spec.rb
StupidCodeFactory: sure! added a suggestion line 76

psd-web/app/controllers/collaborators_controller.rb
slawosz: Its quite common practice in controller, its 'main' object, so I don't see why not.

psd-web/app/controllers/collaborators_controller.rb
slawosz: Is there any value over current approach?

frankieroberto: No. And actually your way checks that the two id params match, so is probably better.

edwardhorsford:
@slawosz can you update the PR title and description with some more details and maybe a screenshot or two?

psd-web/spec/forms/edit_investigation_collaborator_form_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+          let(:params_team) { create(:team) }
+
+          it "raises exception" do
+            expect { form.save }.to raise_error(ActiveRecord::RecordNotFound)
+          end
+        end
+
+        shared_examples "unsuccessful save" do
+          it "returns false" do
+            expect(form.save).to be_falsey
```
StupidCodeFactory, line 88:
```suggestion
            expect(form.save).to be false
```

psd-web/spec/features/case_permissions_management_spec.rb
```diff
@@ -87,6 +91,70 @@
     expect(page).to have_text("Thanks for collaborating on this case with us.")
   end
 
+  scenario "Remove a team from a case (with validation errors)" do
+    # create collaborator
```
StupidCodeFactory, line 95:
```suggestion
```

PR #573, opened 2020-05-06
Allow users to request a new 2fa security code

scruti:
[Trello ticket](https://trello.com/c/Z3k1z1Oq/343-pages-for-not-received-a-code)

In case users didn't receive the 2fa code when enquired for one, they can request a new code to be sent in order to proceed with 2fa authentication.

This is done through introducing a new page with an action that triggers a new secondary authentication request (and sends a new sms). 

The behaviour for requesting secondary authentication was enforcing a redirection back to the path where 2fa was triggered from, and this steps introduces a middleman.

Before
`Original path` --> `Request 2FA` --> `back to original path`
After
`Original path` --> `Request 2FA` --> `Page for resend code` --> `Request 2FA` --> `back to original path`

In order to achieve this, we need to be able to carry the original redirection from the first `Request 2FA` to the second.  I extended `request_secondary_authentication` method so the redirection address can be injected when necessary, while defaulting to the current fullpath.

Also had to forward the `current_operation` set by the original action that required 2FA, so the 2FA expiry time we check against when re-submitting the code is the same expiry time we originally had.

## Checklist:


scruti:
[Pages for 'not received a code'](https://trello.com/c/Z3k1z1Oq/343-pages-for-not-received-a-code)

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## 2020-05-06
+- Allow to request a new secondary authentication code when secondary authentication is required and code was not received via sms.
```
frankieroberto, line 5: Suggested rewording:

```suggestion
- Allow users to request a new security code when the code was not received via SMS.
```

psd-web/app/controllers/secondary_authentications/resend_code_controller.rb
```diff
@@ -0,0 +1,20 @@
+module SecondaryAuthentications
+  class ResendCodeController < ApplicationController
+    skip_before_action :authenticate_user!
+    skip_before_action :set_current_user
+    skip_before_action :require_secondary_authentication
+    skip_before_action :set_raven_context
+    skip_before_action :authorize_user
+    skip_before_action :has_accepted_declaration
+    skip_before_action :has_viewed_introduction
+    skip_before_action :set_cache_headers
```
frankieroberto, line 10: combine these?

psd-web/spec/features/sign_in_spec.rb
```diff
@@ -66,6 +72,35 @@ def expect_incorrect_email_or_password
...
+    click_on "Resend security code"
+
+    second_2fa_code = otp_code
+    expect(second_2fa_code).not_to eq first_2fa_code
+    expect_user_to_have_received_sms_code(second_2fa_code)
+
+    expect(page).to have_css("h1", text: "Check your phone")
+
+    fill_in "Enter security code", with: otp_code
+    click_on "Continue"
```
frankieroberto, line 98: Might be worth changing this to also assert that it's a button being clicked (`clicked_on` works for both buttons and links):

```suggestion
    click_button "Continue"
```

psd-web/spec/features/sign_in_spec.rb
```diff
@@ -66,6 +72,35 @@ def expect_incorrect_email_or_password
...
+
+    click_link "Not received a text message?"
+
+    expect(page).to have_css("h1", text: "Resend security code")
+    expect(page).to have_current_path("/text-not-received")
+
+    click_on "Resend security code"
+
+    second_2fa_code = otp_code
+    expect(second_2fa_code).not_to eq first_2fa_code
```
frankieroberto, line 92: Isn't there a _small_ chance that the user gets the same code twice? 😉

psd-web/config/routes.rb
```diff
@@ -33,6 +33,9 @@
   get "two-factor", to: "secondary_authentications#new", as: :new_secondary_authentication
   post "two-factor", to: "secondary_authentications#create", as: :secondary_authentication
 
+  get "text-not-received", to: "secondary_authentications/resend_code#new", as: :new_resend_secondary_authentication_code
```
frankieroberto, line 36: Suggestion: this might be better a (singular) "resource" route, albeit with a custom path?  That would map the view to the "show" method  and the form action to the "update" method?

```ruby
resource :resend_secondary_authentication, path: "text-not-received", controller: "SecondaryAuthentications ::ResendCodeController", only: [:show, :update]
```

CHANGELOG.md
scruti: Thanks for this!

psd-web/spec/features/sign_in_spec.rb
scruti: Truth, going to re-think how to test this.

psd-web/config/routes.rb
scruti: Mapping odd actions/triggers to REST is hard :/
I like the `new/create` for initiating/triggering the action, as done in `two-factor`. But at the same time I don't want to show to the user a `text-not-received/new`, hence not declaring it as a resource.
For the show/update... even when we are underlying updating the OTP code of an user I see this more like a trigger than a resource update. And the show does not match the "display --> confirm --> trigger" flow that we need (and I think is better matched by new/create).

Going to stick to what is already done until now and if we re-think how we map these actions we can change them all together.

slawosz: I like your approach here @scruti

nasirkhanbeis: seems not working on invite user 2fa journey- resend code doesn't do anything. can you pls double check @scruti ?

nasirkhanbeis:
seems not working on invite user 2fa journey- resend code doesn't do anything. can you pls double check @scruti ?

nasirkhanbeis:
> seems not working on invite user 2fa journey- resend code doesn't do anything. can you pls double check @scruti ?

suggestion: will be good to add a test around invite journey - resend 2fa code

scruti:
Cause: we're not considering the specific operation when triggering re-send code and the check for the "need or not 2FA code?" based on the operation expiry time falls into a default.
Going to fix this so it considers the expiry time for specific operations.

PR #574, opened 2020-05-06
Don't show document links if the user shouldn't see them

edwardhorsford:
This removes the attachment links if the viewing user shouldn't be able to view the document.

Note: we'll be reworking this section shortly with wider permissions work, so this is a short term fix.

Before:
Screenshot 2020-05-06 at 16 49 32

After:
Screenshot 2020-05-06 at 16 49 06


edwardhorsford:
[Document view links let users see files they shouldn't be able to](https://trello.com/c/dFzTsslS/536-document-view-links-let-users-see-files-they-shouldnt-be-able-to)

psd-web/app/views/documents/_document_card.html.slim
```diff
@@ -18,14 +18,14 @@ ruby:
...
-          "#{title}, opens in a new window", document,
-          class: "govuk-link app-block-link govuk-!-margin-right-3",
-            target: '_blank', rel: 'noopener'
-      = link_with_hidden_text_to "Edit #{image_document_text(document)}",
-          title, edit_associated_document_path(parent, document),
-          class: "govuk-link app-block-link govuk-!-margin-right-3"
-      = link_with_hidden_text_to "Remove #{image_document_text(document)}",
-          title, remove_associated_document_path(parent, document),
-          class: "govuk-link app-block-link"
+      - if safe
```
StupidCodeFactory, line 21: is this intentional to now put this in the else branch after the description?

psd-web/spec/features/attachment_visibility.rb
```diff
@@ -0,0 +1,19 @@
+require "rails_helper"
+
+RSpec.feature "Adding an attachment to a case", :with_stubbed_elasticsearch, :with_stubbed_antivirus, :with_stubbed_mailer, type: :feature do
+  let(:user) { create(:user, :activated, has_viewed_introduction: true) }
+  let(:other_user) { create(:user, :activated, has_viewed_introduction: true) }
+  let(:investigation) { create(:allegation, :with_antivirus_checked_document, document_has_consumer_info: true, owner: user) }
+
+  scenario "Displays file link for user" do
```
edwardhorsford, line 8:
```suggestion
  scenario "Displays link to view the file" do
```

psd-web/spec/features/attachment_visibility.rb
```diff
@@ -0,0 +1,19 @@
...
+    sign_in(user)
+    visit "/cases/#{investigation.pretty_id}/attachments"
+    expect(page).to have_link("View TXT")
+  end
+
+  scenario "Does not displays link to restricted case" do
```
edwardhorsford, line 14:
```suggestion
  scenario "Does not link to the file" do
```

psd-web/spec/features/attachment_visibility.rb
```diff
@@ -0,0 +1,19 @@
+require "rails_helper"
+
+RSpec.feature "Adding an attachment to a case", :with_stubbed_elasticsearch, :with_stubbed_antivirus, :with_stubbed_mailer, type: :feature do
```
edwardhorsford, line 3: This test might not be ideal. It works - but in the service it's not possible to add an attachment its own and mark it as containing consumer info.

To get to this scenario in a real case the user would need to add an item of correspondence, mark the correspondence as containing consumer info, and add an attachment to the correspondence.

One way of resolving this could be to rename the spec feature `Viewing an attachment that has consumer info` and don't talk about how it got there.

psd-web/app/views/documents/_document_card.html.slim
edwardhorsford: @StupidCodeFactory yes that's exactly it - both the description and these links should only be shown if the parent condition is met.

psd-web/spec/features/attachment_visibility.rb
edwardhorsford, line 3:
```suggestion
RSpec.feature "Viewing an attachment with that contains consumer info", :with_stubbed_elasticsearch, :with_stubbed_antivirus, :with_stubbed_mailer, type: :feature do
```

PR #578, opened 2020-05-07
Reorganise page expectation helpers into a single support file

frankieroberto:
This should help us to avoid duplication of the helper methods across different feature specs.

As part of this, I spotted that we had some identical helper methods, some slight variants, and some which were named the same thing but were asserting different pages (in which case I renamed one of them).

A few of the page expectations are used both within an edit scenario, where the ID of the thing is know as it is set up before the test using a factory, and a creation scenario, where the ID is generated as part of the flow. Rather than having 2 different helpers, I've merged them together and made the ID optional. If missing then the expectations uses a regex for the ID instead.

### DIFF
--- a/psd-web/spec/support/features/page_expectations.rb
+++ b/psd-web/spec/support/features/page_expectations.rb
@@ -1,0 +1,275 @@
+module PageExpectations
+  def expect_to_be_on_the_homepage
+    expect(page).to have_current_path("/")
+  end
+
+  # Cases pages
+  def expect_to_be_on_case_page(case_id: nil)
+    if case_id
+      expect(page).to have_current_path("/cases/#{case_id}")
+    else
+      expect(page).to have_current_path(/\/cases\/[\d\-]+$/)
+    end
+    expect(page).to have_selector("h1", text: "Overview")
+  end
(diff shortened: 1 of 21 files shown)


frankieroberto:
@StupidCodeFactory thanks. And yeah, we could further split these up – maybe not per-page (as they're all different pages – but by broad sections/flows? I've tried to do this a bit using comments.

Suggest we merge this as a first step and refine more after.

psd-web/spec/support/features/page_expectations.rb
```diff
@@ -0,0 +1,275 @@
...
+    expect(page).to have_selector("h3", text: "Document details")
+    expect(page).to have_link("Back", href: investigation_path(investigation))
+  end
+
+  def expect_to_be_on_record_corrective_action_for_case_page
+    expect(page).to have_current_path("/cases/#{investigation.pretty_id}/corrective_actions/details")
+    expect(page).to have_selector("h1", text: "Record corrective action")
+  end
+
+  def expect_to_be_on_confirmation_page
```
slorek, line 38: This could do with a better name

PR #579, opened 2020-05-07
Task to add case creators as collaborators

frankieroberto:
This is a one-off task that will be needed, as in future this should happen automatically when re-assigning the case (once that feature is implemented).

See https://trello.com/c/V5cvBHmm/531-data-migration-add-all-case-creators-to-their-cases-as-team-collaborator


psd-web/lib/tasks/add_case_creators_as_collaborators.rake
```diff
@@ -0,0 +1,19 @@
...
+    creator_team = creator.team
+    assignee_team = investigation.assignee_team
+    collaborator_teams = investigation.teams
+
+    if creator_team != assignee_team && !collaborator_teams.include?(creator_team)
+      count += 1
+
+      investigation.collaborators.create!(team: creator_team, added_by_user: creator, include_message: false)
+      puts "Adding #{creator_team.name} as a collaborator to case #{investigation.pretty_id}"
+    end
```
StupidCodeFactory, line 15: Suggestion: I think you're missing a bit of error handling here, I think a simple rescue and print to `SDTERR` any record that failed to save. This way we can have a look at just failed records when all has run.
Another thing to consider here, in what state would that leave an investigation if the data migration failed? 
I think it would be safer to wrap this in a transaction and have an all/or nothing strategy to avoid temporary funky states.

frankieroberto: If it fails to save, then an exception will be raised and the task will exit. Then it can be safely re-run and the previously created records will be skipped (as those teams will already be a collaborator).

I'll move the `puts` line before the call to `create!` as that'll then tell us which record failed to save.

PR #585, opened 2020-05-11
Allow whitelisting 2FA OTP for testing

slawosz:
Currently we can't do automated testing on `int` and `staging` environments. This PR introduces `WHITELISTED_2FA_CODES` env variable which is used to store 2FA codes which will always pass. Use with care.


nasirkhanbeis:
Thanks @slawosz  just works as expected. this should allow me to run my automated tests.yay!!

psd-web/app/models/secondary_authentication.rb
```diff
@@ -86,4 +86,13 @@ def increment_attempts_and_try_to_lock
       user.update!(second_factor_attempts_locked_at: Time.now.utc, second_factor_attempts_count: 0)
     end
   end
+
+  def whitelisted_code_valid?(otp)
+    if ENV["WHITELISTED_2FA_CODES"].present?
```
StupidCodeFactory, line 91: maybe only one whitelisted code is enough? what would be the use case for several whitelisted code?

psd-web/app/models/secondary_authentication.rb
```diff
@@ -86,4 +86,13 @@ def increment_attempts_and_try_to_lock
...
+      codes = ENV["WHITELISTED_2FA_CODES"].split(",").map(&:to_s)
```
StupidCodeFactory, line 92: Please change: ✂️ you don't need the `to_s` anything from the env is always string

psd-web/app/models/secondary_authentication.rb
```diff
@@ -86,4 +86,13 @@ def increment_attempts_and_try_to_lock
...
+      codes.include?(otp.to_s)
```
StupidCodeFactory, line 93: ✂️ `to_s`

edwardhorsford:
Should this explicitly guard against running on production? It looks implicit based on the env var not existing rather than any explicit check. Otherwise a comment might be good that this is only meant for non-production environments.

For this we might also want to consider adding something to the readme.

PR #586, opened 2020-05-11
POC: case owner collaborator

StupidCodeFactory:
## spike to change/merge the ownership, collaborator, creator business domains.

This is by no mean production ready. 
It's mostly me playing around with data models and see how it flows through application.
The class name, as described in the comments are subject to changes. 

Things I would have liked to include is migrations to add unique constraint 
on investigation_id, collaborating_type, collaborating_id on the collaborators table

## Case Owner / Collaborator models / permissions

As we are going to add more permissions to the application, we need a more flexible and centralised way for out future account management needs.
Some of those permission will be fairly classic _read_/_write access.

### Current situation

the core of the permission rely on what a `User` and/or `Team` can do.
It is import the permission system is resilient enough to not have to know about that the underlying model being a `User` or a `Team`.
This would lead to spagetti code with lots of conditionals all over the place and an exponential complexity of cases to handle.

At the moment we have 3 models that reference either one or both of the `User` and `Team`

1. `Source`
   This model is polymorphic and reference either a `User` or a `Team` via the `Source` and `UserSource` model.
   This model is used as an association on the investigation to identify who has created an investigation.
   It currently set in a very obstructive way and rely on the fact the a user is logged currently logged in or not.
   This makes our tests data brittle since it is very easy not to setup this assiciation correcty.

2. Owner
   This model is polymorphic and reference either a `User` or a `Team` directly.
   This model is used as an association on the investigation to identify who currently owns an investigation.
   Owners, whether it be a user or a team, determines which user or user within a team can re-assign (i.e change ownershio) of an investigation, view or change the certain attributes of and investgation and also add collaborators.

3. Collaborator
   This model is *not* polymorphic and only directly references a `Team`.
   This is the first milestones of a our permission work and was introduced in order to _allow_ other team to collaborate (i.e add view permission) on a *private* investigation.

### Problem with the current system

In order to know what permissions and a team or a user has for a given case, with the current system we potentially have to query 4 or five different tables and deal with object with a completely different interface.

### (multi) Polymorphic Collaborator model

By adding a polymorphic `:collaborating` association we allows a collaborator to reference either a `Team` or a `User` and possibly in the future anther completely different type of collaboration like may and Admin, a cat or a dog it does not matter. As long as all the `:collaborating` referenced relation have a common interface. We can substitute one for another and the `Collaborator` does not need to care if is a `User`, `Team`, a cat or even a dog. This is comonly know as ducktyping in the ruby community but more generaly in the programming community it's the third pricinple of the S.O.L.I.D principles.

We now can reference either a `Team` or `User` via the Collaborator object. But now, how to we model whether a this `:collaborating` reference is a `Source`, `UserSource`, `Owner` or a just a simple `Collaborator`, hence how to establish what set of permission a `:collaborating` object has in regards to an investigation?

This is where polymorphism on `Collaborator` with Single Table Inheritence is here to help. It is pretty obvious by now that each collaborator do not have the same permissions.
After consulting with Ed, with divised a different sets of permissions

Think of this data modle a bit like a pointer to a pointer in C or C++, you can manipulate the data and defer  as much as possible the constraining access to type of the underlying value;

Here is the rhierarchy, bear in mind that class names are subject, Slawosh has already made some good suggestions:

```ruby
class Collaborators::Base;
  self.abstract_class = true
  belongs_to :invesitgation
  belongs_to :collaborating, polymorphic: true
end

class Collaborators::Current < Collaborators::Base
  self.abstract_class = true
end

# CaseCreator will replace the Source and SourceUser object.
# right now case creator do not not have any type of permission.
# But this record is useful as we always want to show which user and/or team created the investigation.
# This gracefully handles the case where a user changed teams but created the case when he/she was part of another team.
# This way we do not give extra permissions to these types of collaborator.
# This are also extensively used in audit log, the formerly known :source association
class Collaborators::CaseCreator < Collaborators::Base

# Ed wants to store both to be able to easly display with a label which one which are the creators in the collaborator view.
# Currently we store either one or the other, a user or a team.
class Collaborators::CaseCreatorTeam < Collaborators::CaseOwner; end
class Collaborators::CaseCreatorUser < Collaborators::CaseOwner; end


# the owner types
class Collaborators::CaseOwner < Collaborators::Current
  self.abstract_class = true
end
# This allows to assign either a team or user as an owner.
# I've created a services class AssignInvestigation that allows the managing new case owner assignments,
# when assigning a User a CaseOwnerTeam is also created.
# @Ed: what was the rational behing having the two at the same time?
class Collaborators::CaseOwnerTeam < Collaborators::CaseOwner; end
class Collaborators::CaseOwnerUser < Collaborators::CaseOwner; end

# This is your regular collbaborator.
class Collaborator < Collaborators::Current; end

# Ed wants to keep the history of retired collaborator. "Assign to a previously collaborator"
class Collaborators::Historical < Collaborators::Current; end
```

This type of modelling is super flexible has you can retrieve part of the hierarchy tree `investigation.current_collaborators #  => [Collaborator, CaseOwnerTeam, CaseOwnerUser, OtherFuturTypeOfPermission]`
without messing with several boolean columns and complex checking, often order dependent, to determine whether a user or a team has a certain permission. It also avoid boolean flag creap on table (i.e has_accepted_declaration, has_viewed_introduction, has_been_sent_welcome_email, mobile_number_verified... :rofl: which all represent on state of a user can be at time).

It's also flexible enough that is can allow for future collaborator type to be added witout having to add another db field and backfill it.
This also has the flexibily the we might want to scope/override some of the permission for a give type of Collaborator on specific investigation.
```ruby
collaborator.permission = { can_change_owner: true }
```

This class hierarchy will work beautifully with Pundit, as each class with have it own policy class, possibly respecting the Collaborators hierarchy and computing on the fly if any override or special permission has been assigned to a given type of collaborator.

### Freebies(-ish)

We'll have to index ownership slighlty differently. [See these line](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/586/files#diff-2cb85cf5237b33fabd4731cbac15181fR14-R20)
The gives us to be abilty to perform search against a viewing user. And the search would not show cases you do not have access to.
I think this is a slight hurdle to overcome that will however benefit the user.

### Conclusion

This is a fair bit of work, but it should remove a lot of the complexity.
I'd would split this in probaly several steps:
1. First deployable: chunk: Keep the data structures Source, UserSource but start double writing to collaborators (CaseCreator and CaseOwner) table at the same time as source and owner.
2. Second deployable: Backfill hisetorical data and
3. Swap view to use, the new data structure
4. fix the search ( this can be done on a concurrently to step 2 and 3)
5. bundle steps 3 and 4 and deploy together.

### DIFF
--- a/doc/decisions/0002-multi-polymorphoc-collaborator-the-heart-of-account-management.md
+++ b/doc/decisions/0002-multi-polymorphoc-collaborator-the-heart-of-account-management.md
@@ -1,0 +1,104 @@
+## Case Owner / Collaborator models / permissions
+
+As we are going to add more permissions to the application, we need a more flexible and centralised way for out future account management needs.
+Some of those permission will be fairly classic _read_/_write access.
+
+### Current situation
+
+the core of the permission rely on what a `User` and/or `Team` can do.
+It is import the permission system is resilient enough to not have to know about that the underlying model being a `User` or a `Team`.
+This would lead to spagetti code with lots of conditionals all over the place and an exponential complexity of cases to handle.
+
+At the moment we have 3 models that reference either one or both of the `User` and `Team`
+
+1. `Source`
+   This model is polymorphic and reference either a `User` or a `Team` via the `Source` and `UserSource` model.
+   This model is used as an association on the investigation to identify who has created an investigation.
+   It currently set in a very obstructive way and rely on the fact the a user is logged currently logged in or not.
+   This makes our tests data brittle since it is very easy not to setup this assiciation correcty.
+
+2. Owner
+   This model is polymorphic and reference either a `User` or a `Team` directly.
+   This model is used as an association on the investigation to identify who currently owns an investigation.
+   Owners, whether it be a user or a team, determines which user or user within a team can re-assign (i.e change ownershio) of an investigation, view or change the certain attributes of and investgation and also add collaborators.
+
+3. Collaborator
+   This model is *not* polymorphic and only directly references a `Team`.
+   This is the first milestones of a our permission work and was introduced in order to _allow_ other team to collaborate (i.e add view permission) on a *private* investigation.
+
+### Problem with the current system
+
+In order to know what permissions and a team or a user has for a given case, with the current system we potentially have to query 4 or five different tables and deal with object with a completely different interface.
+
+
+### (multi) Polymorphic Collaborator model
+
+By adding a polymorphic `:collaborating` association we allows a collaborator to reference either a `Team` or a `User` and possibly in the future anther completely different type of collaboration like may and Admin, a cat or a dog it does not matter. As long as all the `:collaborating` referenced relation have a common interface. We can substitute one for another and the `Collaborator` does not need to care if is a `User`, `Team`, a cat or even a dog. This is comonly know as ducktyping in the ruby community but more generaly in the programming community it's the third pricinple of the S.O.L.I.D principles.
+
+We now can reference either a `Team` or `User` via the Collaborator object. But now, how to we model whether a this `:collaborating` reference is a `Source`, `UserSource`, `Owner` or a just a simple `Collaborator`, hence how to establish what set of permission a `:collaborating` object has in regards to an investigation?
+
+This is where polymorphism on `Collaborator` with Single Table Inheritence is here to help. It is pretty obvious by now that each collaborator do not have the same permissions.
+After consulting with Ed, with divised a different sets of permissions
+
+
+There basically three main branches in the hierarchy
+
+```ruby
+class Collaborators::Base;
+  self.abstract_class = true
+  belongs_to :invesitgation
+  belongs_to :collaborating, polymorphic: true
+end
+
+class Collaborators::Current < Collaborators::Base
+  self.abstract_class = true
+end
+
+# CaseCreator will replace the Source and SourceUser object.
+# right now case creator do not not have any type of permission.
+# But this record is useful as we always want to show which user and/or team created the investigation.
+# This gracefully handles the case where a user changed teams but created the case when he/she was part of another team.
+# This way we do not give extra permissions to these types of collaborator.
+# This are also extensively used in audit log, the formerly known :source association
+class Collaborators::CaseCreator < Collaborators::Base
+
+# Ed wants to store both to be able to easly display with a label which one which are the creators.
+# Currently we store either one or the other, a user or a team.
+class Collaborators::CaseCreatorTeam < Collaborators::CaseOwner; end
+class Collaborators::CaseCreatorUser < Collaborators::CaseOwner; end
+
+
+# the owner types
+class Collaborators::CaseOwner < Collaborators::Current
+  self.abstract_class = true
+end
+# This allows to assign either a team or user as an owner.
+# I've created a services class AssignInvestigation that allows the managing new case owner assignments,
+# when assigning a User a CaseOwnerTeam is also created.
+# @Ed: what was the rational behing having the two at the same time?
+class Collaborators::CaseOwnerTeam < Collaborators::CaseOwner; end
+class Collaborators::CaseOwnerUser < Collaborators::CaseOwner; end
+
+# This is your regular collbaborator.
+class Collaborator < Collaborators::Current; end
+
+# Ed wants to keep the history of retired collaborator. "Assign to a previously collaborator"
+class Collaborators::Historical < Collaborators::Current; end
+```
+
+This type of modelling is super flexible has you can retrieve part of the hierarchy tree `investigation.current_collaborators #  => [Collaborator, CaseOwnerTeam, CaseOwnerUser, OtherFuturTypeOfPermission]`
+without messing with several boolean columns and complex checking, often order dependent, to determine whether a user or a team has a certain permission. It also avoid boolean flag creap on table (i.e has_accepted_declaration, has_viewed_introduction, has_been_sent_welcome_email, mobile_number_verified... :rofl: which all represent on state of a user can be at time).
+
+It's also flexible enough that is can allow for future collaborator type to be added witout having to add another db field and backfill it.
+This also has the flexibily the we might want to scope/override some of the permission for a give type of Collaborator on specific investigation.
+```ruby
+collaborator.permission = { can_change_owner: true }
+```
+
+This class hierarchy will work beautifully with Pundit, as each class with have it own policy class, possibly respecting the Collaborators hierarchy and computing on the fly if any override or special permission has been assigned to a given type of collaborator.
+
+### Freebies(-ish)
+
+We'll have to index ownership slighlty differently. [See these line](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/586/files#diff-2cb85cf5237b33fabd4731cbac15181fR14-R20)
+The gives us to be abilty to perform search against a viewing user. And the search would not show cases you do not have access to.
+I think this is a slight hurdle to overcome that will however benefit the user.
--- a/doc/decisions/0002-multi-polymorphoc-collaborator-the-heart-of-account-management.md
+++ b/doc/decisions/0002-multi-polymorphoc-collaborator-the-heart-of-account-management.md
@@ -40,8 +40,9 @@ We now can reference either a `Team` or `User` via the Collaborator object. But 
 This is where polymorphism on `Collaborator` with Single Table Inheritence is here to help. It is pretty obvious by now that each collaborator do not have the same permissions.
 After consulting with Ed, with divised a different sets of permissions
 
+Think of this data modle a bit like a pointer to a pointer in C or C++, you can manipulate the data and defer  as much as possible the constraining access to type of the underlying value;
 
-There basically three main branches in the hierarchy
+Here is the rhierarchy, bear in mind that class names are subject, Slawosh has already made some good suggestions:
 
 ```ruby
 class Collaborators::Base;
@@ -102,3 +103,16 @@ This class hierarchy will work beautifully with Pundit, as each class with have 
 We'll have to index ownership slighlty differently. [See these line](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/586/files#diff-2cb85cf5237b33fabd4731cbac15181fR14-R20)
 The gives us to be abilty to perform search against a viewing user. And the search would not show cases you do not have access to.
 I think this is a slight hurdle to overcome that will however benefit the user.
+
+
+
+
+### Conclusion
+
+This is a fair bit of work, but it should remove a lot of the complexity.
+I'd would split this in probaly several steps:
+1. First deployable: chunk: Keep the data structures Source, UserSource but start double writing to collaborators (CaseCreator and CaseOwner) table at the same time as source and owner.
+2. Second deployable: Backfill hisetorical data and
+3. Swap view to use, the new data structure
+4. fix the search ( this can be done on a concurrently to step 2 and 3)
+5. bundle steps 3 and 4 and deploy together.
--- a/psd-web/app/forms/edit_investigation_collaborator_form.rb
+++ b/psd-web/app/forms/edit_investigation_collaborator_form.rb
@@ -31,7 +31,7 @@ class EditInvestigationCollaboratorForm
 private
 
   def collaborator
-    investigation.collaborators.find_by!(team_id: team.id)
+    investigation.collaborators.find_by!(collaborating: team)
   end
 
   def schedule_delete_emails
--- a/psd-web/app/forms/edit_investigation_collaborator_form.rb
+++ b/psd-web/app/forms/edit_investigation_collaborator_form.rb
@@ -7,7 +7,7 @@ class EditInvestigationCollaboratorForm
   attribute :permission_level
   attribute :message
   attribute :investigation
-  attribute :team
+  attribute :collaborator
   attribute :user
   attribute :include_message, :boolean
 
@@ -30,19 +30,15 @@ class EditInvestigationCollaboratorForm
 
 private
 
-  def collaborator
-    investigation.collaborators.find_by!(collaborating: team)
-  end
-
   def schedule_delete_emails
     schedule_delete_email(find_emails)
   end
 
   def find_emails
-    if team.team_recipient_email.present?
-      [team.team_recipient_email]
+    if collaborator.email.present?
+      [collaborator.email]
     else
-      team.users.active.pluck(:email)
+      collaborator.collaborating.users.active.pluck(:email)
     end
   end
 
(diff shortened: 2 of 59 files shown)


psd-web/app/models/collaborators/base.rb
```diff
@@ -0,0 +1,25 @@
+module Collaborators
+  class Base < ApplicationRecord
+    self.table_name = "collaborators"
+
+    belongs_to :investigation, optional: true
+    belongs_to :collaborating, polymorphic: true, optional: false
```
slawosz, line 6: Wouldn't be `collaboration` better name for this? I feel `collaborating` is bad English word in this case

psd-web/app/models/investigation.rb
```diff
@@ -51,22 +48,66 @@ class Investigation < ApplicationRecord
 
   has_many_attached :documents
 
-  has_one :source, as: :sourceable, dependent: :destroy
   has_one :complainant, dependent: :destroy
 
+  # any old collaborators
+  has_many :historic_collaborators, class_name: "Collaborators::Historic", dependent: :destroy
```
slawosz, line 54: Aren't you loosing the collaborator type here?

psd-web/app/models/collaborators/case_owner_user.rb
```diff
@@ -0,0 +1,8 @@
+module Collaborators
+  class CaseOwnerUser < CaseOwner
+
+    def make_collaborator!(*)
+      destroy!
```
slawosz, line 5: Why so?

psd-web/db/schema.rb
```diff
@@ -75,12 +75,16 @@
 
   create_table "collaborators", force: :cascade do |t|
```
slawosz, line 76: Actually, this table (and model) should be named `Collaboration` and users and teams would become `Collaborators`.

psd-web/app/models/collaborators/base.rb
slawosz: I proposed different naming in schema. So this class would become `Collaboration::Base`. Alternatively, you could change `collaborating` to `collaboratee`, but I prefer first option.

doc/decisions/0002-multi-polymorphoc-collaborator-the-heart-of-account-management.md
```diff
@@ -0,0 +1,118 @@
...
+# But this record is useful as we always want to show which user and/or team created the investigation.
+# This gracefully handles the case where a user changed teams but created the case when he/she was part of another team.
+# This way we do not give extra permissions to these types of collaborator.
+# This are also extensively used in audit log, the formerly known :source association
+class Collaborators::CaseCreator < Collaborators::Base
+
+# Ed wants to store both to be able to easly display with a label which one which are the creators.
+# Currently we store either one or the other, a user or a team.
+class Collaborators::CaseCreatorTeam < Collaborators::CaseOwner; end
+class Collaborators::CaseCreatorUser < Collaborators::CaseOwner; end
```
slorek, line 69: Do you mean to say `Collaborators::CaseCreatorUser < Collaborators::CaseCreator` here?

doc/decisions/0002-multi-polymorphoc-collaborator-the-heart-of-account-management.md
```diff
@@ -0,0 +1,118 @@
+## Case Owner / Collaborator models / permissions
```
slorek, line 1: Let's include the text from @frankieroberto 's options ADR from https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/565 so it details the other options we considered. We can then add the reason for choosing the selected option.

psd-web/app/models/collaborators/base.rb
StupidCodeFactory: I'd for `collaborations` that's effectively what it is. That's how I named the reverse association on user and team

psd-web/app/models/collaborators/case_owner_user.rb
StupidCodeFactory: you only want to retain the `CaseOwnerCreator`. The `CaseOwnerUser` is here more as convenience to display in the UI. Once it is no longer the case owner or its team is no longer the case owner  this becomes obsolete

psd-web/app/forms/edit_investigation_collaborator_form.rb
```diff
@@ -30,27 +30,23 @@ def save!
...
-  end
-
   def schedule_delete_emails
     schedule_delete_email(find_emails)
   end
 
   def find_emails
-    if team.team_recipient_email.present?
-      [team.team_recipient_email]
+    if collaborator.email.present?
```
edwardhorsford, line 38: I'd be cautious about renaming the team inbox to this more generic form. To discuss.

doc/decisions/0002-multi-polymorphoc-collaborator-the-heart-of-account-management.md
```diff
@@ -0,0 +1,118 @@
...
+   This is the first milestones of a our permission work and was introduced in order to _allow_ other team to collaborate (i.e add view permission) on a *private* investigation.
+
+### Problem with the current system
+
+In order to know what permissions and a team or a user has for a given case, with the current system we potentially have to query 4 or five different tables and deal with object with a completely different interface.
+
+
+### (multi) Polymorphic Collaborator model
+
+By adding a polymorphic `:collaborating` association we allows a collaborator to reference either a `Team` or a `User` and possibly in the future anther completely different type of collaboration like may and Admin, a cat or a dog it does not matter. As long as all the `:collaborating` referenced relation have a common interface. We can substitute one for another and the `Collaborator` does not need to care if is a `User`, `Team`, a cat or even a dog. This is comonly know as ducktyping in the ruby community but more generaly in the programming community it's the third pricinple of the S.O.L.I.D principles.
```
slorek, line 36: I like the idea of polymorphism to model permissions in this way, as it will integrate with Pundit nicely and provide more flexibility. I think we can improve on the name `collaborating` (for me it's confusing with `collaborator` and doesn't really describe what it does). Perhaps `CollaboratorRole`.

My other question is how best we can gracefully migrate someone from `owner` to `collaborator` to `historical` and avoid data juggling.

doc/decisions/0002-multi-polymorphoc-collaborator-the-heart-of-account-management.md
```diff
@@ -0,0 +1,118 @@
...
+# when assigning a User a CaseOwnerTeam is also created.
+# @Ed: what was the rational behing having the two at the same time?
+class Collaborators::CaseOwnerTeam < Collaborators::CaseOwner; end
+class Collaborators::CaseOwnerUser < Collaborators::CaseOwner; end
+
+# This is your regular collbaborator.
+class Collaborator < Collaborators::Current; end
+
+# Ed wants to keep the history of retired collaborator. "Assign to a previously collaborator"
+class Collaborators::Historical < Collaborators::Current; end
```
slorek, line 87: Perhaps we can track this with a `deleted_at` or retired_at` timestamp instead of moving data between two classes.

psd-web/app/models/investigation.rb
```diff
@@ -51,22 +48,66 @@ class Investigation < ApplicationRecord
...
+  # => add index to enforce CollaboratingType to be "Team"
+  has_one :case_creator_team, class_name: "Collaborators::CaseCreatorTeam", dependent: :destroy, inverse_of: :investigation
+  # => add index to enforce CollaboratingType to be "User"
+  has_one :case_creator_user, class_name: "Collaborators::CaseCreatorUser", dependent: :destroy, inverse_of: :investigation
+
+  has_many :case_owners, class_name: "Collaborators::CaseOwner", dependent: :destroy, inverse_of: :investigation
+  # => add index to enforce (type, CollaboratingType) to be ("Collaborators::CaseOwnerTeam", "Team")
+  has_one :case_owner_team, class_name: "Collaborators::CaseOwnerTeam", dependent: :destroy, inverse_of: :investigation
+  # => add index to enforce (type, CollaboratingType) to be ("Collaborators::CaseOwnerUser, User")
+  has_one :case_owner_user, class_name: "Collaborators::CaseOwnerUser", dependent: :destroy, inverse_of: :investigation
```
slorek, line 66: This might confusing as the case will have _either_ a team or a user owner. Same with creators, but this implies it should have both. I would consider removing these and just having a `has_one :case_owner, class_name: "Collaborators::CaseOwner"`.

psd-web/app/forms/edit_investigation_collaborator_form.rb
slawosz: This is fine, makes code better.

doc/decisions/0002-multi-polymorphoc-collaborator-the-heart-of-account-management.md
StupidCodeFactory: yes - most of will go in the PR description. It was just much easier to write this in Emacs ... like everything else.

doc/decisions/0002-multi-polymorphoc-collaborator-the-heart-of-account-management.md
edwardhorsford: Likewise I find some of the names overlap quite a lot.

doc/decisions/0002-multi-polymorphoc-collaborator-the-heart-of-account-management.md
edwardhorsford: I don't think we need to store rights level (edit / view), but it might be helpful to know if they were ever a case owner.

Historic teams is wow priority - ages ago I'd wanted to be able to show previous assignees (on the assignment page) and I couldn't because we didn't know them. Now that we have teams on a case showing an autocomplete of them achieves a similar thing.

Audit wise it might be helpful to know prior teams too?

doc/decisions/0002-multi-polymorphoc-collaborator-the-heart-of-account-management.md
StupidCodeFactory: I've written a service class below to owner -> collaborator. I think the crux really it is to flip the "type" column and reload the `investigation`.
I agree the names are not ideal, I really focused on trying a couple ways to model the associations in a meaningful way.

psd-web/app/forms/edit_investigation_collaborator_form.rb
slorek: I think it's an improvement and means we can use the same interface to notify a team as a user. We could go even further, i.e.

```
class Team
  def email_recipients
    return [email] if email
    users.map(&:email)
  end
end

class User
  def email_recipients
    [email]
  end
end
```
This would give us a common interface where the class controls the recipients instead of some convoluted mail delivery code which needs to know about these different class types.

psd-web/app/models/collaborators/case_owner_user.rb
slorek: This would imply that there should always be a row for that same user/team as a standard `collaborator` as well. This seems a bit redundant as we can include case creators in the `collaborators` association and get all in one hit, and it could get very complex if a user/team has more than one type of collaboration against a case. I'd advocate just one row per user/team.

doc/decisions/0002-multi-polymorphoc-collaborator-the-heart-of-account-management.md
StupidCodeFactory: no that's really how I mean to model it. But I quite like
```ruby
Collaborators::CaseCreator
Collaborators::CaseCreator::User
``` 
`Collaborators::CaseCreator` => will a aways reference a team. So Ed suggested that we add `Team` at the end. I think it is fine without Team.

originialy I had it modelled this way
```ruby
CaseCreator > CaseOwner > CaseOwner::User
Class Investigation
  has_many :case_owners, -> { order("CASE WHEN type = 'CaseOwner' THEN 0 ELSE 1 END", "ASC") }
  has_one :case_owner
  has_one :case_owner_user 
end
```
this works as well, there is just a little order trick to guaranty that `investigation.case_owner` returns the instance of `CaseOwner`  and not `CaseOwnerUser`

I did actually prefer it this way. But decided otherwise because that meant losing the order clause on the association and didn't want to scare other people with more _advanced_  active record feature.

psd-web/app/models/investigation.rb
edwardhorsford: The intention would be that cases *always* have an owner team, and they can optionally *additionally* have an owner user.

We want permissions to be based around teams (and ultimately case ownership to be around teams).

The user is more the person who is the 'assigned' person. Perhaps this should be renamed to make that clearer. They get no different rights, but some bits of the ui present differently, and notifications don't go to the owner team where there's an assigned user.

doc/decisions/0002-multi-polymorphoc-collaborator-the-heart-of-account-management.md
StupidCodeFactory: even better we can have `Collaborator::Historical::CaseOwner` for each type and no need for a new column the created_at will do just fine.

psd-web/app/models/investigation.rb
StupidCodeFactory: no this is a sub class of root `Collaborators::Base` class

psd-web/db/schema.rb
StupidCodeFactory: Yes collaborations is probably a better name. @slorek also mentioned that maybe  something that mentions the idea roles or permissions would be good. I'm not fussed and I think it is not massive stretch to understand that a Collaboration is holding the permissions.

psd-web/app/forms/edit_investigation_collaborator_form.rb
StupidCodeFactory: I would have to add another if else to check where it is a team or a user. 
This is not actually renamed on the Team object,  and in the database it is still team_recipient_email.

psd-web/app/models/investigation.rb
slorek: I think renaming the user here would make it clearer that there is a distinction and model the data closer to how we want it to work.

frankieroberto:
Thanks for this Yann.

I'm not convinced we need to merge the creator user / team in with the collaborators table, as this has no bearing upon permissions, and doesn’t change. Would be simpler to have a `created_by_user_id` foreign key on the investigations table, as at present.  Also not convinced we should store previous collaborators in the collaborators table yet, until there's a direct need for it.

In terms of language, my first thought is to have:

* `owner` = team which owns the case
* `collaborators` = teams which are working on the case (should include the owner, probably?)
* `created_by` = user who created the case
* `created_by_user_team` = team of the user who created the case (possibly not needed as we can now call `created_by.team` as users always belong to a single team, although there's a small chance they might have moved teams since creating the case).
* `assigned_to` = individual user currently working on the case (can be null)

We also have an additional sub-type of collaborator teams, who additionally have "update" rights. These could be called `collaborators_who_can_update` (struggling to think of a better term)?

edwardhorsford:
> although there's a small chance they might have moved teams since creating the case).

I don't think this is a small chance - it has happened several times now - opss people are moving teams somewhat often. I'd like to mark the creator team, and i'm very wary of doing this where we may be marking the wrong team.

PR #588, opened 2020-05-11
Refactor: move decorator method to view

frankieroberto:
This moves the code that renders the investigation summary out from a decorator class and into a a view.

Also switches from using the [`_govuk_summary_list`](https://github.com/UKGovernmentBEIS/govuk-design-system-rails/blob/master/app/views/components/_govuk_summary_list.html.slim) slim partial to the [`govukSummaryList()`](https://github.com/UKGovernmentBEIS/govuk-design-system-rails/blob/master/app/helpers/govuk_design_system/summary_list_helper.rb) helper.

Rationale is to avoid rendering HTML in decorators, to maintain a clearer distinction between the views (which output HTML) and the decorators, which can be more abstract (eg outputting strings or arrays).

This does mean accepting having _some_ conditions within the templates, although complex logic could be refactored out in some other way (eg helpers).


psd-web/app/views/investigations/tabs/_overview.html.erb
```diff
@@ -81,7 +81,65 @@
...
+        }
+      }
+    ] %>
+
+    <% if @investigation.complainant_reference.present?
+      about_the_case_rows << {
+        key: { text: "Trading Standards reference" },
+        value: { text: @investigation.complainant_reference }
+      }
+     end %>
```
StupidCodeFactory, line 137: Please change:  this is  really  messy and can we have helper methods at least in a helper and tests that goes with it.

frankieroberto: There’s only a single bit of logic: a condition on whether to add a "Trading Standards reference" row to the summary list. The rest of it is just building an object to pass to a helper.

I'll explore creating an `about_the_case_rows(@investigation)` helper, although I'd argue that this view should be tested by feature specs, not a spec for that helper, as the helper would kind-of be an implementation detail of the view, rather than an meaningful abstraction in its own right (no other views would use it).

edwardhorsford: I don't think this section has much business logic at all. It's a summary list that's included with all case types for all users. It doesn't seem "really messy" at all.

I think each summary block could exist as a partial to be called (as some other summary blocks are case dependant).

frankieroberto: @StupidCodeFactory Here's a quick refactoring of the rows into its own helper: #597. Do we think this is easier to follow, or an unnecessary abstraction?

@edwardhorsford Moving the whole section into its own partial (`_about_the_case.html.erb`), including the heading, could be helpful. Although that partial would still need to contain the `if` condition above.

slorek:
I definitely support not rendering HTML in decorators. I've previously added comments to PRs this effect. I have no problem with basic conditions in view templates, for example checking if a property exists or not in order to decide whether to display it. If it's business logic then it should be somewhere else, ideally the object itself and not a helper.

PR #594, opened 2020-05-12
Allow to change unverified phones on 2FA

scruti:
[Trello Card](https://trello.com/c/btooGOgl/397-3-allow-a-user-to-change-their-mobile-number-when-registering-if-they-havent-received-a-code)

## Description
Users that still have not verified their mobile numbers through 2FA can find themselves in the situation of needing to change their mobile number in order to be able to receive 2FA codes by sms.

To avoid locking out users in that situations, we will allow them to update their mobile number when requesting their 2FA code to be re-sent to them.

- The code will be sent to this new phone.
- The phone number will be updated for the user for subsequent 2FA checks. 

#### An user with a verified mobile number will see this
Screenshot from 2020-05-12 17-04-40

#### While an user that still has not verified their mobile number will see this
Screenshot from 2020-05-12 17-12-57
When expanding this they will be able to see their current mobile number and modify it prior to submission of a new code.

#### We keep the validations for mobile number format and presence
So when submitting this
Screenshot from 2020-05-12 17-13-34
We get this error:
Screenshot from 2020-05-12 17-15-20

And when submitting this:
Screenshot from 2020-05-12 17-16-29
We get this error:
Screenshot from 2020-05-12 17-16-39

## Checklist:


scruti:
[(3) Allow a user to change their mobile number when registering if they haven't received a code](https://trello.com/c/btooGOgl/397-3-allow-a-user-to-change-their-mobile-number-when-registering-if-they-havent-received-a-code)

psd-web/app/forms/resend_secondary_authentication_code_form.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  attribute :mobile_number
+  attribute :user
+
+  validates_presence_of :mobile_number, message: I18n.t(:blank, scope: %i[activerecord errors models user attributes mobile_number]),
+                                        if: -> { update_mobile_number? }
+  validates :mobile_number, phone: { message: I18n.t(:invalid, scope: %i[activerecord errors models user attributes mobile_number]) },
+                            if: -> { update_mobile_number? && mobile_number.present? }
+
+  def save!
```
slawosz, line 13: I am very picky here:
I wouldn't use `save!`, rather `save`, as you are returning `false` with error messages to display. I recon that `user.save!` in this case does not influence the flow of the code, its unexpected in normal flow. But itself does not return an exception. So in order to justify use of `!` here, you can write:

```ruby
  def save!
    unless valid?
      raise RecordInvalid # or so
    end
   
￼    if update_mobile_number?
￼      user.mobile_number = mobile_number
￼      user.save!
￼    end
￼    true
￼  end
end
```

In this case bang has sense. However, I would just rename method name from `save!` to `save` as even though you are using `user.save!` inside, but you are not expecting it to raise an exception, as you control state of the code here and in normal circumstances there is no possibility for exceptions. 
Please note, there is lot of methods which does not have bang, and which in extraordinary circumstances (network errors, unintended use) can return exception. In my opinion, your method belongs to this category.

slawosz: Also please note, that no `bang!` method in Rails Api or Ruby STD library returns value, they all return `nil`.

scruti: Good point. Changed it.

edwardhorsford:
This is what I get when there's an error:
Screenshot 2020-05-13 at 12 37 15

Three changes:
* Show `Error: ` in page title - I think you need to pass errors to the title helper.
* Can we please open the details element by default if there's an error (I think you add an `open` attribute to the element or if using a component pass a boolean `true` to the open param.
* The field itself should be in error [like this](https://design-system.service.gov.uk/components/text-input/#error-messages).

Other than those three things, looks good!

nasirkhanbeis:
@scruti 
looks like resend code on reset pwd journey is broken 
image

psd-web/app/views/secondary_authentications/resend_code/new.html.erb
```diff
@@ -8,12 +8,37 @@
...
+    <%= error_summary current_user.errors %>
+
     <h1 class="govuk-heading-l"><%= title %></h1>
 
     <p class="govuk-body">Text messages sometimes take a few minutes to arrive. If you do not receive the text message, you can resend it.</p>
 
-    <p class="govuk-body">If you no longer have access to the phone with the number you registered for this service, contact <%= mail_to t(:enquiries_email), nil, subject: "Update mobile number", class: "govuk-link" %> to reset the number.</p>
+    <%= form_with(url: resend_secondary_authentication_code_path, method: "post") do |form| %>
+      <% if @mobile_number_change_allowed %>
+        <details class="govuk-details" data-module="govuk-details">
```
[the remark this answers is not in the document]
edwardhorsford: If using the above helper, can pass `open: true` if there are errors.

scruti:
Seems the reset password flow is blowing up when requesting again a 2FA code because there is no `current_user` set when attempting to update the password following the reset_password link. 
Working on resolving it + applying the feedback.

psd-web/spec/features/registration_spec.rb
```diff
@@ -37,6 +37,18 @@
...
+      click_link "Not received a text message?"
+
+      expect_to_be_on_resend_secondary_authentication_page
+      find(".govuk-details__summary-text", text: "Change where the text message is sent").click
+      fill_in "Mobile number", with: "70123456789"
+
+      expect {
+        click_button "Resend security code"
+        invitee.reload
+      }.to change(invitee, :mobile_number).to "70123456789"
```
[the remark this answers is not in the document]
scruti: Moved this assertion to the request spec, going to switch it here to the user navigation to "Your account". Thanks for the suggestion.

psd-web/app/views/secondary_authentications/resend_code/new.html.erb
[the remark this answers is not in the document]
scruti: Had no idea about its existence. Thanks.

psd-web/spec/features/registration_spec.rb
```diff
@@ -37,9 +37,24 @@
...
       enter_secondary_authentication_code(invitee.reload.direct_otp)
 
       expect_to_be_on_declaration_page
+      check "I agree"
+      click_button "Continue"
+      click_link "Skip introduction"
+      click_link("Your account", match: :first)
+
+      expect(page).to have_h1("Your account")
+      expect(page).to have_css("dd", text: "70123456789")
```
frankieroberto, line 57: We have a [useful matcher](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/spec/support/matchers/capybara_matchers.rb#L7-L30) for this, which helps to assert which key in the summary table this is a value for:

```suggestion
      expect(page).to have_summary_item(key: "Mobile number", text: "70123456789")      
```

psd-web/app/controllers/secondary_authentications/resend_code_controller.rb
```diff
@@ -9,20 +9,60 @@ class ResendCodeController < ApplicationController
...
                        :set_cache_headers
 
+    before_action :require_user, only: %w[new create]
+
+    def new
+      @mobile_number_change_allowed = !user.mobile_number_verified
+    end
+
     def create
+      @mobile_number_change_allowed = !user.mobile_number_verified
```
slorek, line 19: Suggestion: Move this to a new model method e.g. `User#mobile_number_change_allowed?`. The decision to allow mobile number change is business logic tied to the state of the `mobile_number_verified` flag, implemented twice in this controller and again in the form.

psd-web/app/controllers/secondary_authentications/resend_code_controller.rb
```diff
@@ -9,20 +9,60 @@ class ResendCodeController < ApplicationController
...
+    def resend_code_form
+      @resend_code_form ||= ResendSecondaryAuthenticationCodeForm.new(mobile_number: mobile_number_param, user: user)
+    end
+
+    def user_id_for_secondary_authentication
+      current_user&.id || session[:secondary_authentication_user_id]
+    end
+
+    def user
+      @user ||= (current_user || User.find_by(id: session[:secondary_authentication_user_id]))
```
slorek, line 61: Suggestion: Don't set instance variables in `before_action` methods. @frankieroberto previously commented on #567 that this was a pattern we should consider standardising.

psd-web/app/forms/resend_secondary_authentication_code_form.rb
```diff
@@ -0,0 +1,20 @@
...
+                        message: I18n.t(:blank, scope: %i[activerecord errors models user attributes mobile_number]),
+                        if: -> { update_mobile_number? }
+  validates :mobile_number,
+            phone: { message: I18n.t(:invalid, scope: %i[activerecord errors models user attributes mobile_number]) },
+            if: -> { update_mobile_number? && mobile_number.present? }
+
+private
+
+  def update_mobile_number?
+    !user.mobile_number_verified
```
slorek, line 18: Same suggestion: move this to a new model method.

PR #597, opened 2020-05-13
Refactor the rows object into a helper

frankieroberto:
This moves the building of the `rows` argument passed to the `govukSummaryList()` helper with the "About the case" section to its own helper.

This is mainly to avoid having a condition within the view template.

See [discussion on parent Pull Request](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/588#discussion_r423549075).


scruti:
I like more this approach than having it embedded in the view file. But is a little bit of personal preference I guess.

PR #598, opened 2020-05-13
Fix search date order

slawosz:
Due to internal implementation of `elasticsearch-model`, in `updated_at` column in some cases didn't got synced. Also added kibana to repo, as it helps with browsing elasticsearch data.


psd-web/app/models/concerns/searchable.rb
```diff
@@ -5,6 +5,7 @@ module Searchable
   included do
     include Elasticsearch::Model
     include Elasticsearch::Model::Callbacks
+    after_touch() { __elasticsearch__.index_document }
```
frankieroberto, line 8: Would [`__elasticsearch__.update_document`](https://github.com/elastic/elasticsearch-rails/blob/62191f8cd494595b6ad4dee5e22a06c1091c1805/elasticsearch-model/lib/elasticsearch/model/indexing.rb#L401-L425) be slightly better (as it'll only do a partial update of the changed fields)?

slawosz: It is doing `update_document`, its implementation actually caused this bug.

frankieroberto: 🤦

PR #600, opened 2020-05-14
Fix name in ownership emails

slorek:
https://trello.com/c/eYtohraA/544-refactor-display-names

## Description
This fixes a minor bug in the notification emails which are sent when a team is added or removed from a case.

They did not show the team name of the user who added or removed your team. In normal circumstances the recipient would not be in the same team as the person who added or removed, since in that case they would already be a collaborator on the case.

In the case of the team deleted email I have also refactored the interface to the mailer since it depended on being given a serialised form object, which is unnecessary and slightly cumbersome. Passing the individual arguments instead will make it easier to call this method elsewhere if required and makes it easier to test. Since there were no unit tests for the mailer I added some.

## Checklist:

### DIFF
--- a/psd-web/app/forms/edit_investigation_collaborator_form.rb
+++ b/psd-web/app/forms/edit_investigation_collaborator_form.rb
@@ -63,11 +63,7 @@ private
 
   def schedule_delete_email(emails)
     emails.each do |email|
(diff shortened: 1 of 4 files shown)


nasirkhanbeis:
@slorek  is your latest commit around below issue? current user team not being shown?

image

PR #602, opened 2020-05-14
Remove brackets from link helper

edwardhorsford:
The hidden link helper was forcing hidden text to be in brackets. This isn't always needed.

I've removed the default text and reapplied where we want it.

Should outwardly look the same to users.


psd-web/app/views/link_helper/_link_with_hidden_text_to.html.erb
```diff
@@ -1 +1 @@
-<%= text %><span class="govuk-visually-hidden"> (<%= hidden_text %>)</span>
+<%= text %><span class="govuk-visually-hidden"> <%= hidden_text %></span>
```
edwardhorsford, line 1: Brackets no longer included by defualt.

PR #604, opened 2020-05-14
Rename collaborators to collaborations

StupidCodeFactory:
First step for new permissions model.

# New Models

`Collaboration` - abstract model which provides all relations between `Investigation` and `Collaborator`, where `Collaborator` is polymorfic entity, its either `User` or Team

`Edition` - is subclass of `Collaboration`, denote teams ability to edit `Investigation`.

# New Relations

## In `Investigation` model

`editors` - links `Investigation` with `Team` instances, denotes `Teams` with edit permisions
`editions` - links `Investigation` with `Edition` instance, used by `editors` relations

## In `Collaboration` model

`collaborator` - polymorphic relation linking model with either `Team` or `User`

## In `Edition` model

`editor` - links `Edition` with entities (now only `Teams` ) 

# Follow up tasks

- move `Investigation` creators to use the above model - will require new subclass of `Collaboration`
- add DB trigger for checking for uniqueness of `Edition` records (team, investigation), currently, uniqueness is only checked by Rails, which is prone to race conditions


slorek:
Thanks - first thought is that `edition` does not mean what you think it does, so we will need to come up for a different name for this.

frankieroberto:
I suggest we opt for some slightly more verbose names for the relations:

`teams_with_edit_permission`
`teams_with_view_all_permission`  (we don't currently have this, but it's coming soon)
`owner`
`collaborating_teams` = all of the above 3 combined
`assigned_officer`
`created_by`
`created_by_team`

models could be:

`Collaborator` (abstract super class)
`Collaborator::TeamWithEditPermission`
`Collaborator::TeamWithViewAllPermission`
`Collaborator::AssignedOfficer`
`Collaborator::CreatorUser`
`Collaborator::CreatorTeam`

slawosz:
For now, I would create:
Models:
`Collaboration` - super class
`Collaboration::EditAccess < Collaboration::Access`
`Collaboration::ViewAccess < Collaboration::Access`
`Collaboration::Access` purpose is to make `collaborating_teams` relation easily.
Relations:
`editors` or `teams_with_edit_permission`
`viewers` or `teams_with_view_permission`

Rest can be done as separate work as it does not blocks any further work.

StupidCodeFactory:
looks ok to merge I can not approved as I have created the PR. So someone else should give it a 👍

psd-web/app/models/investigation.rb
```diff
@@ -58,8 +58,8 @@ class Investigation < ApplicationRecord
   has_one :source, as: :sourceable, dependent: :destroy
   has_one :complainant, dependent: :destroy
 
-  has_many :collaborators, dependent: :destroy
-  has_many :teams, through: :collaborators
+  has_many :edit_accesses, dependent: :destroy, class_name: "Collaboration::EditAccess"
+  has_many :editors, through: :edit_accesses, dependent: :destroy, source: :editor, source_type: "Team"
```
frankieroberto, line 62: Think this should reference "teams" still, to make it clear that it always returns an array of teams, not users (which you might expect from the term "editors").

```suggestion
  has_many :teams_with_edit_access, through: :edit_accesses, dependent: :destroy, source: :editor, source_type: "Team"
```

psd-web/app/models/collaboration/access.rb
```diff
@@ -0,0 +1,4 @@
+class Collaboration < ApplicationRecord
+  class Access < Collaboration
+  end
```
frankieroberto, line 3: Is this class really needed, if it doesn't currently do anything? `EditAccess` could inherit directly from `Collaboration`?

psd-web/app/models/user.rb
```diff
@@ -13,6 +13,7 @@ class User < ApplicationRecord
   has_many :activities, through: :investigations
   has_many :user_sources, dependent: :destroy
   has_many :user_roles, dependent: :destroy
+  has_many :collaborations, dependent: :nullify, as: :collaborable
```
frankieroberto, line 16: Don't think we need this line yet?

psd-web/app/models/team.rb
```diff
@@ -2,6 +2,7 @@ class Team < ApplicationRecord
   belongs_to :organisation
   has_many :users, dependent: :restrict_with_exception
   has_many :investigations, dependent: :nullify, as: :owner
+  has_many :collaborations, dependent: :nullify, as: :collaborator
```
frankieroberto, line 5: Should this be `dependent: :destroy`? If a team is deleted, I'd expect their collaboration records to be deleted too.

PR #606, opened 2020-05-14
Only allow collaborators to update case

frankieroberto:
[Trello ticket](https://trello.com/c/i1bE83RF/527-only-case-owner-or-teams-added-to-a-case-can-add-data-or-edit-a-case)

This restricts most of the "update" actions on a case to collaborators. The exception is updating the case status (eg open / close), whether it’s restricted or not, and who the owners and collaborators are, which are all restricted to users of the case owner team.

# Implementation notes

I've reduced the number of policies in the `InvestigationPolicy` and renamed them to better reflect the newly agreed permission levels.


psd-web/app/helpers/activity_helper.rb
```diff
@@ -12,7 +12,7 @@ def activity_types
       "business": "Add a business to the case"
     }
     visibility_text = @investigation.is_private ? "Unrestrict this case" : "Restrict this case for legal privilege"
-    base_types["visibility"] = visibility_text if policy(@investigation).visibility?
+    base_types["visibility"] = visibility_text if policy(@investigation).view_non_protected_details?
```
edwardhorsford, line 15: I _think_ this should be `if policy(@investigation).change_owner_or_status?` - restriction should be scoped to just the case owner - and I *think* this is instead scoping to teams on the case.

If you visit the actual path it doesn't work (that's correctly scoped).

nasirkhanbeis:
@frankieroberto I found a minor issue around attachment - as view only user can delete the attachment, expected not to show any any links against existing attachments in attachment tab
image

psd-web/app/helpers/activity_helper.rb
frankieroberto: @edwardhorsford well spotted, fixed now.

nasirkhanbeis:
@frankieroberto  Also I think added team should be able to change coronavirus flag which isn't  atm.Not sure if we have mentioned this in criteria though. Can you pls look into it? thanks.

frankieroberto:
@nasirkhanbeis fixed both of the issues you spotted, thanks. Can you re-review please?

psd-web/spec/features/add_business_spec.rb
```diff
@@ -13,12 +13,15 @@
   let(:email)            { Faker::TvShows::TheITCrowd.email }
   let(:phone_number)     { Faker::PhoneNumber.phone_number  }
   let(:job_title)        { Faker::Job.title }
-  let(:investigation)    { create(:enquiry) }
-
-  before { sign_in }
+  let(:user)             { create(:user, :activated) }
+  let(:investigation)    { create(:enquiry, owner: user.team) }
+  let!(:another_user_another_team) { create(:user, :activated, email: "active.otherteam@example.com", organisation: user.organisation, team: create(:team)) }
```
scruti, line 18: There is any reason we need a `let!` here? 
If I read this right... this user only is needed in the second scenario and we do not need to be created at all besides on the `sign_in` moment.

psd-web/spec/features/investigation_edit_spec.rb
```diff
@@ -1,12 +1,14 @@
 require "rails_helper"
 
 RSpec.feature "Ability to edit an investigation", :with_elasticsearch, :with_stubbed_mailer, type: :feature do
-  let(:investigation) { create(:project) }
-
-  before { sign_in }
+  let!(:investigation) { create(:project, owner: user.team) }
+  let!(:user) { create(:user, :activated) }
+  let!(:another_user_another_team) { create(:user, :activated, email: "active.otherteam@example.com", organisation: user.organisation, team: create(:team)) }
```
scruti, line 6: same as before comment: Do we really need this user to be created on the first scenario?

psd-web/app/controllers/investigations/businesses_controller.rb
```diff
@@ -37,10 +38,12 @@ def create
   end
 
   def show
+    authorize @investigation, :update?
```
slorek, line 41: I'm not clear why we need update permission on the case to show a business. Code comment might help.

psd-web/spec/requests/investigations/updating_the_status_of_a_case_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  context "when the user belongs to the case owner’s team" do
+    before do
+      sign_in user_from_owner_team
+      User.current = user_from_owner_team
+      patch status_investigation_path(investigation),
+            params: {
+              investigation: { is_closed: "true", status_rationale: "Test" }
+            }
```
scruti, line 29: nitpick: Guess this could be extracted/abstracted too as is the same call in the 3 contexts... but not a requirement at all, I don't mind a little bit of duplication.

scruti:
Still looks good to me :+1:

psd-web/app/controllers/investigations/ownership_controller.rb
```diff
@@ -39,7 +39,7 @@ def clear_session
 
   def set_investigation
     investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
-    authorize investigation, :show?
+    authorize investigation, :view_non_protected_details?
```
slorek, line 42: 🤔 I'm not sure this controller is sufficiently protected from changing ownership by someone who shouldn't be able to.

psd-web/app/controllers/investigations/products_controller.rb
```diff
@@ -14,12 +14,14 @@ def index
 
   # GET /cases/1/products/new
   def new
+    authorize @investigation, :update?
```
slorek, line 17: Do we need to duplicate this line for each action, or can we move it to a `before_action`?

psd-web/app/views/investigations/tabs/_businesses.html.slim
```diff
@@ -19,14 +20,16 @@
...
       = link_with_hidden_text_to "View business",
                                  "(#{ib.business.trading_name})",
                                  ib.business,
                                  class: "govuk-link app-block-link govuk-!-margin-right-3"
-      = link_with_hidden_text_to "Remove business",
-                                 "(#{ib.business.trading_name})",
-                                 remove_investigation_business_path(@investigation, ib.business),
-                                 class: "govuk-link app-block-link"
+      - if policy(@investigation).update?(user: current_user)
+        = link_with_hidden_text_to "Remove business",
```
StupidCodeFactory, line 29: Suggestion: would be nicer to use `link_to_if`

psd-web/app/helpers/investigations_helper.rb
```diff
@@ -250,33 +250,53 @@ def suggested_previous_owners
 
   # This builds an array from an investigation which can then
   # be passed as a `rows` argument to the govukSummaryList() helper.
-  def about_the_case_rows(investigation)
+  def about_the_case_rows(investigation:, user:)
```
StupidCodeFactory, line 253: I don't understand why keyword arguments here? 
is it just so that you can pass the arguments in whatever order to that method, or to trade wrong number of argument error for required argument error?
if so: 1) It is not great for consistency, 2) it's a lot more typing to do for zero benefits. 

Keywords argument are only useful when you have 3 or more arguments or, if two arguments, one of them is optional or has a default.
Other than that it is just a waste of typing.

psd-web/app/policies/investigation_policy.rb
```diff
@@ -1,37 +1,30 @@
...
+    @record.owner.blank? || @record.owner.in_same_team_as?(user) || @record.owner == user
   end
 
-  def edit_summary?
-    visible_to(user: @user)
-  end
+  # Ability to add and remove other teams as collaborators, and to set their
+  # permission levels.
+  def manage_collaborators?
+    return false if @record.owner.nil?
```
StupidCodeFactory, line 19: Dunno: isn't owner mandatory?

psd-web/spec/factories/contact.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :contact do
+    association :business, factory: :business
```
StupidCodeFactory, line 3: Suggestion(/Please Change?): Leave this blank and let the test setup inject the business as a dependency. 
consider this very likely scenario:
```ruby
let(:contact) { create(:contact) } 
let(:business) { create(:business, contact: contact)
```
you've now created an extra business!!

We must resists the temptation to make our factories omniscient and omnipotent as this habit will result in slow and brittle test suit, guarantied!!

psd-web/spec/features/editing_business_details_spec.rb
```diff
@@ -0,0 +1,51 @@
+require "rails_helper"
+
+RSpec.feature "Editing business details", :with_stubbed_mailer, :with_stubbed_elasticsearch do
+  let(:user)      { create(:user, :activated) }
+  let(:business)  { create(:business, trading_name: "OldCo", contacts: [create(:contact)]) }
```
StupidCodeFactory, line 5: oh look ^^^ `create(:contact)` this creates an extra business in the db

psd-web/test/controllers/investigations/products_controller_test.rb
```diff
@@ -4,6 +4,11 @@ class Investigations::ProductsControllerTest < ActionDispatch::IntegrationTest
   setup do
     sign_in users(:southampton)
     @investigation = load_case(:one)
+    @investigation.collaborators.create!(
```
StupidCodeFactory, line 7: Suggestion: use the service class `AddTeamToAnInvestigation` you won't have to maintain the business logic in every tests, in this PR this introduces 5 places more where we'll have to maintain the business logic contained in the service, far from ideal!

psd-web/app/controllers/investigations/products_controller.rb
StupidCodeFactory: I know it's very tempting to use `before_action` but they ultimately are as bad active record callbacks. I think this is clearer and less hard to follow.

psd-web/app/helpers/documents_helper.rb
```diff
@@ -8,7 +8,6 @@ def set_parent
     @parent ||= Investigation.find_by!(pretty_id: params[:inquiry]) if params[:inquiry]
     @parent ||= Product.find(params[:product_id]) if params[:product_id]
     @parent ||= Business.find(params[:business_id]) if params[:business_id]
-    authorize @parent, :show? if @parent.is_a? Investigation
```
StupidCodeFactory, line 11: just to make sure: is this now in the view here: https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/606/files#diff-d5d43c7323555bc672ef43650b5c21f9R29

psd-web/app/controllers/documents_flow_controller.rb
```diff
@@ -36,6 +37,10 @@ def update
 
 private
 
+  def authorize_if_attached_to_investigation
+    authorize @parent, :update? if @parent.is_a? Investigation
```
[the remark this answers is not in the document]
frankieroberto: Agree – I was super confused by this controller, as it's used in two completely different places. Left this for a future refactor though.

psd-web/spec/factories/contact.rb
frankieroberto: I agree in general, but our current schema / database model seems to assume that all contacts belong to a single business, and so generating a contact with no business associated seems invalid. (We don't currently enforce the presence of a `business_id` on a contact in the database, but arguably we should)

psd-web/app/controllers/investigations/businesses_controller.rb
frankieroberto: Yeah, this is used for the edit action (part of the WickedWizard). I'll add a comment.

psd-web/app/controllers/investigations/ownership_controller.rb
frankieroberto: Well spotted! This should be using the `change_owner_or_status?` policy.

psd-web/app/controllers/investigations/products_controller.rb
frankieroberto: I'm in two minds about this. If the `authorize` call is needed for every action in the controller, I think it's probably ok to use a `before_action`. But if it differs per action I've tended to have it inline (rather than adding `except` or `only` conditions to the before_filter).

psd-web/spec/requests/investigations/updating_the_status_of_a_case_spec.rb
frankieroberto: Yeah, I don't mind a bit of duplication in tests, personally.

psd-web/app/helpers/documents_helper.rb
frankieroberto: No, this was moved into the controller as a private `authorize_if_attached_to_investigation` method.

psd-web/spec/features/editing_business_details_spec.rb
frankieroberto: Good spot. I'll create the contact separately to avoid this. Shame FactoryBot isn't smart enough to avoid this.

psd-web/app/helpers/investigations_helper.rb
frankieroberto: I've found keyword arguments useful for 2+ arguments, but they definitely get _more useful_ the more arguments you have. Mainly because if you miss an argument, the error tells you what's missing, rather than having to check the method definition.

But 🤷

psd-web/test/controllers/investigations/products_controller_test.rb
frankieroberto: Let's look at this in a future refactor, when converting this test to rspec.

psd-web/app/policies/investigation_policy.rb
```diff
@@ -1,37 +1,30 @@
 class InvestigationPolicy < ApplicationPolicy
-  def show?(user: @user)
-    visible_to(user: user)
+  # Used for all updating of the case, including adding and removing related
+  # records, such as products, businesses and documents, with the exception of
+  # changing the case owner, its status (eg 'open' or 'closed'), and whether
+  # or not it is 'restricted'.
+  def update?(user: @user)
```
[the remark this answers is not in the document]
frankieroberto: Makes sense, I just followed the existing convention in the file here. Going to leave this for a future refactor for now though.

psd-web/app/policies/investigation_policy.rb
```diff
@@ -1,37 +1,30 @@
...
-
-  def change_owner?(user: @user)
-    can_change_owner(user: user)
-  end
-
-  def visibility?(user: @user)
-    visible_to(user: user, private: true)
+  # Ability to change the case owner, the status of the case (eg 'open' or 'closed'),
+  # and whether or not it is 'restricted'.
+  def change_owner_or_status?(user: @user)
```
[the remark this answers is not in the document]
frankieroberto: I think this method name is short enough to be ok. But if it gets any longer then it'd definitely be worth reconsidering.

psd-web/app/views/documents/_document_card.html.slim
```diff
@@ -25,11 +25,13 @@ ruby:
...
-        = link_with_hidden_text_to "Edit #{image_document_text(document)}",
-                                   "(#{title})",
-                                   edit_associated_document_path(parent, document),
-                                   class: "govuk-link app-block-link govuk-!-margin-right-3"
-        = link_with_hidden_text_to "Remove #{image_document_text(document)}",
-                                   "(#{title})",
-                                   remove_associated_document_path(parent, document),
-                                   class: "govuk-link app-block-link"
+
+        - if !parent.is_a?(Investigation) || policy(parent).update?(user: current_user)
```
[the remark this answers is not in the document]
frankieroberto: Agree but let's come back to this when refactoring the controller.

PR #607, opened 2020-05-14
Prevent accidential otp whitelisting on production

slawosz:
Using json from `VCAP_APPLICATION` to make sure that otp can be done. If `VCAP_APPLICATION` would change, whitelisting will stop working, which is expected behavior.


psd-web/app/models/secondary_authentication.rb
```diff
@@ -88,11 +88,23 @@ def increment_attempts_and_try_to_lock
...
       false
     end
   end
+
+  def otp_whitelisting_allowed?
+    uris = JSON(ENV["VCAP_APPLICATION"])["application_uris"]
+    return false if uris.length != 1
+
+    uri = uris.first
+    uri =~ /staging\.product-safety-database\.service\.gov\.uk|psd-pr-\d{3,5}\.london\.cloudapps\.digital/
```
frankieroberto, line 106: Feels a bit weird to me to be hardcoding our staging / review app URL patterns into the application code...

Would an alternative be to have some kind of environment variable flag for a "non-production" (review app or staging) environment? Either `RAILS_ENV` or something else? Alternatively we could rely upon `BASIC_AUTH_USERNAME` which should be set in review apps (and maybe we could enable it for staging too)?

slawosz: Hardcoding apps urls is safest way to do this. Any flag can be set by accident and it will be hard to see it. We could add environment for staging, but it would involve discussion.

scruti: What do you thing about maintaining a list of urls or url-regexps in a YAML config file like is done with the whitelisted emails?
This way it is not as easy as messing up an env flag and requires to commit to change their values. While we don't have hard-coded environments urls in our business logic.

slawosz: This is much better, thanks @scruti

psd-web/spec/models/secondary_authentication_spec.rb
```diff
@@ -59,6 +60,106 @@
...
+
+      let(:application_uris) { %w[foo] }
+
+      let(:vcap_application) do
+        { "application_uris" => application_uris }.to_json
+      end
+
+      before do
+        ENV["WHITELISTED_2FA_CODE"] = whitelisted_otp
+        ENV["VCAP_APPLICATION"] = vcap_application
```
StupidCodeFactory, line 87: Please change: This modify the operating system's the ENV. 
Every configuration in the application is in initialised in `config/application.rb`, and the accessible with
```ruby
Rails.configuration.white_list_two_fa_code = ENV.fetch("WHITELISTED_2FA_CODE", "")
# then in your test
allow(Rails.configuration).to receive(:white_list_two_fa_code).and_return(whatever_you_want_without_modifying_the_system_env)
```
this pattern has been used everywhere else and it would be great to remain consistent!

PR #610, opened 2020-05-15
Clean up forbidden layout

edwardhorsford:
Small tweak to the forbidden layout:

* Move the back link outside of main - it's not appropriate within it, and screws with the layout.
* Reword backlink slightly
* Change text referring to processing team to their new team name

Before:
Screenshot 2020-05-15 at 16 18 03

After:
Screenshot 2020-05-15 at 16 44 03


psd-web/app/views/errors/forbidden.html.slim
```diff
@@ -1,10 +1,11 @@
 = page_title "Access denied"
 
-- if user_signed_in?
-  = link_to "Back to list", investigations_path, class: "govuk-back-link"
+- content_for :after_header do
+  - if user_signed_in?
+    = link_to "Back to case list", investigations_path, class: "govuk-back-link"
```
frankieroberto, line 5: We have a nunjucks-style helper for this now:

```suggestion
    = govukBackLink text: "Back to case list", href: investigations_path
```

PR #613, opened 2020-05-15
Use sentence case for back links

edwardhorsford:
Lots of our back links were incorrectly capitalised.

This uses sentence case for a bunch of them.

Before:
Screenshot 2020-05-15 at 17 53 11

After:
Screenshot 2020-05-15 at 18 01 24


edwardhorsford:
> the indentation on the multiline `= link_to` is a bit off. But we don't have any linter nor decided on how we make our Erb template readable so I'm going to let that slide.

What _should_ the indentation be?

nasirkhanbeis:
@edwardhorsford  wonder if we want to fix this too 
image

edwardhorsford:
@nasirkhanbeis that was addressed already in #612

PR #621, opened 2020-05-18
Combine system tests into feature spec

frankieroberto:
We currently have both a system (browser) test for adding/removing businesses to a case, as well as a feature spec.

This removes the system tests, and incorporates some of those tests (for validation errors, and for removing the business) into the feature spec.


psd-web/spec/features/add_business_spec.rb
```diff
@@ -63,5 +88,10 @@
     expected_contact = [name, job_title, phone_number, email].join(", ")
     expect(page).to have_css("dt.govuk-summary-list__key",   text: "Contact")
     expect(page).to have_css("dd.govuk-summary-list__value", text: expected_contact)
+
+    click_link "Remove business"
+
+    expect_to_be_on_remove_business_page
+    click_button "Remove business"
   end
```
nasirkhanbeis, line 96: expect_confirmation_banner("Business was successfully removed.")

PR #630, opened 2020-05-19
Add restriction to case overview

edwardhorsford:
This adds case restriction status as a row on the case overview, removing the feature from 'add activity'.

I've branched off of #606 to get the new restriction policies for this.

Shows the current restriction status, and if the viewing user has rights to change status, a change link.

Screenshot:

Screenshot 2020-05-19 at 15 39 56

I could do with some help adding tests (as needed) for this.


psd-web/app/views/investigations/tabs/_overview.html.erb
```diff
@@ -32,16 +32,24 @@
     <%= govuk_hr %>
     <h2 class="govuk-heading-m">Sharing and collaboration</h2>
 
-    <% if policy(@investigation).change_owner_or_status? %>
+       <% if policy(@investigation).change_owner_or_status? %>
```
scruti, line 35: This tab change seems wrong.

edwardhorsford:
Closing in favour of #641

PR #631, opened 2020-05-19
Generate activity log views for case created activities dynamically

slorek:
https://trello.com/c/3HP9NsVf/554-hide-enquiry-source-contact-details-in-activity-log

## Description
At a high level this implements a tweak whereby when a case is created, in the activity log we want to hide only the complainant's details (due to data protection) from users who should not have access.

At a low level I have modified the way activity log entries are generated. Rather than pre-generating a mix of Markdown and HTML and storing this in the database, we now store just the metadata required to construct the HTML as an arbitrary JSON object, and render it in the view dynamically. This means we can alter the way these are displayed much more easily and don't need to store HTML in the database.

I also converted the Slim template to ERB (but I didn't attempt to improve the markup here).

I had to remove three system tests as they were failing locally but not on CI. Two of them have been rewritten as feature specs (with considerable speed improvement - and fixed a bug where the flash message was not working) and the third was redundant, already implemented in a feature spec.

**There's plenty of opportunity to improve things in future, for instance the HTML markup, and a lot of things in the code, but I haven't attempted to fix everything I found here. These can be fixed in subsequent PRs.**

As part of this I fixed a bug related to the way the owner name was displayed (it was always shown in the context of the same user, rather than the viewing user, so it never showed their team name), and in doing so removed a lot of references to `User.current` in the model. Now the `current_user` is passed in from the view layer and passed along when needed. This is another small step towards removing `User.current` completely.

I have not applied this change to all of the various activity types. The intent of this change is to establish a pattern for doing so. We had reasonable test coverage already of this particular activity type, but this is not true of all of them. Assuming this change is accepted we can then roll it out to the other types. This will provide more opportunities to clean up the code considerably, but until then we need to support both ways of generating activities.

I have also not attempted to migrate old database records as I think parsing these is likely to be risky and difficult to verify that all have been read correctly. Therefore we need to support reading the title/body from the database as well as rendering it dynamically, at least for now.

I think as a next step it would be fairly easy to remove the ActiveRecord callbacks which create this activity, but I have left that for a separate change to keep this PR as small and focused as possible.

## Screenshots

Before (TS user viewing an OPSS-owned enquiry):

Screenshot 2020-05-21 at 14 17 28

After:

Screenshot 2020-05-21 at 14 16 03

Same view as case owner:

Screenshot 2020-05-21 at 14 15 33

## Checklist:


psd-web/app/controllers/investigations/correspondence_controller.rb
```diff
@@ -21,7 +21,7 @@ def create
       attach_files
       save_attachments
       audit_class.from(@correspondence, @investigation)
-      redirect_to investigation_path @investigation, flash: { success: "Correspondence was successfully recorded" }
+      redirect_to investigation_path(@investigation), flash: { success: "Correspondence was successfully recorded" }
```
slorek, line 24: This fixes the flash message not working.

psd-web/app/models/investigation.rb
```diff
@@ -146,10 +146,10 @@ def reported_reason
     @reported_reason ||= ActiveSupport::StringInquirer.new(super)
   end
 
-  def child_should_be_displayed?
+  def child_should_be_displayed?(user)
     # This method is responsible for white-list access for assignee and their team, as described in
     # https://regulatorydelivery.atlassian.net/wiki/spaces/PSD/pages/598933517/Approach+to+case+sensitivity
-    owner.in_same_team_as?(User.current)
```
slorek, line 152: This is why we are passing `user` around everywhere.

psd-web/app/models/user_source.rb
```diff
@@ -5,7 +5,7 @@ def show(viewer = nil)
     user.present? ? user.decorate.display_name(viewer: viewer) : "anonymous"
   end
 
-  def user_has_gdpr_access?(user: User.current)
```
slorek, line 8: This is why we are passing `user` around everywhere.

psd-web/app/views/investigations/activities/_restricted.html.erb
```diff
@@ -0,0 +1,6 @@
+<%
+  creator = activity.source&.user&.organisation&.name || activity.source&.show
```
slorek, line 2: This is copied from the previous Slim template - can be refactored later.

psd-web/app/views/investigations/activities/investigation/_add_allegation.html.erb
```diff
@@ -0,0 +1,31 @@
+<strong>Allegation details</strong>
```
slorek, line 1: All the markup is as was in the previous Markdown model implementation. This can be improved later.

psd-web/spec/features/record_email_activity_spec.rb
```diff
@@ -0,0 +1,213 @@
+require "rails_helper"
+
+RSpec.feature "Adding a record email activity to a case", :with_stubbed_elasticsearch, :with_stubbed_antivirus, :with_stubbed_mailer, type: :feature do
```
slorek, line 3: This is unrelated but fixes a flaky system test.

psd-web/spec/features/record_phone_call_activity_spec.rb
```diff
@@ -0,0 +1,204 @@
+require "rails_helper"
+
+RSpec.feature "Adding a record phone call activity to a case", :with_stubbed_elasticsearch, :with_stubbed_antivirus, :with_stubbed_mailer, type: :feature do
```
slorek, line 3: This is unrelated but fixes a flaky system test.

psd-web/app/models/audit_activity/investigation/add.rb
```diff
@@ -2,31 +2,67 @@ class AuditActivity::Investigation::Add < AuditActivity::Investigation::Base
...
-    details += "<br>Name: **#{sanitize_text complainant.name}**" if complainant.name.present?
-    details += "<br>Type: **#{sanitize_text complainant.complainant_type}**" if complainant.complainant_type.present?
-    details += "<br>Phone number: **#{sanitize_text complainant.phone_number}**" if complainant.phone_number.present?
-    details += "<br>Email address: **#{sanitize_text complainant.email_address}**" if complainant.email_address.present?
-    details += "<br><br>#{sanitize_text complainant.other_details}" if complainant.other_details.present?
-    details
+  def self.build_metadata(investigation)
+    metadata = {
+      owner_id: investigation.owner&.id,
+      complainant_id: investigation.complainant&.id,
```
slorek, line 14: I debated whether to include the full user/complainant details here in case they changed, but I don't think this is necessary,

psd-web/app/models/audit_activity/investigation/add.rb
```diff
@@ -2,31 +2,67 @@ class AuditActivity::Investigation::Add < AuditActivity::Investigation::Base
...
-  def self.build_complainant_details(complainant)
-    details = "<br><br>**Complainant**<br>"
-    details += "<br>Name: **#{sanitize_text complainant.name}**" if complainant.name.present?
-    details += "<br>Type: **#{sanitize_text complainant.complainant_type}**" if complainant.complainant_type.present?
-    details += "<br>Phone number: **#{sanitize_text complainant.phone_number}**" if complainant.phone_number.present?
-    details += "<br>Email address: **#{sanitize_text complainant.email_address}**" if complainant.email_address.present?
-    details += "<br><br>#{sanitize_text complainant.other_details}" if complainant.other_details.present?
-    details
+  def self.build_metadata(investigation)
+    metadata = {
```
slorek, line 12: I wonder whether it is worth versioning with a number these payloads, in case we decide to change their structure later.

psd-web/app/views/investigations/activities/investigation/_add_allegation.html.erb
```diff
@@ -0,0 +1,31 @@
...
+
+<br>
+
+<% if activity.metadata['investigation']['coronavirus_related'] %>
```
scruti, line 5: I am a little bit lost on what calls what, but can be 100% sure that `activity.metadata['investigation']` won't be `nil` here (and in following calls & partials) ? 
Thinking about the possibility of `NoMethodError (undefined method `[]' for nil:NilClass)` and maybe switching to `dig` to avoid that.

psd-web/app/models/audit_activity/investigation/add.rb
scruti: Guess migrations will be enough in our case? I haven't a strong opinion here.

psd-web/app/views/investigations/activities/investigation/_add_allegation.html.erb
slorek: Yes at the moment it will always be set.

PR #636, opened 2020-05-20
Update flash message content

edwardhorsford:
Updates the flash messages for several different investigation activities.

Also fixes a bug where a flash message wasn't shown

Introduces a flash message when adding a team to a case.

### DIFF
--- a/psd-web/app/controllers/collaborators_controller.rb
+++ b/psd-web/app/controllers/collaborators_controller.rb
@@ -26,7 +26,7 @@ class CollaboratorsController < ApplicationController
     )
 
     if result.success?
-      redirect_to investigation_collaborators_path(@investigation), flash: {success: "#{result.collaborator.team.name}  added to the case"}
+      redirect_to investigation_collaborators_path(@investigation), flash: { success: "#{result.collaborator.team.name}  added to the case" }
     else
       @teams = teams_without_access
       @collaborator = result.collaborator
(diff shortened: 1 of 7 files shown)


psd-web/app/controllers/collaborators_controller.rb
```diff
@@ -26,7 +26,7 @@ def create
     )
 
     if result.success?
-      redirect_to investigation_collaborators_path(@investigation)
+      redirect_to investigation_collaborators_path(@investigation), flash: { success: "#{result.collaborator.team.name}  added to the case" }
```
scruti, line 29:
```suggestion
      redirect_to investigation_collaborators_path(@investigation), flash: { success: "#{result.collaborator.team.name} added to the case" }
```

psd-web/app/controllers/collaborators_controller.rb
```diff
@@ -26,7 +26,7 @@ def create
     )
 
     if result.success?
-      redirect_to investigation_collaborators_path(@investigation)
+      redirect_to investigation_collaborators_path(@investigation), flash: { success: "#{result.collaborator.team.name} added to the case" }
```
StupidCodeFactory, line 29: potential XSS exploit. Probably good to check if this gets automatically escaped

StupidCodeFactory: I think it should be fine but worth checking. in which case quite a few flash message might have potential xss exploit

PR #642, opened 2020-05-22
Fix broken styling on activities

edwardhorsford:
#631 inadvertently changed the styling of activities. This fixes it.

After:
Screenshot 2020-05-22 at 12 19 11

Before:
Screenshot 2020-05-22 at 12 19 19

Other fixes:
 * Uses paragraphs where appropriate rather than `<br>`
 * Moves complainant type out of restricted area as it's not meant to be hidden.
 * Reformats the complainant details to use a single paragraph.

### DIFF
--- a/psd-web/app/views/investigations/activities/_activity_card.html.erb
+++ b/psd-web/app/views/investigations/activities/_activity_card.html.erb
@@ -6,7 +6,7 @@
 <li>
   <h3 class="govuk-heading-s"><%= title %></h3>
 
-  <p class="govuk-body-s.timeline-details"><%= activity.subtitle %>
+  <p class="govuk-body-s timeline-details"><%= activity.subtitle %></p>
 
   <% if restricted %>
     <%= render "restricted", activity: activity %>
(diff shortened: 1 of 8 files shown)


psd-web/app/views/investigations/activities/_activity_card.html.erb
```diff
@@ -6,7 +6,7 @@
 <li>
   <h3 class="govuk-heading-s"><%= title %></h3>
 
-  <p class="govuk-body-s.timeline-details"><%= activity.subtitle %>
+  <p class="govuk-body-s timeline-details"><%= activity.subtitle %></p>
```
StupidCodeFactory, line 9: 🤣

psd-web/app/views/investigations/activities/investigation/_complainant.html.erb
```diff
@@ -1,33 +1,40 @@
...
+    <% complainant_info << "Other details: <strong>#{escape_once(complainant.other_details)}</strong>" %>
   <% end %>
 
-  <% if complainant.other_details %>
-    <br><br>
-    <%= complainant.other_details %>
+  <% if complainant_info %>
+    <p class="govuk-body">
+      <%= complainant_info.join("<br />").html_safe %>
+    </p>
```
slorek, line 37: I think it would be better to leave this as it was. No need for Arrays or `escape_once`.

edwardhorsford: As is it doesn't correctly handle when information isn't present - there should be a break between each item, but you need all items to know where to put the breaks - and the wrapping paragraph.

PR #650, opened 2020-05-26
added flash message test step

nasirkhanbeis:
Added flash message test within existing tests
## Description

With the recent change on flash message  content, added test step within existing test to validate flash messages on different occasions such as change case owner, coronavirus flag update etc. 

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


psd-web/spec/features/investigation_edit_spec.rb
```diff
@@ -22,9 +22,16 @@
     choose "Yes, it is (or could be)"
     click_on "Update coronavirus status"
 
+    expect_confirmation_banner("Coronavirus status updated on project")
     expect(page).to have_css(".govuk-summary-list__row .govuk-summary-list__key + .govuk-summary-list__value", text: "Coronavirus related case")
   end
 
+  scenario "user from other team cannot edit coronavirus flag" do
```
StupidCodeFactory, line 29: this seems like a very expensive test for such a small check, I wonder this could not be checked in another test where, for example, this `other_user` is used.

nasirkhanbeis: @StupidCodeFactory 
agree but just wanted to add along with relevant test and we agreed to write view only tests in wherever existing specs since we started whole permissions stuff

nasirkhanbeis: agree, done!

PR #651, opened 2020-05-26
Fix flash message showing wrong name

edwardhorsford:
The change of owner flash message was showing the previous owner rather than the new owner.


slorek:
Please add tests from #650.

PR #653, opened 2020-05-26
Only display complainant contact details on case summary page to teams added to the case

slorek:
https://trello.com/c/G1tdeHxB/547-2-dont-show-enquiry-contact-details-to-teams-not-involved-in-case

## Description
This displays a hint text and conditionally displays the contact details of a complainant only if the user is a member of a team on the case (which may be the owner's team or a collaborating team).

I confirmed with @edwardhorsford that this should also apply to allegations as well as enquiries. The hint text changes accordingly.

There's a bit of existing code duplication on the feature specs which I've made a bit worse - I haven't addressed this as it would need a bit more refactoring of the test and creation of some generic helpers, which wasn't directly related. We can do that in a separate pull request.

#### Before:
Screenshot 2020-05-26 at 14 31 27

#### After (Team added to case)
Screenshot 2020-05-26 at 11 42 17

#### After (Team not on case)
Screenshot 2020-05-26 at 11 39 11

## Checklist:

### DIFF
--- a/psd-web/app/decorators/investigation_decorator.rb
+++ b/psd-web/app/decorators/investigation_decorator.rb
@@ -42,11 +42,12 @@ class InvestigationDecorator < ApplicationDecorator
   end
 
   def source_details_summary_list(user)
-    contact_details = if complainant.can_be_displayed?(user)
-                        complainant.decorate.contact_details
-                      else
-                        "Reporter details are restricted because they contain GDPR protected data."
-                      end
+    contact_details = h.tag.p(I18n.t("case.protected_details", data_type: "#{object.case_type} contact details"), class: "govuk-hint")
+
+    if Pundit.policy(user, object).view_protected_details?
+      contact_details << complainant.decorate.contact_details
+    end
+
     rows = [
       should_display_date_received? ? { key: { text: "Received date" }, value: { text: date_received.to_s(:govuk) } } : nil,
       should_display_received_by? ? { key: { text: "Received by" }, value: { text: received_type.upcase_first } } : nil,
--- a/psd-web/app/decorators/investigation_decorator.rb
+++ b/psd-web/app/decorators/investigation_decorator.rb
@@ -41,12 +41,9 @@ class InvestigationDecorator < ApplicationDecorator
     h.render "components/govuk_summary_list", rows: rows, classes: "govuk-summary-list--no-border"
   end
 
-  def source_details_summary_list(user)
+  def source_details_summary_list(view_protected_details = false)
     contact_details = h.tag.p(I18n.t("case.protected_details", data_type: "#{object.case_type} contact details"), class: "govuk-hint")
-
-    if Pundit.policy(user, object).view_protected_details?
-      contact_details << complainant.decorate.contact_details
-    end
+    contact_details << h.tag.p(complainant.decorate.contact_details) if view_protected_details
 
     rows = [
       should_display_date_received? ? { key: { text: "Received date" }, value: { text: date_received.to_s(:govuk) } } : nil,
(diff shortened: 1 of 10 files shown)


psd-web/app/decorators/investigation_decorator.rb
```diff
@@ -42,11 +42,12 @@ def product_summary_list
   end
 
   def source_details_summary_list(user)
-    contact_details = if complainant.can_be_displayed?(user)
-                        complainant.decorate.contact_details
-                      else
-                        "Reporter details are restricted because they contain GDPR protected data."
-                      end
+    contact_details = h.tag.p(I18n.t("case.protected_details", data_type: "#{object.case_type} contact details"), class: "govuk-hint")
```
slorek, line 45: I realise that there is a school of thought that decorators should not output HTML, however the existing decorator method was already outputting HTML. I did look into how we could refactor this but it would be quite a difficult change and cannot be easily unit tested, so I've left this for future discussion and iteration, since I'm not entirely convinced that moving this logic to the view or a helper would be an improvement, and the new policy introduced unblocks some other tickets.

psd-web/app/decorators/investigation_decorator.rb
```diff
@@ -42,11 +42,12 @@ def product_summary_list
...
+
+    if Pundit.policy(user, object).view_protected_details?
```
slorek, line 47: We could consider adding `include Pundit` in `ApplicationDecorator` since they are currently part of the view layer (outputting HTML). This is a point for discussion.

psd-web/app/decorators/investigation_decorator.rb
```diff
@@ -42,11 +42,12 @@ def product_summary_list
...
+      contact_details << complainant.decorate.contact_details
```
StupidCodeFactory, line 48: instead of doing decorating the complainant object here, add it to the list of automatically decorated associations.
https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/decorators/investigation_decorator.rb#L3

psd-web/app/policies/investigation_policy.rb
```diff
@@ -36,6 +36,16 @@ def view_non_protected_details?(user: @user, private: @record.is_private)
     false
   end
 
+  def view_protected_details?(user: @user)
+    # Is the user a member of the case owner's team?
+    return true if @record.owner.present? && (@record.owner&.team == user.team)
```
StupidCodeFactory, line 41:
```suggestion
    return true if @record.owner&.team == user.team
```
I think this is enough. as `user` **must** have a team. Worst case scenario this should always be false `nil == instance_of(Team)`

edwardhorsford: I think the rights needed are the same as on L7. We could either reuse that, or use the same code?

We don't need to test the owner - it's purely about whether the current user's team is in `record.teasm_with_access` I think.

slorek: This was copied from the preceding `view_non_protected_details?` policy for consistency but  I'll look at this.

psd-web/app/decorators/investigation_decorator.rb
slorek: Yes I considered injecting this, I agree it is a better approach.

psd-web/spec/decorators/investigation_decorator_spec.rb
```diff
@@ -137,19 +138,46 @@
...
     end
 
-    it "displays the Complainant email address" do
-      expect(source_details_summary_list).to summarise("Contact details", text: /#{Regexp.escape(investigation.complainant.email_address)}/)
+    context "when the user is not on a collaborating team" do
+      let(:viewing_user) { create(:user) }
+
+      it "does not display the Complainant details", :aggregate_failures do
+        expect_to_display_protect_details_message
+        expect(source_details_summary_list).not_to summarise("Contact details", text: /#{Regexp.escape(investigation.complainant.name)}/)
```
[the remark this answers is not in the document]
slorek: Regexp.escape returns a string: https://ruby-doc.org/core-2.6.6/Regexp.html#method-c-escape

PR #656, opened 2020-05-27
Supporting informations page

StupidCodeFactory:
## CHANGES:

I added a table of supporting information, Supporting information is a collection of 
`Correspondence::Email`, `Correspondence::PhoneCall`, `Correspondence::Meeting`, `Test::Result` and `CorrectiveAction`

At the bottom are the `Generic attachments`: attachments that are directly linked to a case but and therefore not linked to any of the supporting information
 
Things to watch out for, link to the view page for corrective actions will be removed in the table if the viewing user does not have access to view restricted details.


psd-web/app/controllers/investigations/supporting_information_controller.rb
```diff
@@ -7,7 +7,13 @@ def index
       authorize investigation, :view_non_protected_details?
 
       @investigation                = investigation.decorate
-      @supporting_information       = investigation.supporting_information_attachments.decorate
+      @supporting_information       = investigation.supporting_information_attachments.includes(:blob).decorate
```
frankieroberto, line 10: @StupidCodeFactory this looks like it's only returning attachments associated with "supporting information", but I think the intent is to list all supporting information records, whether they have attachments or not. For example, if you record a meeting or an email, but _don’t_ include any file attachments, then they should still be listed.

I think we’d need something like this:

```ruby
class Investigation
  def supporting_information
    corrective_actions + correspondences + tests
  end
end
```

...which wouldn’t be super efficient, as this is combining 3 database calls, but probably ok for now? Ordering would have to be done in Ruby rather than Postgres too.

StupidCodeFactory: In the story it does say "We'll have two tables - one for attachments, one for activities". 
Now after discussing with @edwardhorsford the other table "@other_supporting_information" are for attachments that aren't images but attached on the case and not an image.
So seems like the correct implementation to me.

psd-web/app/decorators/test/result_decorator.rb
```diff
@@ -0,0 +1,40 @@
...
+      result_text = if passed?
+                      "Passed test"
+                    elsif failed?
+                      "Failed test"
+                    else
+                      "Test result"
+                    end
+
+      "#{result_text}: #{product.name}"
+    end
```
StupidCodeFactory, line 14: This method was copied over from the TestDecorator. It was moved because it didn't work if you decorated a `Test::Request` which is a subclass of `Test`.

StupidCodeFactory:
[(5) Table view for supporting information](https://trello.com/c/hySkYqQe/494-5-table-view-for-supporting-information)

nasirkhanbeis:
looks good to me!

psd-web/app/views/investigations/tabs/_supporting_information.erb
```diff
@@ -18,11 +18,12 @@
 <% if @supporting_information.any? %>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
+      <h2>Documents</h2>
```
edwardhorsford, line 21:
```suggestion
      <h2 class="govuk-heading-m">Documents</h2>
```

PR #660, opened 2020-05-27
Fix migration

slawosz:
Proper class.


psd-web/db/migrate/20200514123702_add_collaborators_references_to_collaborations.rb
```diff
@@ -7,7 +7,7 @@ def change
       end
       rename_column :collaborations, :team_id, :collaborator_id
       Collaboration.reset_column_information
-      Collaboration.update_all(collaborator_type: "Team", type: "Collaborator::EditAccess")
+      Collaboration.update_all(collaborator_type: "Team", type: Collaboration::EditAccess.to_s)
```
StupidCodeFactory, line 10: don't need the `to_s`

PR #664, opened 2020-05-28
Don't attach activities attachments into investigation documents

scruti:
The documents attached to activities (communications, testing results, corrective actions...) get also attached to the associated investigation.
This is causing issues when having to distinguish between the attachments added directly to the document and the attachments added to one of the document activities.

This PR does:
- 1. Stops re-attaching the activities attachments o the investigations. The attachment will be associated only with the activity entity itself.
- 2. Adds a rake task to be able to remove from the existent investigation documents all those activity attachments that where duplicated into the investigations.

Note: Do not confuse attachments belonging to activities ( with record type: Test, Communication...) with the attachments associated to the activity log (with record type: Activity).

## Checklist:


psd-web/lib/tasks/attachments.rake
```diff
@@ -0,0 +1,18 @@
...
+  # 'Activity' record type corresponds to the activity log entry. Activity
+  # log entries are registering the attachment being added to the Investigation.
+  # When this happen the Activity entry will the same attachment blob as the
+  # Investigation.
+  desc "Delete unrelated attachments from investigations"
+  task delete_unrelated_from_investigations: :environment do
+    ActiveStorage::Attachment
+      .where(record_type: "Investigation")
+      .where(
+        blob_id: ActiveStorage::Attachment.where.not(record_type: %w[Investigation Activity])
```
StupidCodeFactory, line 15: This does not look correct. You probably wanted to look up another table `ActiveStorage::Blob` Also what of `Investigation::Allegation` , `Investigation::Enquiry`, `Investigation::Project`? 
Also this will delete attachments on products and businesses no?

scruti: argh copied this wrong from the manual execution.
The line was: `blob_id: ActiveStorage::Attachment.where.not(record_type: %w[Investigation Activity]).pluck(:blob_id)`

And the resulting SQL would be:
```
SELECT "active_storage_attachments"."blob_id" FROM "active_storage_attachments" WHERE "active_storage_attachments"."record_type" NOT IN ($1, $2)  [["record_type", "Investigation"], ["record_type", "Activity"]]

  ActiveStorage::Attachment Destroy (10.7ms)  DELETE FROM "active_storage_attachments" WHERE "active_storage_attachments"."record_type" = $1 AND "active_storage_attachments"."blob_id" IN ($2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20), [["record_type", "Investigation"],...```
First would get the list of ids from the activity blobs in the attachments table. Then delete from the investigation attachments all the ones that are pointing to a blob shared with Activities attachments.

For Products and Business, their blobs are not being re-attached to the Investigations, so I just omitted them. Will re-visit and test anyway.

**Comment by scruti** [comment]
Moving it to WIP to cover the rake task with tests and to display the activity attachments in the Investigation attachments tab.

File: psd-web/spec/support/shared_contexts/rake.rb [hunk 2]
```diff
@@ -0,0 +1,22 @@
...
+  def loaded_files_excluding_current_rake_file
+    $LOADED_FEATURES.reject { |file| file == Rails.root.join("#{task_path}.rake").to_s }
+  end
+
+  before do
+    Rake.application = rake
+    Rake.application.rake_require(task_path, [Rails.root.to_s], loaded_files_excluding_current_rake_file)
+
+    Rake::Task.define_task(:environment)
+  end
```
StupidCodeFactory, line 21: I think you've over complicated yourself, all you need is:
ruby```
subject(:task) { Rake::Task[task_name] }
before { Rails.application.load_tasks }

it { expect(task.invoke).to have_done_something }
```

**Comment by scruti** [comment]
Displaying the activities attachments wasn't feasible as would have lots of conflicts with other peoples work done in parallel. So decided to leave this ready to go after their work gets released.

File: psd-web/lib/tasks/attachments.rake [hunk 3]
```diff
@@ -0,0 +1,18 @@
...
+  # 'Activity' record type corresponds to the activity log entry. Activity
+  # log entries are registering the attachment being added to the Investigation.
+  # When this happen the Activity entry will the same attachment blob as the
+  # Investigation.
+  desc "Delete activities attachments from investigations"
+  task delete_activities_from_investigations: :environment do
+    ActiveStorage::Attachment
+      .where(record_type: "Investigation")
+      .where(
+        blob_id: ActiveStorage::Attachment.where.not(record_type: %w[Investigation Activity]).pluck(:blob_id)
```
StupidCodeFactory, line 15: I think this is not quite right:
The activity should really point to the investigations attachment that was added at that time. This will still keep duplicates.
In reality Activity ought to be polymorphic association, like `auditable` (maybe)?
They would still remain attached to an investigation but I think there is a need to retain and extra model added to an investigation, like business, product, attachments and such. 
Rather than have plenty of foreign keys that are useless most of the time.
I realise that this is digging down a rabbit whole
@slorek what's your take on this?

StupidCodeFactory: I'm not suggesting to this now, but maybe address this soon, now that we are tinkering with all that mess. It also causes me headaches because supporting information and images are duplicated and I need to come up with some strange SQL query to filter out stuff.

scruti: I think the bit about "Activity" duplicates is a totally different beast and this PR didn't mean to solve all the attachments duplication around the codebase, but to "remove things we shouldn't have attached to investigations", in order to be able to get just "generic attachments" from the investigation.

Activity log attachments being duplicated from investigations will need to be addressed too, maybe in the feature branch itself? But rather have it done in its own PR as would imply data model changes as you suggested :+1:

PR #665, opened 2020-05-28
Prevent instantiation of Investigation class

slorek:
## Description
This prevents an `Investigation` from being instantiated or saved, since cases should always be an instance of one of its subclasses.

This was primarily to ensure our tests more closely reflect real data; I noticed some slightly odd behaviour when writing a test with an investigation factory rather than an allegation.

This is a small step toward creating a service which is responsible for creating the investigation, setting default data, adding audit log entries and sending emails, rather than callbacks. We can then use that in our tests instead.

## Checklist:

### DIFF
--- a/psd-web/app/models/investigation.rb
+++ b/psd-web/app/models/investigation.rb
@@ -17,6 +17,8 @@ class Investigation < ApplicationRecord
 
   before_validation { trim_line_endings(:user_title, :description, :non_compliant_reason, :hazard_description) }
 
+  validates :type, presence: true # Prevent saving instances of Investigation; must use a subclass instead
+
   validates :description, presence: true, on: :update
   validates :owner_id, presence: { message: "Select case owner" }, on: :update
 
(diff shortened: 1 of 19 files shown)


psd-web/app/models/investigation.rb
```diff
@@ -65,6 +67,12 @@ class Investigation < ApplicationRecord
   before_create :set_source_to_current_user, :set_owner_as_current_user, :add_pretty_id
   after_create :create_audit_activity_for_case, :send_confirmation_email
 
+  def initialize(*args)
```
StupidCodeFactory, line 70: Rails has already solved this problem. Simply add 
```ruby
  self.abstract_class = true
end
```
then no need for `validates :type, presence: true` 
if you try to instanciate a factory with `create(:investigation)` and it will raise: 
```
     NotImplementedError:
       Investigation is an abstract class and cannot be instantiated.
```
I think i would also be good to remove the default on the type column of `Investigation::Allegation` and make it a not null column if it is not already.

slorek: @StupidCodeFactory I have already tested this; `abstract_class` is not intended to be used with STI. There were all sorts of issues with it.

PR #673, opened 2020-06-01
Remove unused helper methods

frankieroberto:
These aren't called from anywhere, so deleting.


frankieroberto:
Moving back to draft as @edwardhorsford has indicated we may actually want to use these helpers.

frankieroberto:
Closed in favour of #676

PR #677, opened 2020-06-01
Convert Slim partials to ERb

frankieroberto:
This is a straight conversion, apart from also fixing a `<p>` which was nested within an `<h1>` (which is invalid).

### DIFF
--- a/psd-web/app/views/application/_breadcrumbs_navigation.html.erb
+++ b/psd-web/app/views/application/_breadcrumbs_navigation.html.erb
@@ -1,0 +1,9 @@
+<% if breadcrumbs[:is_simple_link] %>
+  <%= link_to breadcrumbs[:text], breadcrumbs[:href], class: "govuk-back-link" %>
+<% else %>
+  <%= render "components/govuk_breadcrumbs",
+    items: breadcrumbs[:items],
+    classes: local_assigns[:classes],
+    attributes: local_assigns[:attributes]
+  %>
+<% end %>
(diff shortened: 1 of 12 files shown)


psd-web/app/views/application/_product_heading.html.erb
```diff
@@ -0,0 +1,7 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <span class="govuk-caption-l">Product</span>
+    <h1 class="govuk-heading-l"><%= product.name %></h1>
+    <p class="gov-uk-caption-m"><%= product.created_at.to_s(:govuk) %></p>
```
edwardhorsford, line 5:
```suggestion
    <p class="govuk-caption-m"><%= product.created_at.to_s(:govuk) %></p>
```

psd-web/app/views/application/_breadcrumbs_navigation.html.erb
```diff
@@ -0,0 +1,9 @@
+<% if breadcrumbs[:is_simple_link] %>
+  <%= link_to breadcrumbs[:text], breadcrumbs[:href], class: "govuk-back-link" %>
+<% else %>
+  <%= render "components/govuk_breadcrumbs",
+    items: breadcrumbs[:items],
+    classes: local_assigns[:classes],
```
StupidCodeFactory, line 6: Can't we just use the local variables `items` and `classes` instead of fetching them from the `local_assigns` hash? bit less verbose and unnecessary coupling to the `Hash` syntax.

frankieroberto: I presume this was originally done because not all templates which render this partial set all the variables? Agee we should fix this though, but let’s consider that separately and keep this PR to purely converting the templates.

PR #681, opened 2020-06-02
Extract callbacks from Investigation creation into service

slorek:
## Description
This removes the callbacks and dependency on `User.current` in the model when creating a case.

It is a stepping stone rather than an example of perfect separation of concerns - it only replaces the `save` step in the creation of the record, and still requires some manual construction of the `Investigation` object. This was not tackled in this change as it would require more extensive refactoring of the wizard controllers and existing tests.

### Summary
- Removed model callbacks and dependency on `User.current` when a case is created. Instead of `investigation.save` we now need to call the `CreateCase` service.
- Removed the copying of the first attachment on a case to the case created activity, as agreed
- New `creator` attribute on factories to replace the need to set `User.current` in tests.
- The new service requires an instance of `investigation`, since the wizard controllers would need extensive refactoring to fully move case creation into the service. Also the tests would need extensive reworking.
- Saving the `Investigation` instance is necessary to generate the `pretty_id` as the generated number depends on the number of records in the table. Without saving each record, duplicate IDs would be generated. I had to use a callback on the factory to integrate this with the tests and had to change some usage of `build` to `create` accordingly, but this has not impacted test run time appreciably. This callback approach provides a more convenient interface in tests.
- Removed support for `owner_id` to be `nil` as this would require some additional work on the factory which would be quite convoluted. I checked and there are no instances of a case with a nil `owner_id`. There is an imminent change to move `owner` to `collaborators` so we can think about adding a constraint then.
- It was necessary to remove the callbacks when adding a product to a case as well. This was due to the race conditions in the audit activity callbacks depending on each other being saved. I've replaced these with services for adding/removing products and ensured that when a product is added to an investigation on creation, the correct behaviour is observed (i.e. no 'case updated' emails are sent - previously these were generated albeit without any recipients due to the particular order of the callbacks before)
- Added an advisory transaction lock when generating the `pretty_id` to eliminate a potential race condition in the previous implementation, and uniqueness/not null constraints to ensure data integrity.
- Removed dependency on `User.current` when creating seeds. Seeded cases are now created by a dummy 'Seed User'. This is because each environment has different ways of creating users and we need a way that works both locally and in review apps. The seeds need extensive reworking anyway to create more representative data so this can be improved later.
- `User.current` is still required for generation of other types of audit activity records and there's still a number of callbacks in various models. I believe the patterns introduced here can help us quickly eliminate those and in turn remove all the excess code from the activity models. The changes in this PR are just a step towards this and I will implement the next round of refactoring after this is merged.

## Checklist:


psd-web/app/models/activity.rb
```diff
@@ -47,21 +47,6 @@ def template_name
...
-  def subtitle_slug; end
-
-  def notify_relevant_users
-    entities_to_notify.each do |entity|
-      email = entity.is_a?(Team) ? entity.team_recipient_email : entity.email
-      NotifyMailer.investigation_updated(investigation.pretty_id, entity.name, email, email_update_text(entity), email_subject_text).deliver_later
-    end
-  end
-
   def entities_to_notify
```
slorek, line 50: This file is unchanged except to make `entities_to_notify` a public method so that the services can access it. Eventually I think we will want to move the case update email sending to a separate service, and remove all of this code from the Activity model and the individual services.

psd-web/app/models/audit_activity/investigation/add.rb
```diff
@@ -1,15 +1,10 @@
 class AuditActivity::Investigation::Add < AuditActivity::Investigation::Base
-  include ActivityAttachable
-  with_attachments attachment: "attachment"
-
-  private_class_method def self.from(investigation)
-    activity = super(investigation, nil, nil, build_metadata(investigation))
-    activity.attach_blob investigation.documents.first.blob if investigation.documents.attached?
-    activity
+  def self.from(*)
```
slorek, line 2: We need to keep these `from` methods for now while it remains inherited from parent classes. When we move all activity generation to services we can remove them completely.

psd-web/app/models/audit_activity/investigation/add_allegation.rb
```diff
@@ -1,5 +1,2 @@
 class AuditActivity::Investigation::AddAllegation < AuditActivity::Investigation::Add
```
slorek, line 1: I don't see a lot of value in having a different class for each case type as they are all identical apart from the name. We could consider simplifying this sometime..

psd-web/app/models/audit_activity/product/base.rb
```diff
@@ -1,12 +1,7 @@
...
-    create(
-      source: UserSource.new(user: User.current),
-      investigation: investigation,
-      title: title,
-      product: product
-    )
-  end
+  # Do not send investigation_updated mail when product added/removed. This
+  # overrides inherited functionality in the Activity model :(
+  def notify_relevant_users; end
```
slorek, line 6: When we are able to refactor the remaining activity types we can remove all of this email sending code and overrides.

psd-web/app/models/investigation.rb
```diff
@@ -157,12 +145,12 @@ def child_should_be_displayed?(user)
     owner.in_same_team_as?(user)
   end
 
-private
-
-  def create_audit_activity_for_case
+  def case_created_audit_activity_class
```
slorek, line 148: If we only had one class (see earlier comment) we could remove this.

psd-web/app/services/add_product_to_case.rb
```diff
@@ -0,0 +1,40 @@
+class AddProductToCase
+  include Interactor
+
+  delegate :product, :investigation, :user, to: :context
+
+  def call
```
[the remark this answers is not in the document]
slorek: I think these are needed to fail early and document the interface.

psd-web/app/services/add_product_to_case.rb
```diff
@@ -0,0 +1,40 @@
...
+  def create_audit_activity_for_product_added
+    AuditActivity::Product::Add.create(
+      source: UserSource.new(user: user),
+      investigation: investigation,
+      title: product.name,
+      product: product
+    )
+  end
+
+  def send_notification_email
```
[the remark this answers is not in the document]
slorek: We can improve this later by extracting the case updated email sending to a shared service, but this only really makes sense once the code is also removed from `Activity`.

psd-web/app/services/remove_product_from_case.rb
```diff
@@ -0,0 +1,40 @@
+class RemoveProductFromCase
+  include Interactor
+
+  delegate :product, :investigation, :user, to: :context
+
+  def call
```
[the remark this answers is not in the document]
slorek: I think these are needed to fail early and document the interface.

psd-web/app/services/remove_product_from_case.rb
```diff
@@ -0,0 +1,40 @@
...
+  def create_audit_activity_for_product_removed
+    AuditActivity::Product::Destroy.create(
+      source: UserSource.new(user: user),
+      investigation: investigation,
+      title: "Removed: #{product.name}",
+      product: product
+    )
+  end
+
+  def send_notification_email
```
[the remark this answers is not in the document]
slorek: We can improve this later by extracting the case updated email sending to a shared service, but this only really makes sense once the code is also removed from `Activity`.

psd-web/app/services/remove_product_from_case.rb
```diff
@@ -0,0 +1,42 @@
...
+    context.activity.entities_to_notify.each do |recipient|
+      email = recipient.is_a?(Team) ? recipient.team_recipient_email : recipient.email
```
slorek, line 31: I think we should consider just using the attribute name `email` on a `Team` so we could remove this.

psd-web/db/seeds.rb
```diff
@@ -19,6 +19,23 @@ def create_blob(filename, title: nil, description: nil)
 if run_seeds
 
   Rails.logger.info("Running seeds.rb")
+
```
slorek, line 22: This ensures the `source` and `owner` are set to something, which is not presently happening. We need to rework the seeds to have more representative data so this can be improved in future.

psd-web/app/services/add_product_to_case.rb
```diff
@@ -0,0 +1,42 @@
+class AddProductToCase
+  include Interactor
+
+  delegate :product, :investigation, :user, to: :context
+
+  def call
+    context.fail!(error: "No product supplied") unless product.is_a?(Product)
+    context.fail!(error: "No investigation supplied") unless investigation.is_a?(Investigation)
+    context.fail!(error: "No user supplied") unless user.is_a?(User)
```
slawosz, line 9: Would be nice to get rid of writing `context`, you could delegate `fail!` to `context.fail!`?

psd-web/app/services/add_product_to_case.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    investigation.products << product
+
+    context.activity = create_audit_activity_for_product_added
```
slawosz, line 13: Delegate context?

psd-web/app/services/add_product_to_case.rb
```diff
@@ -0,0 +1,42 @@
...
+    AuditActivity::Product::Add.create(
+      source: UserSource.new(user: user),
+      investigation: investigation,
+      title: product.name,
+      product: product
+    )
+  end
+
+  def send_notification_email
+    context.activity.entities_to_notify.each do |recipient|
```
slawosz, line 30: Delegate `context`?

psd-web/app/services/create_case.rb
```diff
@@ -0,0 +1,79 @@
+class CreateCase
+  include Interactor
+
+  delegate :investigation, :user, to: :context
```
slawosz, line 4: Delegate all to get rid of `context`?

slorek: @slawosz I've been using `delegate` as a way to kind-of make it obvious what inputs are expected as there is no interface declaration at all in the Interactor `call` methods.

psd-web/app/services/create_case.rb
```diff
@@ -0,0 +1,79 @@
+class CreateCase
```
slawosz, line 1: I would name it `CaseCreator`

psd-web/app/services/remove_product_from_case.rb
slawosz: Is it big change to make?
Isn't it:
```ruby
class Team
  def email
    team_recipient_email
  end
end
```

psd-web/app/services/create_case.rb
```diff
@@ -0,0 +1,79 @@
...
+      user.name,
+      user.email,
+      investigation.decorate.title,
+      investigation.case_type
+    ).deliver_later
+  end
+
+  def date
+    Time.zone.now
+  end
```
frankieroberto, line 74: Suggest deleting this private method and calling [`Date.current`](https://api.rubyonrails.org/classes/Date.html#method-c-current) directly within `#generate_pretty_id` as we don't seem to need the time.

psd-web/app/services/remove_product_from_case.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    context.activity = create_audit_activity_for_product_removed
+
+    send_notification_email
+  end
+
+private
+
+  def create_audit_activity_for_product_removed
+    AuditActivity::Product::Destroy.create(
```
frankieroberto, line 21: Should this raise an exception if it fails to save?

```suggestion
    AuditActivity::Product::Destroy.create!(
```

psd-web/db/migrate/20200604094957_add_constraints_on_investigations.rb
```diff
@@ -0,0 +1,9 @@
+class AddConstraintsOnInvestigations < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured do
+      remove_index :investigations, :pretty_id
+      add_index :investigations, :pretty_id, unique: true
```
frankieroberto, line 5: 👏

psd-web/app/services/remove_product_from_case.rb
slorek: yes, good spot

psd-web/app/services/create_case.rb
slorek: It's used three times - also in `number_of_cases_this_month`.

psd-web/app/services/create_case.rb
```diff
@@ -0,0 +1,79 @@
...
+        create_audit_activity_for_product_added(product)
+      end
+    end
+
+    send_confirmation_email
+  end
+
+private
+
+  def generate_pretty_id
```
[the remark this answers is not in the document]
slorek: I prefer `generate_pretty_id` as it is more descriptive.

psd-web/app/services/create_case.rb
slorek: I think it's a question of consistency with other services. Currently we are using verbs.

psd-web/app/services/create_case.rb
frankieroberto: Ah ok. Maybe just change to Date.current then? (Seems mildly confusing that a method called “date“ returns a time object).

slorek: We need the time component for the `number_of_cases_this_month` to work properly.

slorek: Renamed to `now`.

PR #668, opened 2020-05-29
Add page for test results

frankieroberto:
This adds a new view for "test results", and links to this view from the relevant entries in the activity timeline.

In future it will also be linked to from a "Supporting information" tab.

## Checklist:
- [x] Have you documented your changes in the pull request description?
- [x] Does the change present any security considerations?
- [ ] Has acceptance criteria been tested by a peer?
- [ ] Has the CHANGELOG been updated? (If change is worth telling users about.)

#### Before

Activity timeline links directly to the test result file:

Screenshot 2020-06-01 at 10 46 32

#### After

Activity timeline links to new page for the test result:

Screenshot 2020-06-01 at 10 46 43

New test result page:

Screenshot 2020-06-01 at 16 16 48

PR #684, opened 2020-06-03
Page for showing corrective actions

frankieroberto:
Following on from #668, this adds a new page for viewing "corrective actions" (product recalls, etc). 

This is currently only linked to from the Activity timeline, but will soon be listed and linked to from a "Supporting information" tab, replacing the "Attachments" tab.  Due to existing database model limitations, it is currently only possibly to create the link from the Activity timeline if a file was attached to the corrective action. This can be remedied in a future change.

### Before

Activity timeline entry links only to the attachment:

Screenshot 2020-06-04 at 09 21 40

### After

Activity timeline entry links to the corrective action (only if an attachment was added):

Screenshot 2020-06-04 at 09 21 53

New corrective action page:

Screenshot 2020-06-04 at 09 22 07

### Implementation notes

We have an existing `CorrectiveActionsController` but that uses the `show` action as part of the "add a corrective action" flow. To avoid conflicting with or refactoring that immediately, I’ve added a new `ViewCorrectiveActionsController`, which isn't ideal but is a short-term compromise.

### DIFF
--- a/psd-web/app/views/investigations/activities/_activity_card.html.erb
+++ b/psd-web/app/views/investigations/activities/_activity_card.html.erb
@@ -53,7 +53,7 @@
     <% if activity.is_a? AuditActivity::Test::Result %>
       <%= link_with_hidden_text_to "View test result", " from #{activity.test_result.date.to_s(:govuk)}", investigation_test_result_path(@investigation, activity.test_result.id) %>
 
-    <% elsif activity.is_a?(AuditActivity::CorrectiveAction::Add) && activity.corrective_action%>
+    <% elsif activity.is_a?(AuditActivity::CorrectiveAction::Add) && activity.corrective_action %>
       <%=  link_to "View corrective action", investigation_action_path(@investigation, activity.corrective_action), class: "govuk-link" %>
     <% elsif activity.has_attachment? %>
       <% activity.attachments.each do |name, attachment| %>
--- a/psd-web/app/views/investigations/activities/_activity_card.html.erb
+++ b/psd-web/app/views/investigations/activities/_activity_card.html.erb
@@ -52,6 +52,9 @@
 
     <% if activity.is_a? AuditActivity::Test::Result %>
       <%= link_with_hidden_text_to "View test result", " from #{activity.test_result.date.to_s(:govuk)}", investigation_test_result_path(@investigation, activity.test_result.id) %>
+
+    <% elsif activity.is_a?(AuditActivity::CorrectiveAction::Add) && activity.corrective_action%>
+      <%=  link_to "View corrective action", investigation_action_path(@investigation, activity.corrective_action), class: "govuk-link" %>
     <% elsif activity.has_attachment? %>
       <% activity.attachments.each do |name, attachment| %>
         <%= link_to "View #{name}", attachment, class: "psd-block-link", target: "_blank", rel: 'noopener' %>
--- a/psd-web/app/views/investigations/corrective_actions/show.html.erb
+++ b/psd-web/app/views/investigations/corrective_actions/show.html.erb
@@ -1,0 +1,75 @@
+<% page_heading = @corrective_action.summary %>
+<% page_title page_heading %>
+
+<% content_for :after_header do %>
+  <%= govukBackLink(
+    text: "Back",
+    href: investigation_path(@investigation)
+  ) %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+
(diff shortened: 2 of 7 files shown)


psd-web/app/views/investigations/activities/_activity_card.html.erb
```diff
@@ -52,6 +52,9 @@
 
     <% if activity.is_a? AuditActivity::Test::Result %>
       <%= link_with_hidden_text_to "View test result", " from #{activity.test_result.date.to_s(:govuk)}", investigation_test_result_path(@investigation, activity.test_result.id) %>
+
+    <% elsif activity.is_a?(AuditActivity::CorrectiveAction::Add) && activity.corrective_action %>
+      <%=  link_to "View corrective action", investigation_action_path(@investigation, activity.corrective_action), class: "govuk-link" %>
```
frankieroberto, line 57: Not ideal, but we should refactor this to use the new `metadata` json object and separate partials per activity entry in a separate PR.

psd-web/config/routes.rb
```diff
@@ -140,6 +140,8 @@
 
     resources :test_results, controller: "investigations/test_results", only: :show, path: "test-results"
 
+    resources :actions, controller: "investigations/view_corrective_actions", only: :show, path: "corrective-actions"
```
frankieroberto, line 143: Resource uses a shorter name (`:actions` instead of `:corrective_actions`) to avoid conflicting with the existing resource.

psd-web/app/views/investigations/corrective_actions/show.html.erb
```diff
@@ -0,0 +1,75 @@
...
+        key: { text: "Duration of measure" },
+        value: { text: @corrective_action.duration.upcase_first }
+      },
+      {
+        key: { text: "Scope" },
+        value: { text: @corrective_action.geographic_scope }
+      }
+    ]
+
+    if @corrective_action.details.present?
```
StupidCodeFactory, line 50: might be wrong but I can't see a test for this?

frankieroberto: It's covered in the feature spec: https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/684/files#diff-72f1fd770cc58e1731a1a09a80797d35R68

psd-web/app/views/investigations/corrective_actions/show.html.erb
```diff
@@ -0,0 +1,75 @@
...
+
+    <div class="app-meta-area">
+      <p class="govuk-body govuk-hint">
+        Added <%= @corrective_action.created_at.to_s(:govuk) %>
+      </p>
+    </div>
+
+    <% rows = [
+      {
+        key: { text: "Date" },
```
edwardhorsford, line 25:
```suggestion
        key: { text: "Date of action" },
```

psd-web/app/views/investigations/corrective_actions/show.html.erb
```diff
@@ -0,0 +1,75 @@
...
+      {
+        key: { text: "Type of action" },
+        value: { text: @corrective_action.measure_type.upcase_first }
+      },
+      {
+        key: { text: "Duration of measure" },
+        value: { text: @corrective_action.duration.upcase_first }
+      },
+      {
+        key: { text: "Scope" },
```
edwardhorsford, line 45:
```suggestion
        key: { text: "Geographic scope" },
```

frankieroberto: Heh, it was "Scope" in the prototype screenshot. I actually think it's better than "Geographic scope", which feels a bit redundant once you can see the value, and a slightly awkward word?

psd-web/app/views/investigations/corrective_actions/show.html.erb
StupidCodeFactory: Sorry I meant when the details are not present, I saw this was added to the feature test and wondered where the other branch was covered.

psd-web/app/views/investigations/corrective_actions/show.html.erb
edwardhorsford: Ok, persuaded.

psd-web/app/views/investigations/corrective_actions/show.html.erb
frankieroberto: I didn't feel the need to test the row not appearing if there was no data, in this instance. Mainly to avoid the feature specs getting slow.

If this gets any more complicated, we can consider refactoring into a helper and testing that, but for now I think it's simple enough not to need it.

PR #662, opened 2020-05-28
Split attachment tab

StupidCodeFactory:
Split the attachment for images and supporting informations

## Checklist:
- [ ] Have you documented your changes in the pull request description?
- [ ] Does the change present any security considerations?
- [ ] Is any gem functionality overloaded? Eg: Devise controller methods being overloaded.
- [ ] Has acceptance criteria been tested by a peer?
- [ ] Has the CHANGELOG been updated? (If change is worth telling users about.)

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

### Accessibility testing (author)
- [ ] Works keyboard only
- [ ] Tested with one screen reader
- [ ] Zoom page to 400% - content still visible
- [ ] Disable CSS - does content make sense and appear in a logical order?

PR #685, opened 2020-06-04
Feature Branch: Attachments changes

scruti:
Feature branch for bringing up changes in how we display the attachments in our service:

Included work:
- [Remove duplication of activities attachments into investigation attachments ](https://trello.com/c/cwtcqW1x/583-remove-duplication-of-activities-attachments-into-investigation-attachments): https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/664 
- [Split attachments tab into Images and Supporting Information](https://trello.com/c/Tv9Wx6eT/462-3-split-attachments-tab-in-to-images-and-supporting-information-tabs): https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/662
- [Don't reveal generic attachments to teams not involved in a case](https://trello.com/c/u4CDPR33/553-3-dont-reveal-generic-attachments-to-teams-not-involved-in-a-case): https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/689
- [Table view for supporting information](https://trello.com/c/hySkYqQe/494-5-table-view-for-supporting-information): https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/656
- [Link to add activity flow from add supporting information flow](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/711)
- [Imrpove display of thumbnail graphics and generic attachments](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/685/commits/90124ff3bd00482b83b624a3c1f7adfa9887a9ea)


psd-web/app/models/investigation.rb
```diff
@@ -77,12 +77,32 @@ def initialize(*args)
     super
   end
 
+  def images
```
slorek, line 80: Please change: Remove `AttachmentConcern` from this model, as that only provides an `images` method which is overridden here.

slorek: We should also consider splitting the `documents` relation out to create an `images` relation if we don't want them all together. This query could be quite slow.

StupidCodeFactory: I forgot about it, I wanted to look at the query plan and add the correct index. 
I did think it would be ideal to have image on one side and documents on the other side, but this would require quite a big change as the way documents are added is shared everywhere across the app, you'd need to add some validation and probably some UI changes.
I think it might be worth doing but adding and index on `content_type`, `record_type` and `record_id` should be sufficient to significantly speed up the query

StupidCodeFactory: @slorek I think this query would be fine:
here is the query plan:

```
   Join Filter: (active_storage_attachments.blob_id = active_storage_blobs.id)
   ->  Index Scan using index_active_storage_attachments_uniqueness on active_storage_attachments  (cost=32.69..40.73 rows=1 width=92)
         Index Cond: (((record_type)::text = 'Investigation'::text) AND (record_id = '9'::bigint) AND ((name)::text = 'documents'::text))
   ->  Seq Scan on active_storage_blobs  (cost=0.00..2.40 rows=1 width=294)
         Filter: ("left"((content_type)::text, 5) = 'image'::text)
```
It first uses the index to filter down the record and then it would do a scan on the remaining sequence. so searching for left(content_type), 5) should be fairly insignificant as I don't expect investigation to have that much documents.

slorek: ok cool thanks!

PR #687, opened 2020-06-04
Adding views for correspondence

frankieroberto:
This adds three new views for viewing the details of "correspondence" added to the case. These are linked to from the Activity timeline, but in future will be linked to from the Supporting Information tab.

Unlike test results and corrective actions, these views are only available to users whose teams have been added to the case, as the correspondence may contain personal contact details.

[Trello card](https://trello.com/c/GVOQQI1M/497-3-view-correspondence-on-a-page)

### Meeting

Screenshot 2020-06-09 at 11 39 20

### Phone call

Screenshot 2020-06-09 at 11 40 29

### Email

Screenshot 2020-06-09 at 11 39 33

## Implementation notes

We have existing controllers for these models (eg `Investigations::EmailsController`) but these all use the `show` action for the creation flow, using the WickedWizard gem. To avoid conflicting with that, I've added new controllers (eg `Investigations::ViewEmailsController`). These are routed to from the same path, but with a constraint on the ID being numeric, so that:

`/caes/case-id/emails/ID` -> gets routed to the new "view" controller.
`/cases/case-id/emails/details` -> gets routed to the existing controller.

This is a bit unconventional, but avoids a bigger refactoring job. Hopefully at some point we can merge these back together by rethinking how the creation flow works.


CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## 2020-06-05
+– Added the ability to view the details of correspondence added to the case.
```
edwardhorsford, line 5:
```suggestion
– Added the ability to view the details of correspondence added to the case on their own page.
```

psd-web/app/views/investigations/emails/show.html.erb
```diff
@@ -0,0 +1,82 @@
...
+
+    <div class="app-meta-area">
+      <p class="govuk-body govuk-hint">
+        Added <%= @email.created_at.to_s(:govuk) %>
+      </p>
+    </div>
+
+    <% rows = [
+      {
+        key: { text: "Date" },
```
edwardhorsford, line 25:
```suggestion
        key: { text: "Date of email" },
```

psd-web/app/views/investigations/meetings/show.html.erb
```diff
@@ -0,0 +1,69 @@
...
+
+    <div class="app-meta-area">
+      <p class="govuk-body govuk-hint">
+        Added <%= @meeting.created_at.to_s(:govuk) %>
+      </p>
+    </div>
+
+    <% rows = [
+        {
+          key: { text: "Date" },
```
edwardhorsford, line 25:
```suggestion
          key: { text: "Date of meeting" },
```

psd-web/app/views/investigations/phone_calls/show.html.erb
```diff
@@ -0,0 +1,54 @@
...
+
+    <div class="app-meta-area">
+      <p class="govuk-body govuk-hint">
+        Added <%= @phone_call.created_at.to_s(:govuk) %>
+      </p>
+    </div>
+
+    <% rows = [
+        {
+          key: { text: "Date" },
```
edwardhorsford, line 25:
```suggestion
          key: { text: "Date of call" },
```

edwardhorsford:
A couple further changes needed:

Can we change the link text for the `view email` / `view transcript` to be the filename with a filesize? eg `foobar.msg (456kb)`

We noted we've got a bug that the following are not validated and so might be missing:
* to / from
* name
* email
* email summary

We'll raise a separate bug to validate them - but this template should sensibly handle when they're missing.

frankieroberto:
@slorek good idea! I might do that as a quick follow-up PR, so as to keep this one easier to follow.

tobyhughes:
Similar to Ed's comment above, 'Record phone call' may be missing:

- name
- phone number
- summary

view page needs to be able to handle these scenarios

@edwardhorsford suggest similar to emails for h1, i.e. 'Phone call on {date}'

Suggest removing 'Call with' row from table if a name is not provided, however, if the user includes a phone number but no name, we may want to keep the 'Call with' row... 
 
image

tobyhughes:
@frankieroberto sorry to be a huge pain, the same problem is repeated in 'Record a meeting' flow

Users don't have to provide:

- name
- summary (if they provide a transcript)

Which is affecting the h1 and the view table. Suggest we go with 'Meeting on {date}' as per above

image

PR #689, opened 2020-06-05
Restrict visibility of generic attachments

scruti:
[Jira ticket](https://trello.com/c/u4CDPR33/553-3-dont-reveal-generic-attachments-to-teams-not-involved-in-a-case)

Only users belonging to a team that has permissions on the investigation will be able to view the details of the investigation generic attachments when visiting the investigation supporting information tab.

Other users will be aware of the attachment existence but won't be able to view the preview, title or description from them.

### An user not belonging to any of the teams with access to the investigation will see:
Screenshot from 2020-06-05 13-30-06

### While an user with access will see 
Screenshot from 2020-06-05 13-25-48

(don't mind the preview format showing different extensions, they're different docs):

### DIFF
--- a/psd-web/app/controllers/investigations/supporting_information_controller.rb
+++ b/psd-web/app/controllers/investigations/supporting_information_controller.rb
@@ -9,6 +9,7 @@ module Investigations
       @investigation                = investigation.decorate
       @supporting_information       = investigation.supporting_information_attachments.decorate
       @other_supporting_information = investigation.generic_supporting_information_attachments.decorate
+      @generic_attachments_visible  = investigation.teams_with_access.include?(current_user.team)
 
       @breadcrumbs = {
         items: [
--- a/psd-web/app/controllers/investigations/supporting_information_controller.rb
+++ b/psd-web/app/controllers/investigations/supporting_information_controller.rb
@@ -9,7 +9,6 @@ module Investigations
       @investigation                = investigation.decorate
       @supporting_information       = investigation.supporting_information_attachments.decorate
       @other_supporting_information = investigation.generic_supporting_information_attachments.decorate
-      @generic_attachments_visible  = investigation.teams_with_access.include?(current_user.team)
 
       @breadcrumbs = {
         items: [
--- a/psd-web/app/views/documents/_generic_document_card.html.erb
+++ b/psd-web/app/views/documents/_generic_document_card.html.erb
@@ -1,0 +1,34 @@
+<%
+  document = document.decorate
+  title = document.title
+  safe = document.metadata["safe"]
+%>
+
+<div class="govuk-grid-row govuk-!-padding-bottom-6">
+  <div class="govuk-grid-column-one-quarter">
+    <%= render "documents/document_preview", document: document, dimensions: "300x200", restricted: false %>
+  </div>
+
+  <div class="govuk-grid-column-three-quarters">
+    <h2 class="govuk-heading-m govuk-!-margin-bottom-1"><%= title %></h2>
+    <span class="govuk-hint govuk-!-font-size-16">
+      <%= formatted_file_updated_date(document) %>
+    </span>
+
+    <p><%= t("supporting_information.restricted_generic.placeholder") %></p>
+    <p class="govuk-body"><%= document.description %></p>
+
+    <% if safe %>
+      <%= link_with_hidden_text_to "View #{pretty_type_description(document)}",
+          "(#{title}, opens in a new window)",
+          document,
+          class: "govuk-link app-block-link govuk-!-margin-right-3",
+          target: '_blank',
+          rel: 'noopener' %>
+
+      <% if policy(parent).update?(user: current_user) %>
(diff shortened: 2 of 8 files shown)


psd-web/app/controllers/investigations/supporting_information_controller.rb
```diff
@@ -9,6 +9,7 @@ def index
       @investigation                = investigation.decorate
       @supporting_information       = investigation.supporting_information_attachments.decorate
       @other_supporting_information = investigation.generic_supporting_information_attachments.decorate
+      @generic_attachments_visible  = investigation.teams_with_access.include?(current_user.team)
```
slorek, line 12: Please change: move this logic to a Pundit policy.

psd-web/app/views/investigations/tabs/_supporting_information.erb
```diff
@@ -20,8 +20,14 @@
...
-<% @other_supporting_information.each do |attachement| %>
-  <%= render "documents/document_card", document: attachement, parent: @investigation %>
+<% if @generic_attachments_visible %>
+  <% @other_supporting_information.each do |attachement| %>
+    <%= render "documents/generic_document_card", document: attachement, parent: @investigation %>
+  <% end %>
+<% else %>
+  <% @other_supporting_information.each do |attachement| %>
+    <%= render "documents/restricted_generic_document_card", document: attachement %>
+  <% end %>
```
slorek, line 30: Typo fix:
```suggestion
<% if @generic_attachments_visible %>
  <% @other_supporting_information.each do |attachment| %>
    <%= render "documents/generic_document_card", document: attachment, parent: @investigation %>
  <% end %>
<% else %>
  <% @other_supporting_information.each do |attachment| %>
    <%= render "documents/restricted_generic_document_card", document: attachment %>
  <% end %>
```

scruti: Wops, thanks.

psd-web/app/controllers/investigations/supporting_information_controller.rb
scruti: Thanks for highlighting this @slorek . There was a policy already wrapping this logic, so using it.

scruti:
@slorek fixed! And was even worse as we already had a policy for this.

psd-web/app/views/documents/_generic_document_card.html.erb
```diff
@@ -0,0 +1,34 @@
+<%
+  document = document.decorate
```
frankieroberto, line 2: Can we call `.decorate` within the controller?

psd-web/app/views/investigations/tabs/_supporting_information.erb
```diff
@@ -20,8 +20,14 @@
...
-  <%= render "documents/document_card", document: attachement, parent: @investigation %>
+<% if policy(@investigation).view_protected_details?(user: current_user) %>
+  <% @other_supporting_information.each do |attachment| %>
+    <%= render "documents/generic_document_card", document: attachment, parent: @investigation %>
+  <% end %>
+<% else %>
+  <% @other_supporting_information.each do |attachment| %>
+    <%= render "documents/restricted_generic_document_card", document: attachment %>
+  <% end %>
 <% end %>
```
frankieroberto, line 31: I wonder if this is slightly easier to follow by reversing the nesting of the loop and the condition? (Also slightly fewer lines)

```suggestion
<% @other_supporting_information.each do |attachment| %>
  <% if policy(@investigation).view_protected_details?(user: current_user) %>
    <%= render "documents/generic_document_card", document: attachment, parent: @investigation %>
  <% else %>
    <%= render "documents/restricted_generic_document_card", document: attachment %>
  <% end %>
<% end %>
```

psd-web/app/views/documents/_restricted_generic_document_card.html.erb
```diff
@@ -0,0 +1,16 @@
+<% document = document.decorate %>
```
frankieroberto, line 1: Can the controller pass a pre-decorated document to the views?

psd-web/app/views/investigations/tabs/_supporting_information.erb
scruti: I considered it but ended discarding it because it would check the policy for each individual document on the loop, while the policy applies over all the documents in the investigation and should only need to be checked once.

Ideally we would check the condition once, and store the partial to render in a variable, then the loop would just use that variable.
Problem is that their interfaces are different as one needs parent while the other doesn't. 

Will re-think this!

nasirkhanbeis:
looks good however seems we are not adding restriction for jpg files(in images tab) which are attached as generic attachment. Also will be good to add a test for same.

nasirkhanbeis:
> looks good however seems we are not adding restriction for jpg files(in images tab) which are attached as generic attachment. Also will be good to add a test for same.

ignore this comment as its allowed in story criteria :) , looks all good to me!

PR #690, opened 2020-06-05
Bugfix: only reveal attachment description when the right input is selected

frankieroberto:
This fixes a bug whereby currently the "Attachment description" textarea is revealed when you select either file input on the page, rather than just the attachment file input.


slorek:
Thanks! Could you include a screenshot or URL to the problem so i can see what it does?

frankieroberto:
@slorek if you visit https://psd-pr-691.london.cloudapps.digital/cases/2006-0049/activity/new and add a "meeting", fill out the form and press Continue, you’ll get to the second page. Currently if you select a file under the "Details" heading, then the "Attachment description" textarea is revealed, which is unintentional.

This fixed that so that it is only revealed when a file is selected under the "Related attachments" heading.

(Spotted by @edwardhorsford )

PR #691, opened 2020-06-05
Bugfix: prevent fieldsets being nested

frankieroberto:
This fixes an HTML issue, where the fieldsets were inadvertently (I think) nested. Nested fieldsets can be tricky to navigate for screen reader users.


slorek:
Thanks! Can you include a screenshot or URL so I can see the change?

frankieroberto:
@slorek if you visit https://psd-pr-691.london.cloudapps.digital/cases/2006-0049/activity/new and add a "meeting", fill out the form and press Continue, you’ll get to the second page. Currently the fieldset with the "Related attachments" legend is nested within the fieldset with the "Details" legend. This PR separates them so they are no longer nested. 

The visual appearance is the same.

(Spotted as part of work on #687)

PR #693, opened 2020-06-05
Reject mobile phone numbers with too many digits

scruti:
Numbers with too many digits are being rejected by UK Notify API based in this validation:

https://github.com/alphagov/notifications-utils/blob/652d9bab827238e9fd247bd2d4e273336419d9cd/notifications_utils/recipients.py#L425

Right now these errors are silently happening in the background ([Error tracker issue](https://sentry.io/organizations/beis/issues/1661381821/?project=1329381)) so the user is not aware of them.

This PR will bring a form validation error to the user so the mobile number is rejected in our service form submission instead of being rejected by Notify API in the background.


scruti:
[Bugfix for mobile number validation error](https://trello.com/c/9QkIt0qV/590-bugfix-for-mobile-number-validation-error)

PR #696, opened 2020-06-08
Only show complainant details on case added activity if user is collaborating on the case

slorek:
https://trello.com/c/3HP9NsVf/554-hide-enquiry-source-contact-details-in-activity-log

## Description
This tweaks the logic which determines whether a user can view complainant details on the activity log entry for case created. It now uses the new policy which specifies that the user must be on a team added as a collaborator to the case. Previously, the user had to be in the same team or organisation as the person who created the case.

I've also removed the now-redundant `Complainant#can_be_displayed?` method which was only used elsewhere in the search results text highlighting feature. We are currently discussing how best to improve this feature and the indexing of complainants so I have not changed this.

Note: The copy has not been changed from the existing generic text. I think the copy is probably fine as-is, but we might want something more tailored or consistent with other places where contact details are hidden.

### Screenshots

Example case: https://psd-pr-696.london.cloudapps.digital/cases/2006-0018 (created by OPSS user).

#### User on a team on the case

Screenshot 2020-06-08 at 13 54 06

#### Searching for complainant email address (team on case)

Screenshot 2020-06-08 at 14 00 34

#### User on a team not on the case

Screenshot 2020-06-08 at 13 55 08

#### Searching for complainant email address (not on case)

Screenshot 2020-06-08 at 13 59 16

## Checklist:

### DIFF
--- a/psd-web/app/models/complainant.rb
+++ b/psd-web/app/models/complainant.rb
@@ -19,16 +19,4 @@ class Complainant < ApplicationRecord
 
   validates :name, length: { maximum: 100 }
   validates :other_details, length: { maximum: 10_000 }
-
-  def can_be_displayed?(user)
-    can_be_seen_by_user?(user) || investigation.child_should_be_displayed?(user)
-  end
-
-private
-
-  def can_be_seen_by_user?(user)
-    return true if investigation.creator_user.has_gdpr_access?(user)
-
-    complainant_type != "Consumer"
-  end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## 2020-06-12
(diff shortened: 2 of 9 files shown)


psd-web/app/models/complainant.rb
```diff
@@ -19,16 +19,4 @@ class Complainant < ApplicationRecord
...
-    can_be_seen_by_user?(user) || investigation.child_should_be_displayed?(user)
-  end
-
-private
-
-  def can_be_seen_by_user?(user)
-    return true if investigation.source&.user_has_gdpr_access?(user)
-
-    complainant_type != "Consumer"
-  end
```
scruti, line 33: :clap:

nasirkhanbeis:
thanks @slorek  search result display text looking good as expected

slorek:
Updated copy:

Screenshot 2020-06-12 at 13 14 57

PR #698, opened 2020-06-09
Make migration fail proof

slawosz:
Quickfix.

### DIFF
--- a/psd-web/db/migrate/20200604192242_create_creator_users_from_user_source.rb
+++ b/psd-web/db/migrate/20200604192242_create_creator_users_from_user_source.rb
@@ -5,6 +5,7 @@ class CreateCreatorUsersFromUserSource < ActiveRecord::Migration[5.2]
         UserSource.where(sourceable_type: "Investigation").each do |us|
           user = us.user
           next unless user
+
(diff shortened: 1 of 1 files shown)


psd-web/db/migrate/20200604192242_create_creator_users_from_user_source.rb
```diff
@@ -3,9 +3,14 @@ def change
     safety_assured do
       UserSource.transaction do
         UserSource.where(sourceable_type: "Investigation").each do |us|
+          user = us.user
+          next unless user
+
           Collaboration::CreatorUser.create!(investigation_id: us.sourceable_id, collaborator_id: us.user_id, collaborator_type: User)
           team = us.user.team
-          Collaboration::CreatorTeam.create!(investigation_id: us.sourceable_id, collaborator_id: team.id, collaborator_type: Team)
+          if team
```
frankieroberto, line 11: We shouldn’t have any users without a team now, as this column is non-nullable?

slorek: It's possible the team may have been deleted as there's no foreign key constraint, but I don't think this is the case.

frankieroberto: Doh, I assumed there was a foreign key constraint...

PR #706, opened 2020-06-11
Add ability to edit test results

frankieroberto:
## Description

This adds the ability, for users who belong to team involved with a case, to update the details of any test results associated with that case.

The edit form is identical to one used when adding a test result, with the exception of a [Details component](https://design-system.service.gov.uk/components/details/) which allows a replacement file to be uploaded.

## Implementation notes

* The test result model and existing flow uses the DateConcern, which is a bit weird (it adds a bunch of extra getters and setters). Rather than refactor this, I've left it as is for now, even though it requires a few workarounds in the update flow.
* The file attachment description is stored in the `metadata` json on the activestorage blob, which seems a bit unconventional (as far as I can tell that JSON object is intended for use by 'analysers'), which makes updating it a bit awkward. I've left as-is for now though.
* The test result model claims to support multiple file attachments (`has_many_attached :documents`) but the interface assumes that there's only one, and that you're replacing it rather than adding/removing them. I've left this as-is for now, but we should resolve this one way or another in a future PR.

## Checklist:

#### Test result page showing 'Edit' link (if you have permission)

Screenshot 2020-06-19 at 15 37 55

#### Edit test result page

Screenshot 2020-06-19 at 15 38 07

Screenshot 2020-06-19 at 15 42 03

#### Activity timeline showing summary of edit

Screenshot 2020-06-19 at 15 38 40


nasirkhanbeis:
@frankieroberto  I found few  minor issues to address
1. On edit test results, I think user will expect to see confirmation banner on successful update
2. Also I think user should be returned to supporting inform tab or overview after I edit test results
3. Doesn't records new activity when I replace file with excel sheet, should log new activity as like any other field change
4.When I replace file with .txt it over writes existing old activity with new file instead creating a new activity
@edwardhorsford  what do you think on point 1?

nasirkhanbeis:
@frankieroberto  I see we had a question already whether to store who edited document or not, if you think if its not a quick fix then we can always fix it later and I will add a card as followup for this.

KamQureshi:
All good @frankieroberto

frankieroberto:
@nasirkhanbeis I've fixed the 3rd issue you reported – an activity entry and email is now generated if you change _just_ the file attached. Also the activity entry now lists the name of the new file attached, if you uploaded a new file.

I think we've decided not to show a confirmation message for now, and to redirect to the view page for the updated test result instead.

psd-web/app/models/audit_activity/test/test_result_updated.rb
```diff
@@ -0,0 +1,78 @@
...
+    "Edited"
+  end
+
+  def test_result
+    ::Test::Result.find(metadata["test_result_id"])
+  end
+
+  def new_date_of_test
+    if updates["date"]
+      Date.parse(updates["date"][1])
```
slawosz, line 34: Here and in other places: is it save to use `magic numbers`?

psd-web/app/services/update_test_result.rb
```diff
@@ -0,0 +1,92 @@
...
+    # Currently required by DateConcern
+    test_result.set_dates_from_params(new_attributes)
+
+    test_result.attributes = new_attributes.except(:date)
+
+    @previous_attachment = test_result.documents.first
+
+    ActiveRecord::Base.transaction do
+      if new_file
+        # remove previous attachment
```
slawosz, line 29: you can extract this lines to `remove_previous_attachment` method which will remove comment.

psd-web/spec/services/update_test_result_spec.rb
```diff
@@ -0,0 +1,192 @@
...
+    context "with a missing parameters" do
+      let(:test_result) { build(:test_result) }
+      let(:result) { described_class.call(test_result: test_result) }
+
+      it "returns a failure" do
+        expect(result).to be_failure
+      end
+    end
+
+    context "with required parameters" do
```
slawosz, line 22: I am not fan of very long context. I think that happy path one does not need to be encapsulated in context block.

psd-web/spec/services/update_test_result_spec.rb
```diff
@@ -0,0 +1,192 @@
+require "rails_helper"
+
+RSpec.describe UpdateTestResult, :with_stubbed_mailer, :with_stubbed_elasticsearch, :with_stubbed_antivirus, :with_test_queue_adapter do
+  describe ".call" do
```
slawosz, line 4: There is only one method, I don't think it is necessary to use `describe` block in such case

psd-web/app/models/audit_activity/test/test_result_updated.rb
frankieroberto: This comes from having the hash format of:

```ruby
{
  "date": ["2010-01-01", "2012-02-02"]
}
```

which is what the [`previous_changes`](https://api.rubyonrails.org/classes/ActiveModel/Dirty.html#method-i-previous_changes) Rails method gives you. We _could_ change this to a more verbose format, eg something like

```ruby
[
  {
    "attribute": "date",
    "old_value": "2010-01-01",
    "new_value": "2012-02-02",
]
```

which might make the code a bit more readable (but would require transforming the results of `previous_changes`)?

psd-web/spec/services/update_test_result_spec.rb
frankieroberto: Good point – I think I agree. Although the existing specs all do this too, so maybe we could remove all these as a separate PR if there's a consensus.

psd-web/app/services/update_test_result.rb
slawosz:
```
def try_to_remove_previous_attachment
  return unless new_file
  test_result.documents.detach
  test_result.documents.attach(new_file)
```

psd-web/app/controllers/investigations/test_results_controller.rb
```diff
@@ -4,4 +4,39 @@ def show
     authorize @investigation, :view_non_protected_details?
     @test_result = @investigation.test_results.find(params[:id]).decorate
   end
+
+  def edit
+    @investigation = Investigation.find_by(pretty_id: params[:investigation_pretty_id]).decorate
```
slorek, line 9:
```suggestion
    @investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id]).decorate
```

Raises exception if not found.

psd-web/app/controllers/investigations/test_results_controller.rb
```diff
@@ -4,4 +4,39 @@ def show
...
+    authorize @investigation, :update?
+    @test_result = @investigation.test_results.find(params[:id])
+
+    @file_blob = @test_result.documents.first.blob
+  end
+
+  def update
+    @investigation = Investigation.find_by(pretty_id: params[:investigation_pretty_id]).decorate
```
slorek, line 17:
```suggestion
    @investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id]).decorate
```

psd-web/spec/features/edit_test_result_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      end
+
+      within_fieldset "What was the result?" do
+        choose "Fail"
+      end
+
+      fill_in "Further details", with: "Final result"
+
+      # TODO: lookup details using the summary text
+      find("details").click
```
slorek, line 64: Is this to be done before merging?

frankieroberto:
@slorek good spot! I broke the validations with a refactor... 🤦 . Here's the fix (adds a test for it too): 2c37b3626d5cb62224c765ebb523280f24ffc382

psd-web/spec/services/update_test_result_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+
+      it "returns a failure" do
+        expect(result).to be_failure
+      end
+    end
+
+    context "with required parameters that trigger a validation error" do
+      let(:test_result) { build(:test_result) }
+      let(:user) { create(:user, :activated) }
+      let(:new_attributes) do
```
slorek, line 25: Suggestion: I think we should assume it's any kind of `Hash` as services should be able to accept params from any source.

psd-web/spec/services/update_test_result_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+        let(:test_result) do
+          create(:test_result,
+                 investigation: investigation,
+                 legislation: legislation,
+                 details: details,
+                 product: product,
+                 result: :failed)
+        end
+        let(:new_attributes) do
+          ActionController::Parameters.new({
```
slorek, line 70: Suggestion: I think we should assume it's any kind of Hash as services should be able to accept params from any source.

frankieroberto: Yeah, they _should_. Unfortunately the DateConcern, which the service is using, is expecting a params object, in order to be able to call `require` and `permit` here: https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/models/concerns/date_concern.rb#L35

I’d like to refactor that concern, but I think we should do that as a separate task, as it's used all over the place.

psd-web/spec/services/update_test_result_spec.rb
frankieroberto: See https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/706#discussion_r445645263

psd-web/spec/features/edit_test_result_spec.rb
frankieroberto: Yep, opps. Fixed here: 703557227c27c00757f5f241ee74308f70a3c1db 👍

psd-web/spec/services/update_test_result_spec.rb
slorek: I see

psd-web/app/models/audit_activity/test/test_result_updated.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+  end
+
+  def new_result
+    if updates["result"]
+      updates["result"][1]
+    end
```
scruti, line 47: Nitpick:
```suggestion
    updates["result"]&.at(1)
```

psd-web/app/models/audit_activity/test/test_result_updated.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def new_details
+    if updates["details"]
+      updates["details"][1]
+    end
```
scruti, line 53: Nitpick:
```suggestion
    updates["details"]&.at(1)
```

psd-web/app/models/audit_activity/test/test_result_updated.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def new_legislation
+    if updates["legislation"]
+      updates["legislation"][1]
+    end
```
scruti, line 59: Nitpick:
```suggestion
    updates["legislation"]&.at(1)
```

psd-web/app/models/audit_activity/test/test_result_updated.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def new_filename
+    if updates["filename"]
+      updates["filename"][1]
+    end
```
scruti, line 65: Nitpick:
```suggestion
    updates["filename"]&.at(1)
```

psd-web/app/models/audit_activity/test/test_result_updated.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def new_file_description
+    if updates["file_description"]
+      updates["file_description"][1]
+    end
```
scruti, line 71: Nitpick:
```suggestion
    updates["file_description"]&.at(1)
```

PR #710, opened 2020-06-15
Move owner to collaborations

slawosz:
Most of the changes are in class `Investigation`. Please take a look into `Owner`, `OwnerUser`, `OwnerTeam` models.


psd-web/app/models/concerns/searchable.rb
```diff
@@ -61,6 +61,7 @@ module Searchable
     end
 
     def self.full_search(query)
+      # binding.pry
```
slorek, line 64: ✂️

psd-web/app/models/investigation.rb
```diff
@@ -71,14 +65,39 @@ class Investigation < ApplicationRecord
...
+
   def owner_team
-    owner&.team
+    super || owner_user&.team
+  end
+
+  def owner=(team_or_user)
+    if team_or_user.is_a? User
+      self.owner_user = team_or_user
+      self.owner_team = nil
```
slorek, line 95: This should be `team_or_user.team`

psd-web/spec/features/change_ownership_for_investigation_spec.rb
```diff
@@ -45,13 +45,15 @@
   end
 
   scenario "change owner to your team" do
+    # binding.pry
```
slorek, line 48: ✂️

psd-web/spec/features/change_ownership_for_investigation_spec.rb
```diff
@@ -45,13 +45,15 @@
...
     choose user.team.name
     click_button "Continue"
 
     fill_and_submit_change_owner_reason_form
 
     expect_confirmation_banner("Allegation owner changed to " + user.team.name)
     expect_page_to_show_case_owner(user.team)
+    # binding.pry
```
slorek, line 56: ✂️

slorek:
`Investigation#owner` should always be a `Team` with the user an optional assignee property. Let's discuss whether we want to merge this as-is as an intermediary step, or implement the change as part of this story.

slorek:
I think we also need some constraint to ensure `owner_team` is always present.

psd-web/app/models/investigation.rb
```diff
@@ -20,23 +20,24 @@ class Investigation < ApplicationRecord
...
   validates :description, length: { maximum: 10_000 }
   validates :non_compliant_reason, length: { maximum: 10_000 }
   validates :hazard_description, length: { maximum: 10_000 }
 
+  # validate :owner_team_presence, on: :create
+  # def owner_team_presence
+  #   unless owner_team
+  #     errors[:owner_team] << 'Please add owner'
+  #   end
+  # end
```
slorek, line 33: Should this be uncommented?

psd-web/app/models/investigation.rb
```diff
@@ -70,12 +71,37 @@ class Investigation < ApplicationRecord
...
+  def owner_id
+    owner&.id
+  end
+
+  def owner=(team_or_user)
+    if team_or_user.is_a? User
+      self.owner_user = team_or_user
+      self.owner_team = team_or_user.team
+    end
+    if team_or_user.is_a? Team
```
slorek, line 99: `elsif`

psd-web/db/migrate/20200615121613_move_owners_to_collaboration.rb
```diff
@@ -0,0 +1,13 @@
+class MoveOwnersToCollaboration < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured do
+      Investigation.transaction do
+        Investigation.all.each do |i|
+          owner = i[:owner_type].constantize.find(i[:owner_id])
+          i.owner = owner
+          i.save
```
slorek, line 8: I think we should use `save!` here so the migration rolls back the transaction in case of error.

psd-web/db/migrate/20200615121613_move_owners_to_collaboration.rb
```diff
@@ -0,0 +1,13 @@
+class MoveOwnersToCollaboration < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured do
+      Investigation.transaction do
```
slorek, line 4: Redundant as migrations happen in a transaction already.

psd-web/app/models/investigation.rb
```diff
@@ -20,23 +20,24 @@ class Investigation < ApplicationRecord
...
-  # Can currently only validate owner_id on update since create case wizards
-  # validate partially-built Investigation instances
-  validates :owner_id, presence: true, on: :update
-
   validates :user_title, length: { maximum: 100 }
   validates :description, length: { maximum: 10_000 }
   validates :non_compliant_reason, length: { maximum: 10_000 }
   validates :hazard_description, length: { maximum: 10_000 }
 
+  # validate :owner_team_presence, on: :create
```
slorek, line 28: Validate on `update` as well.

psd-web/spec/features/create_project_spec.rb
```diff
@@ -26,6 +26,7 @@
       expect(page).to have_error_summary "Select whether or not the case is related to the coronavirus outbreak"
 
       choose "Yes, it is (or could be)"
+      puts "Already"
```
slorek, line 29: ✂️

psd-web/spec/models/investigation_spec.rb
```diff
@@ -78,4 +78,30 @@
       end
     end
   end
+
+  describe "ownership" do
+    let(:user)          { create(:user, :activated, has_viewed_introduction: true) }
+    let(:investigation) { create(:investigation, creator: user) }
```
slorek, line 84: The owner defaults to the creator by the factory if not set. Since we're testing the owner behaviour here we should set it explicitly:

```suggestion
    let(:investigation) { create(:investigation, owner: user, creator: user) }
    let(:owner) { user }
```
We can then override `let(:owner)` in the following `context` blocks.

slawosz: I am testing setter here, so I would prefer to do it explicitly.

psd-web/app/models/investigation.rb
slawosz: Actually, `required: true` by association, does the job.

psd-web/app/models/investigation.rb
slawosz: Actually, `required: true` by association, does the job.

PR #713, opened 2020-06-16
Move case owner to collaborations table

StupidCodeFactory:
## CHANGES:

Originally this PR was meant to do this [trello](https://trello.com/c/p08HDIqY/434-5-add-view-only-case-permissions-for-teams-backend-support) but will merging some of the "move owner to collaborations" story it turned into finishing the "move owner collaborations" as I wanted to change the terminology and topology of the collaboration classes.
The main rational was to make `Owner` a subclass of `Edit` as they have the same access.

I've created a read only policy (which I'll use in a later story) along side it's investigation policy which I will leave in this PR. It's a very small which I started doing initially. They are not yet used. 

## important consideration for reviewers
I would to improve the names of the collaborations and teams ( has_many :through) as I think they are far from amazing. I you can give a few suggestion or whether you're happy with the current ones.
In that spirit I've renamed `team_owner` and `user_owner` to just `team` and `user` the owner bit seem a bit redundant and the `Owner` semantic is already carried by the investigation/user or investigation/team joining relationship. 
I'm happy to change it back to what it was if most prefer the previous names. 

## new `Team` and `User` concerns 
I've extracted in `psd-web/app/models/concerns/team_collaborator_interface.rb` and `psd-web/app/models/concerns/user_collaborator_interface.rb` all the common interface needed by `Team` and `User` instances to so that collaborations don't have to know about the internals of a collaborator and removing the need for endless type checking `if/else`.


StupidCodeFactory:
[(5) Add view-only case permissions for teams (backend support)](https://trello.com/c/p08HDIqY/434-5-add-view-only-case-permissions-for-teams-backend-support)

psd-web/app/controllers/collaborators_controller.rb
```diff
@@ -4,7 +4,11 @@ class CollaboratorsController < ApplicationController
   end
 
   def index
-    @teams = @investigation.teams_with_edit_access.order(:name)
+    @collaborators = @investigation
```
slawosz, line 7: How about moving it to method with meaningful name?

StupidCodeFactory: yes, I will do that later :)

psd-web/app/decorators/investigation_decorator.rb
```diff
@@ -99,6 +99,10 @@ def generic_attachment_partial(viewing_user)
     "documents/generic_document_card"
   end
 
+  def owner
+    object.owner&.decorate
```
StupidCodeFactory, line 103: In reality every investigation has an owner. the `&` really only is there to cater for minitest's fixtures which are totally messed up and would take ages to fix. So I'm adding this here sadly.

psd-web/app/models/collaboration.rb
```diff
@@ -1,4 +1,8 @@
 class Collaboration < ApplicationRecord
-  belongs_to :investigation
+  belongs_to :investigation, optional: true
```
StupidCodeFactory, line 2: not sure why this was not enforced before. It now is!

psd-web/app/services/change_case_owner.rb
```diff
@@ -14,15 +14,23 @@ def call
     return if old_owner == owner
 
     ActiveRecord::Base.transaction do
-      if owner.is_a?(Team)
-        # Remove the new owner team as a collaborator if they were one.
-        investigation.edit_access_collaborations.where(editor: owner).delete_all
+      unless owner.team == investigation.owner_team_collaboration.collaborator
+        investigation.owner_team_collaboration.swap_to_edit_access!
```
StupidCodeFactory, line 18: this stops the previous owner being removed its edit access

psd-web/db/schema.rb
```diff
@@ -83,6 +83,8 @@
     t.text "message"
     t.string "type", null: false
     t.datetime "updated_at", null: false
+    t.index ["investigation_id", "collaborator_type"], name: "index_collaborations_on_investigation_id_and_collaborator_type", unique: true, where: "((collaborator_type)::text = 'Collaboration::Access::OwnerTeam'::text)"
+    t.index ["investigation_id", "collaborator_type"], name: "yann_unique_test", unique: true, where: "((collaborator_type)::text = 'Collaboration::Access::Owner'::text)"
```
StupidCodeFactory, line 87: oops need to remove this  on my local db and regenerate the schema!!

psd-web/db/migrate/20200622120007_create_unique_index_for_collaboration_and_owner.rb
```diff
@@ -0,0 +1,7 @@
+class CreateUniqueIndexForCollaborationAndOwner < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured do
+      add_index :collaborations, %i[investigation_id collaborator_type], unique: true, where: "collaborator_type = 'Collaboration::Access::OwnerTeam'"
```
StupidCodeFactory, line 4: This should ensure there is only one `OwnerTeam` per investigation

psd-web/app/models/team.rb
```diff
@@ -1,11 +1,15 @@
...
+  include TeamCollaboratorInterface
+
   belongs_to :organisation
   has_many :users, dependent: :restrict_with_exception
   has_many :investigations, dependent: :nullify, as: :owner
   has_many :collaborations, dependent: :destroy, as: :collaborator
 
   validates :name, presence: true
 
+  alias_attribute :email, :team_recipient_email
```
StupidCodeFactory, line 11: TODO: going to move this to `psd-web/app/models/concerns/team_collaborator_interface.rb`

psd-web/app/models/investigation.rb
```diff
@@ -55,19 +55,29 @@ class Investigation < ApplicationRecord
...
   has_one :creator_team, through: :creator_team_collaboration, dependent: :destroy, source_type: "Team"
   has_one :creator_user, through: :creator_user_collaboration, dependent: :destroy, source_type: "User"
 
-  has_one :owner_user_collaboration, dependent: :destroy, class_name: "Collaboration::OwnerUser"
-  has_one :owner_team_collaboration, dependent: :destroy, class_name: "Collaboration::OwnerTeam"
-  has_one :owner_team, through: :owner_team_collaboration, dependent: :destroy, source_type: "Team", required: true
-  has_one :owner_user, through: :owner_user_collaboration, dependent: :destroy, source_type: "User"
+  has_many :collaboration_access_owners, class_name: "Collaboration::Access::Owner"
+  has_one :owner_user_collaboration, class_name: "Collaboration::Access::OwnerUser", dependent: :destroy, inverse_of: :investigation
+  has_one :owner_team_collaboration, class_name: "Collaboration::Access::OwnerTeam", dependent: :destroy, required: true
```
StupidCodeFactory, line 78: Ideally I'd love to move those to `Collaboration::Access::Owner::Team` and `Collaboration::Access::Owner::User` and same for creator. wdyt?

psd-web/app/models/collaboration.rb
slorek: Why is it now `optional`?

psd-web/app/models/collaboration/access.rb
```diff
@@ -1,4 +1,6 @@
 class Collaboration < ApplicationRecord
   class Access < Collaboration
```
slorek, line 2: 🤔 is the `::Access` namespace pretty much redundant? We could just have `Collaboration::Read` and `Collaboration::Write` and `Collaboration::OwnerTeam` etc.

psd-web/app/models/concerns/team_collaborator_interface.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def user
+    nil
+  end
+
+  def in_same_team_as?(user)
+    users.include?(user)
+  end
+
+  def own!(investigation, collaborator = nil)
```
slorek, line 20: I can see how these `own!` methods are useful, I have a small concern that it's not obvious the `ChangeCaseOwner` service should be used, especially as you are calling `own!` directly in some of the tests.

psd-web/app/models/team.rb
slorek: It will be good to finally do this - we should also replace existing conditionals around `team_recipient_email` and `email`.

psd-web/app/models/user.rb
```diff
@@ -9,21 +11,18 @@ class User < ApplicationRecord
...
-  has_many :owner_user_collaboration, class_name: "Collaboration::OwnerUser", foreign_key: "collaborator_id"
-  has_many :investigations, through: :owner_user_collaboration, dependent: :nullify, as: :owner_user
+  has_many :owner_user_collaborations, class_name: "Collaboration::Access::OwnerUser", foreign_key: "collaborator_id"
+  has_many :investigations, through: :owner_user_collaborations, dependent: :nullify, as: :user
   has_many :activities, through: :investigations
   has_many :user_sources, dependent: :destroy
   has_many :user_roles, dependent: :destroy
   has_many :collaborations, dependent: :destroy, as: :collaborator
 
-  belongs_to :team
```
slorek, line 19: I'm not sure this belongs in `UserCollaboratorInterface` - it's not immediately obvious there's a 'hidden' relationship. I would keep it in here.

psd-web/app/services/change_case_owner.rb
slorek: I think we need to check, as I'm not actually sure, whether swapping the type is preferable - or if it should be destroyed and a new collaboration created. We would not have an `added_by_user` in this case and the timestamps would be different.

psd-web/app/services/change_case_owner.rb
```diff
@@ -14,15 +14,23 @@ def call
...
       end
 
-      investigation.update!(owner: owner)
+      investigation.reload # force cached associations to be reloaded
+
+      old_collaborator = investigation
+                           .collaboration_accesses
+                           .where.not(type: ["Collaboration::Access::OwnerUser", "Collaboration::Access::OwnerTeam"])
+                           .find_by(collaborator: owner)
```
slorek, line 26:
```suggestion
                           .find_by(collaborator: owner.team)
```
I think you might need to ensure you're checking for a `Team`, as `owner` could be a `User`.

psd-web/db/migrate/20200616142120_rename_edit_access_to_access_edit.rb
```diff
@@ -0,0 +1,5 @@
+class RenameEditAccessToAccessEdit < ActiveRecord::Migration[5.2]
+  def change
+    Collaboration.where(type: "Collaboration::EditAcess").update_all(type: "Collaboration::Access::Edit")
```
slorek, line 3: We should consider putting the app in maintenance mode or at least doing this out of hours to avoid errors.

psd-web/app/models/collaboration/access.rb
StupidCodeFactory: @slorek it is not no. Because the _creator_ records should not be retrieved when asking for the who has access, hence this parent class to differentiate between collaboration that have any kind of access to a  case or no access level collaboration.

psd-web/db/migrate/20200616142120_rename_edit_access_to_access_edit.rb
StupidCodeFactory: Migrations are run a transactions which locks the table, the update_all issue and SQL `UPDATE` so this should be really fast (as it does not trigger any callbacks or instantiate any active record object)

psd-web/spec/factories/investigations.rb
```diff
@@ -95,7 +96,7 @@
     # We need to do this before rather than after create because database
     # constraints on pretty_id need to be satisfied
     before(:create) do |investigation, options|
-      CreateCase.call(investigation: investigation, user: options.creator)
+      CreateCase.call(investigation: investigation, user: options.owner || options.creator)
```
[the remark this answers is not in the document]
StupidCodeFactory: I think I added this because some tests assumes that it is possble that a team can be a creator, in which case I will remove them.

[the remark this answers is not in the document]
slorek: I _thought_ I had fixed any instances, but there certainly shouldn't be any tests which try to set the creator to a team.

psd-web/app/services/change_case_owner.rb
StupidCodeFactory: the swap method now creates a new collaboration record

nasirkhanbeis:
picking up this now, please do not merge until I confirm.

PR #715, opened 2020-06-17
Add supporting information flow

scruti:
[Trello card](https://trello.com/c/N2GL9Emf/506-5-create-add-supporting-information-page-and-flow)

## Description

We are moving logic from "Add activity" page to "Add Supporting Information", that introduces a page that will act as a router to different existing actions/flows depending on the user selection.

When choosing "Correspondence", the user will be sent to a second selection page where can choose between the 3 different types of correspondence (email, meeting and phone call).  

- Redirections changed so when finishing adding a supporting information element we get sent to the supporting information tab.
- Added some helpers to redirect to the right `attachments/activity/images/supporting-information` tab depending on the parent of the attachment (business, product, investigation) and the type of attachment (is an image?)  when possible. This will also affect other flows like when a document gets uploaded/edited/deleted.
- For things like "Add a comment" that can be accessed from different tabs, defaulted the back link to Activity tab. 
Experimented with using the request referrer to generate a dynamic back link but in case of submission validation error the referrer was becoming the same page, so ended discarding this approach.

Screenshot from 2020-06-22 10-03-59
Screenshot from 2020-06-22 10-04-08
Screenshot from 2020-06-22 10-04-21
Screenshot from 2020-06-22 10-04-34
Screenshot from 2020-06-22 10-05-12
Screenshot from 2020-06-22 10-08-42

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


psd-web/app/controllers/investigations/correspondence_controller.rb
```diff
@@ -10,9 +10,24 @@ class Investigations::CorrespondenceController < ApplicationController
...
+    case params[:correspondence_type]
+    when "email"
+      redirect_to new_investigation_email_path(@investigation)
+    when "meeting"
+      redirect_to new_investigation_meeting_path(@investigation)
+    when "phone_call"
+      redirect_to new_investigation_phone_call_path(@investigation)
+    else
+      @correspondence_type_empty = true
+    end
```
scruti, line 30: I am not totally convinced on this bit.
Correspondence Controller has been some sort of abstract controller until now.
When having a page to "add new correspondence" the concept brings us naturally to this `Correspondence#new` action. 
The downside is that we force all the subclasses to re-write the `new` action.
I didn't move the `case` logic into `Correspondence#create` to keep this as contained as possible until we decide to keep this or a different better approach is found.

psd-web/app/views/documents/_generic_document_card.html.erb
```diff
@@ -40,7 +40,4 @@
 
   </div>
 
-
-
-
```
scruti, line 45: This was introduced in a merge conflict resolution. Can undo it if we consider shouldn't have been introduced.

psd-web/app/controllers/investigations/correspondence_controller.rb
StupidCodeFactory: I don't think it's a big deal at all to have to overload the new actions in the subclasses, if anything it's a bit more declarative.

I would introduce a new controller to just handle the redirection part, not sure how I'd call it: `CorrespondenceRoutingController` (does not sound great but at least it states its purpose clearly. 

This would allow you to remove the  `return unless params[:commit] == "Continue"` and in the new action just take care of the breadcrumb.
In the other controller you'd redirect if the form object would not be valid, i.e it does not have a correspondence type chosen. This would also be a lot cleaner in the view.

psd-web/app/controllers/investigations/supporting_information_controller.rb
```diff
@@ -1,15 +1,51 @@
 module Investigations
   class SupportingInformationController < ApplicationController
     def index
-      investigation = Investigation
-                        .find_by!(pretty_id: params[:investigation_pretty_id])
+      set_investigation
```
StupidCodeFactory, line 4: I think you can get rid of the set_investigation and just rewrite the `set_investigation`  method like this:
```ruby
def investigation
  @investigation ||= investigation
end
```
and then just call 

```ruby
authorize investigation, :view_non_protected_details?
```
this removes two lines of unnecessary code.

psd-web/app/views/investigations/correspondence/new.html.erb
```diff
@@ -0,0 +1,37 @@
...
+    href: new_investigation_supporting_information_path(@investigation)
+  ) %>
+<% end %>
+
+<%
+  correspondence_types = {
+      "email": "Record email",
+      "meeting": "Record meeting",
+      "phone_call": "Record phone call"
+  }
```
StupidCodeFactory, line 14: this would be better located as constant on a form object.

psd-web/app/views/investigations/correspondence/new.html.erb
```diff
@@ -0,0 +1,37 @@
...
+%>
+
+<%= form_with local: true, url: new_investigation_correspondence_path(@investigation), method: :get do |form| %>
+  <% error_message = @correspondence_type_empty && { text: "Correspondence type must not be empty", href: "#activity_type-error" } %>
+  <div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+  <% if error_message %>
+    <%= render "components/govuk_error_summary", titleText: "There is a problem", errorList: [error_message] %>
```
StupidCodeFactory, line 22: there is a helper for this: `ApplicationHelper#error_summary`

scruti:
Great suggestions @StupidCodeFactory , jumping on them :clap:

nasirkhanbeis:
Hi @scruti I feel we also need to need to address some back navigation issues along with this feature, as otherwise it may break whole journey. 
such as
1. When I select an option eg. Test result and if I want to go immediate back then I can't. Back link is expected
2. When I add a thing, I expect to return back to supporting information tab
3. Edit attachments pages, back link is suppose to return to supporting information flow( but we can argue as this is existing behaviour)
4. When I'm recording correspondence, and I want to change type to email from meeting, then there is no way to go immediate back. I think "Back to allegation<id>" will not be applicable on this page.

psd-web/app/forms/supporting_information_form.rb
```diff
@@ -0,0 +1,17 @@
+class SupportingInformationForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  AVAILABLE_TYPES = {
+    comment: "Comment or case note",
```
edwardhorsford, line 6:
```suggestion
    comment: "Comment",
```

Sorry, 'case note' was a speculative thing for the future.

psd-web/app/views/investigations/correspondence/new.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    text: "Back to #{@investigation.type.demodulize.downcase}",
+    href: new_investigation_supporting_information_path(@investigation)
+  ) %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+    <%= error_summary @correspondence_form.errors %>
+    <% header = capture do %>
+      <h1 class="govuk-heading-l"> What type of correspondence are you adding? </h1>
```
edwardhorsford, line 13: Please change: you don't need to capture html like this. You can pass the text directly to the component and set `isPageHeading: true`. Might need to pass a class too, which would be `govuk-fieldset__legend--l` in this case.

psd-web/app/views/investigations/correspondence/new.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    <% end %>
+    <%= form_with local: true, url: new_investigation_correspondence_path(@investigation), method: :get do |form| %>
+      <%= render "form_components/govuk_radios",
+                 form: form,
+                 key: :type,
+                 errorMessage: format_errors_for(@correspondence_form,
+                                                 @correspondence_form.errors.full_messages_for(:type)),
+                 items: CorrespondenceForm::AVAILABLE_TYPES.map { |type, label| { text: label, value: type } },
+                 fieldset: { legend: { html: header } } %>
+      <%= form.submit "Continue", class: "govuk-button" %>
```
edwardhorsford, line 23: I think we might have a helper for govuk buttons - `govukButton`

psd-web/app/views/investigations/correspondence/new.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    <% end %>
+    <p class="govuk-body">
+      <%= link_to 'Cancel', investigation_supporting_information_index_path(@investigation) %>
```
edwardhorsford, line 26: This probably needs the `govuk-link` class.

psd-web/app/views/investigations/supporting_information/new.html.erb
```diff
@@ -0,0 +1,38 @@
+<%= render "investigations/pages_top" %>
```
edwardhorsford, line 1: Needs a page title, and to pass the errors up to the title

psd-web/app/views/investigations/supporting_information/new.html.erb
```diff
@@ -0,0 +1,38 @@
...
+                    items.push divider: "or"
+                    items.push text: SupportingInformationForm::AVAILABLE_TYPES[:generic_information], value: :generic_information
+                  end
+%>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+    <%= error_summary @supporting_information_form.errors %>
+    <% header = capture do %>
+      <h1 class="govuk-heading-l"> What type of information are you adding? </h1>
```
edwardhorsford, line 22: As above, the component can build headings for you.

psd-web/app/views/investigations/supporting_information/new.html.erb
```diff
@@ -0,0 +1,38 @@
...
+    <% end %>
+    <%= form_with local: true, url: investigation_supporting_information_index_path(@investigation) do |form| %>
+      <%= render "form_components/govuk_radios",
+                 form: form,
+                 key: :type,
+                 errorMessage: format_errors_for(@supporting_information_form,
+                                                 @supporting_information_form.errors.full_messages_for(:type)),
+                 items: radio_items,
+                 fieldset: { legend: { html: header } } %>
+      <%= form.submit "Continue", class: "govuk-button" %>
```
edwardhorsford, line 32: As above, we have a component for this.

psd-web/app/forms/correspondence_routing_form.rb
```diff
@@ -0,0 +1,14 @@
+class CorrespondenceRoutingForm
```
frankieroberto, line 1: suggestion: rename to `TypeOfCorrespondenceForm` to better match the question ("What type of correspondence are you adding?")

psd-web/app/views/investigations/supporting_information/new.html.erb
```diff
@@ -0,0 +1,36 @@
...
+  ) %>
+<% end %>
+
+<%
+  radio_items = SupportingInformationForm::AVAILABLE_TYPES
+                  .except(:generic_information)
+                  .map { |type, label| { text: label, value: type } }.tap do |items|
+                    items.push divider: "or"
+                    items.push text: SupportingInformationForm::AVAILABLE_TYPES[:generic_information], value: :generic_information
+                  end
```
frankieroberto, line 16: I wonder if you could move more of this to the form object by having something like

```ruby
MAIN_TYPES = [ {comment: "Comment"}, ...  ]
GENERIC_TYPE = { generic_information: "Other document or attachment" }
AVAILABLE_TYPES = MAIN_TYPES.merge(GENERIC_TYPE)
```

Perhaps also the transforming of the hash into the array could become a helper, as this is used in a few places:

```ruby
def radio_items_from_hash(hash)
  hash.collect do |key, value|
    {value: key, text: value}
  end
end
```

We could also consider moving the label text into the locale file, and then using something like

```ruby
AVAILABLE_TYPES = [
  :comment,
  :corrective_action,
  :correspondence
].freeze
```

```ruby
def radio_items(keys, locale_scope)
  keys.collect do |key|
    {value: key, text: I18n.t(key, scope: locale_scope)}
  end
end
```

These are just ideas though, and could be future refactorings.

psd-web/config/locales/en.yml
```diff
@@ -146,20 +159,14 @@ en:
               not_whitelisted: "The email address is not recognised. Check you’ve entered it correctly, or email %{opss_enquiries_email} to add it to the approved list."
               already_in_team: "%{email} is already a member of %{team}"
               existing_user: "You cannot invite this person to join your team because they are already a member of another team. Contact %{opss_enquiries_email} if the person’s team needs to be changed."
-
+        supporting_information_form:
+          attributes:
+            type:
+              blank: "Supporting information type must not be empty"
```
edwardhorsford, line 165:
```suggestion
              blank: "Select the type of information you’re adding"
```

psd-web/config/locales/en.yml
```diff
@@ -137,6 +137,19 @@ en:
             owner_id:
               blank: Select case owner
               not_found: User or team not found
+        correspondence_routing_form:
+          attributes:
+            type:
+              blank: "Correspondence type must not be empty"
```
edwardhorsford, line 143:
```suggestion
              blank: "Select the type of correspondence you’re adding"
```

psd-web/app/views/investigations/supporting_information/new.html.erb
scruti: This are great suggestions and is not like a massive refactor. Going to apply them, thanks!

psd-web/app/controllers/investigations/supporting_information_controller.rb
```diff
@@ -1,21 +1,50 @@
...
+    def new
+      authorize investigation, :update?
+      supporting_information_type_form
+    end
+
+    def create
+      authorize investigation, :update?
+      return render(:new) if supporting_information_type_form.invalid?
+
+      case supporting_information_type_form.type
```
slawosz, line 24: Maybe you should extract this redirecting to method?

psd-web/app/forms/supporting_information_type_form.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  MAIN_TYPES = {
+    comment: "Comment",
+    corrective_action: "Corrective action",
+    correspondence: "Correspondence",
+    image: "Image",
+    testing_result: "Test result"
+  }.freeze
+  GENERIC_TYPE = { generic_information: "Other document or attachment" }.freeze
+  AVAILABLE_TYPES = MAIN_TYPES.merge(GENERIC_TYPE)
```
slawosz, line 13: Do we really need to distinguish 2 different types?

psd-web/config/routes.rb
```diff
@@ -138,6 +138,8 @@
 
     resources :ownership, controller: "investigations/ownership", only: %i[show new create update], path: "assign"
     resources :corrective_actions, controller: "investigations/record_corrective_actions", only: %i[show new create update]
+
+    resources :correspondence, controller: "investigations/correspondence_routing", only: %i[new create]
```
StupidCodeFactory, line 142:
```suggestion
    resources :correspondence, controller: "investigations/correspondence-routing", only: %i[new create]
```

psd-web/app/forms/supporting_information_type_form.rb
scruti: This is mainly caused because the `or` in the radio button options that forces us to deal with 2 lists or a single list but doing `exclude last element` and then `just last element` when building the radio buttons. 
Having two different elements helps to simplify his current usage.

psd-web/spec/requests/add_correspondence_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+RSpec.describe "Adding correspondence to a case", type: :request, with_stubbed_mailer: true, with_stubbed_elasticsearch: true do
+  let(:user) { create(:user, :activated, has_viewed_introduction: true) }
+  let(:investigation) { create(:allegation, owner: user.team) }
+
+  before { sign_in(user) }
+
+  context "when not choosing any option" do
+    before do
+      post investigation_correspondence_index_path(investigation),
+           params: { commit: "Continue" }
```
StupidCodeFactory, line 12: ✂️  I don't think you need to test on the value of the commit button anymore.

psd-web/spec/requests/add_correspondence_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    it "renders an error" do
+      expect(response.body).to include("Select the type of correspondence you’re adding")
+    end
+  end
+
+  context  "when choosing a correspondence type" do
+    before do
+      post investigation_correspondence_index_path(investigation),
+           params: {
+             commit: "Continue",
```
StupidCodeFactory, line 28: ✂️

psd-web/app/controllers/investigations/supporting_information_controller.rb
scruti: I didn't extract it because, even by adding a bunch on lines, this method responsibility becomes quite explicit as a "route to X/Y/Z". So I personally didn't see any added value into breaking it.

psd-web/spec/requests/add_supporting_information_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+RSpec.describe "Adding supporting information to a case", type: :request, with_stubbed_mailer: true, with_stubbed_elasticsearch: true do
+  let(:user) { create(:user, :activated, has_viewed_introduction: true) }
+  let(:investigation) { create(:allegation, owner: user.team) }
+
+  before { sign_in(user) }
+
+  context "when not choosing any option" do
+    before do
+      post investigation_supporting_information_index_path(investigation),
+           params: { commit: "Continue" }
```
StupidCodeFactory, line 12: ✂️  I don't think you need to test the value of the commit parameter

psd-web/spec/requests/add_correspondence_spec.rb
scruti: Good catch!

psd-web/config/routes.rb
scruti: This is the actual file name for the controller, not the path (the path will be `correspondence/new`

PR #723, opened 2020-06-19
Change export columns to remove complainant and add case owner team

slorek:
https://trello.com/c/eHNM3ToZ/588-update-data-export-columns

## Description
Removes complainant details and adds owner team/user following recent changes so the owner is correctly represented. If there is no user assigned the column will be empty.

## Checklist:


psd-web/app/views/investigations/index.xlsx.axlsx
```diff
@@ -15,11 +15,9 @@ book.add_worksheet name: "Cases" do |sheet_investigations| # rubocop:disable Met
       investigation.product_category,
       investigation.hazard_type,
       investigation.coronavirus_related?,
-      investigation.owner ? investigation.owner.decorate.display_name(viewer: current_user) : "No case owner",
+      investigation.owner_team&.name,
```
slorek, line 18: There should always be an `owner_team` but in the legacy Minitest suite the fixtures create cases with no owner.

PR #724, opened 2020-06-19
Remove redundant minitest tests

frankieroberto:
This removes 3 of the minitest system tests which are covered by rspec feature specs:

* `RecordMeetingCorrespondenceTest` is covered by `record_meeting_activity_spec.rb`
* `AlertTest` is covered by `send_alert_spec.rb`
* `CreateNewRecordTest ` is covered by:
  * `create_project_spec.rb`
  * `create_allegation_as_opss_user_spec.rb`
  * `create_enquiry_as_opss_user_spec.rb`


frankieroberto:
@slorek thanks for looking into those in more detail than I did 😉 !

I've updated the feature scenarios now to expand the coverage to match that provided by the old mintiest system tests. Could you re-review?

psd-web/spec/features/record_meeting_activity_spec.rb
```diff
@@ -101,6 +101,51 @@
     expect_case_activity_page_to_show_meeting_information
   end
 
+  scenario "Adding a meeting using a just a summary and notes" do
```
slorek, line 104:
```suggestion
  scenario "Adding a meeting using just a summary and notes" do
```

PR #725, opened 2020-06-19
Remove OPSS and creator/owner org users ability to view restricted cases where they are not a collaborator

slorek:
https://trello.com/c/1N8vvk4x/533-2-remove-opss-blanket-ability-to-view-restricted-cases

## Description
This removes a user's ability to view a restricted case when their team was not added to the case. Currently a user can view it if they are an OPSS user, or they are in the same organisation as the owner or original creator of the case.

The test coverage for viewing a restricted case was incomplete and didn't cover most of those scenarios; therefore I didn't need to change the tests too much. The current test coverage now accurately reflects the behaviour as implemented.

⚠️ **DO NOT MERGE**: We need to communicate this change to affected users first. Will work with @tobyhughes on this.

## Checklist:

### DIFF
--- a/psd-web/app/policies/investigation_policy.rb
+++ b/psd-web/app/policies/investigation_policy.rb
@@ -33,9 +33,7 @@ class InvestigationPolicy < ApplicationPolicy
     return true unless private
 
     # Has the user's team been added to the case as a collaborator?
-    return true if @record.teams_with_access.include?(user.team)
-
-    false
+    @record.teams_with_access.include?(user.team)
   end
 
   def view_protected_details?(user: @user)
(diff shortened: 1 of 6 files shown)


psd-web/app/policies/investigation_policy.rb
```diff
@@ -26,12 +26,9 @@ def manage_collaborators?
   # with businesses.
   def view_non_protected_details?(user: @user, private: @record.is_private)
     return true unless private
-    return true if user.is_opss?
-    return true if @record.owner.present? && (@record.owner&.organisation == user.organisation)
-    return true if @record.creator_user&.has_gdpr_access?(user)
 
     # Has the user's team been added to the case as a collaborator?
-    return true if @record.teams_with_edit_access.include?(user.team)
+    return true if @record.teams_with_access.include?(user.team)
```
frankieroberto, line 31: remove `return true if` and the implicit `false` return on line 33, now that this is the only condition left?

PR #729, opened 2020-06-22
Replace minitests with specs

slorek:
## Description
Replaces some deprecated Minitest tests with equivalent specs. A starting point to improve test coverage on some of these areas.


psd-web/spec/models/location_spec.rb
```diff
@@ -0,0 +1,24 @@
+require "rails_helper"
+
+RSpec.describe Location do
+  subject(:location) { described_class.new(county: county, country: country) }
+
+  let(:county) { "L" }
+  let(:country) { "C" }
```
frankieroberto, line 7: It’d be nice if these were a little more realistic?

psd-web/spec/models/complainant_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    context "with no investigation" do
+      let(:investigation) { nil }
+
+      it "generates errors and does not save the record", :aggregate_failures do
+        expect(complainant).not_to be_persisted
+        expect(complainant.errors).not_to be_empty
+      end
+    end
+  end
```
frankieroberto, line 26: Feels a bit weird that these are describing the `#save` method, which is provided by the framework rather than our code. At the very least I'd expect this to be testing whether the record is valid, rather than whether it saves or not?

I'm also kinda confused by it as the Complainant class has:

```ruby
  belongs_to :investigation, optional: true
```

...which implies that investigation ought not to be required?

psd-web/spec/models/business_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    context "with blank trading_name" do
+      let(:trading_name) { nil }
+
+      it "returns false" do
+        expect(business).not_to be_valid
+      end
+    end
+  end
```
frankieroberto, line 22: I'm personally a bit wary of the value of these kind of validation tests, as they're purely asserting usage of `validates :trading_name, presence: true` rather than any custom validation we've written. But 🤷‍♂️

slorek: Agreed - but if it's important to maintain data integrity the tests can be a useful firestop to prevent the validation being changed or removed.

psd-web/spec/models/complainant_spec.rb
slorek: There are different behaviours on `create` and `update`. The key other line is https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/models/complainant.rb#L17

This is why I tested it on `#save` rather than `#valid?` as in other validation specs.

psd-web/spec/models/location_spec.rb
slorek: It seems to be the interface to this model, and is the same as the Minitest. I've no idea what `L` and `C` mean though. We can definitely improve it, but for the purposes of replacing the Minitest I think it does the job.

psd-web/spec/models/complainant_spec.rb
frankieroberto: Ah, I didn’t spot that line. Wonder why it’s only validating on create and update... 🤔

I think you _could_ test this validation by calling `valid?(:create)` to avoid the database call, but it’s probably more understandable as it is. 👍🏻

psd-web/spec/models/business_spec.rb
frankieroberto: I’ve worked on rails projects where every validation has a model unit test, but then I’ve also worked on others where none of them do (but where some of the validations are covered by integration/feature tests)... Tricky one to balance!

psd-web/spec/models/complainant_spec.rb
slorek: There was a comment about old records, so maybe there's some legacy data that needs to be supported.

PR #733, opened 2020-06-23
Case bar

slorek:
https://trello.com/c/WaHfLJ0L/210-3-add-case-bar-without-actions

## Description

This adds a title bar underneath the primary nav which displays the case title. The bar is shown across all case pages. In a future story there will be a menu added on the right side.

### TO DO:
* Possibly tweak the padding above the sub nav, since the case badges add a bit of extra space not seen in the prototype (question for @edwardhorsford @KamQureshi)

I've re-used the `pages_top` partial across all pages for consistency. Optionally, this template now shows the sub-nav. I have not enabled this on pages where it was not previously, but it is now trivial to do so by supplying the `current_tab` variable when rendering the partial.

Note: I have not added the case bar to the view product/business pages. This is because they link to the products/businesses sections in the top-level nav. The back link is currently changed conditionally according to the `referer` request header, but it would be a much bigger change to include the case bar - and in this case I think the top-level nav should stay on 'cases'. one for discussion and possibly a future story.

Since we are now importing the [HMCTS design system](http://hmcts-design-system.herokuapp.com/) into our project in order to use the [identity bar component](http://hmcts-design-system.herokuapp.com/components/identity-bar), we can now also use their CSS for the [banner component](http://hmcts-design-system.herokuapp.com/components/banner), [badge component](http://hmcts-design-system.herokuapp.com/components/badge) and [sub-nav component](http://hmcts-design-system.herokuapp.com/components/sub-navigation), instead of the duplicated copies we had. I have therefore made this change as well.

As well as the core changes I have done the following:

* Converted a large number of Slim templates to ERB (there may be bugs...)
* Changed all case page back links to use the `govukBackLink` helper
* Fixed the layout of the 'remove product' and 'remove business' pages
* Fixed back links on add/remove product pages to link back to the products tab rather than the overview tab
* Fixed back link text on case team pages for consistency
* Fixed page container width on first 'add business' wizard page
* Fixed back links on add/remove business pages to link back to the businesses tab rather than the overview tab
* Fixed back link on 'add business details' page (second in wizard flow) to link to the previous wizard page (as consistent with send alert)
* Fixed back links on all of the wizard pages

Note: the back link on the add image/document pages links back to the 'supporting information' tab, but it is also possible to get to this from the 'images' tab. I have not attempted to fix this existing issue.

#### Case overview
Screenshot 2020-06-24 at 11 20 38

### Viewing supporting information
Screenshot 2020-06-24 at 11 20 50

### Adding supporting information
Screenshot 2020-06-24 at 11 21 06

## Checklist:

### General testing (author)
- [x] Test without JavaScript
- [x] Test on small screen

- [x] Works keyboard only


KamQureshi:
@slorek all good, the padding above the sub nav is fine

KamQureshi:
@slorek all good, the padding above the sub nav is fine

psd-web/app/assets/stylesheets/case-bar.scss
```diff
@@ -0,0 +1,10 @@
+.app--case-bar .app--case-bar-details {
+  display: block;
+}
+
+.app--case-bar .app--case-bar-title {
+  display: block;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
```
frankieroberto, line 10: Could you rename this file to start with an underscore (ie `_case-bar.scss`) to match the convention for sass partials? (We don't yet follow this consistently, but we've started to). The `@import` doesn't need to include the underscore though.

psd-web/package.json
```diff
@@ -7,6 +7,7 @@
     "yarn": "1.22.x"
   },
   "dependencies": {
+    "@hmcts/frontend": "^0.0.41-alpha",
```
frankieroberto, line 10: I’m not totally convinced we should be importing this using node, rather than copying and pasting (forking) it? Especially as it’s only in alpha and hasn't been touched since December? Pros and cons though...

psd-web/app/assets/stylesheets/case-bar.scss
frankieroberto, line 10: I think this should be:

```suggestion
.app-case-bar .app-case-bar__details {
  display: block;
}

.app-case-bar .app-case-bar__title {
  display: block;
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
}
```

...following the [BEM conventions](http://getbem.com/naming/). ie `app-` is the prefix, `__` is used to denote a child element, and `--` is used for a modifier.

psd-web/app/views/investigations/alerts/about_alerts.html.erb
```diff
@@ -0,0 +1,51 @@
...
+      <span class="govuk-caption-l"><%= @investigation.pretty_description %></span>
+
+      <h1 class="govuk-heading-l">You cannot send an alert about a restricted case</h1>
+      <p>Email alerts can only be sent for cases that are not restricted. To send an alert about this case you need to unrestrict it.</p>
+
+      <%= link_to "Change case visibility",
+                  visibility_investigation_path(@investigation),
+                  class: "govuk-button",
+                  role: "button",
+                  draggable: false %>
```
frankieroberto, line 23: We have a helper for this, which adds the class and other attributes:

```suggestion
      <%= govukButton( 
         text: "Change case visibility",
         href: visibility_investigation_path(@investigation)
         ) %>
```

psd-web/app/views/investigations/_pages_top.html.erb
```diff
@@ -1,15 +1,34 @@
+<% content_for :action_bar do %>
+  <div class="hmcts-identity-bar app--case-bar">
+    <div class="hmcts-identity-bar__container">
+      <div class="hmcts-identity-bar__details app--case-bar-details">
+        <span class="hmcts-identity-bar__title app--case-bar-title"><%= @investigation.pretty_id %> &nbsp;&nbsp; <%= @investigation.decorate.title %></span>
```
frankieroberto, line 5: It would be good to avoid ` &nbsp;&nbsp;` and add the spacing using CSS instead.

psd-web/package.json
slorek: I think it is preferable to import it rather than duplicating the code - it's the same way we're using the govuk design system.

psd-web/app/assets/stylesheets/case-bar.scss
slorek: 👍 thanks for pointing to the docs.

psd-web/app/views/investigations/_pages_top.html.erb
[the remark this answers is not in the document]
slorek: I did consider this, but it would mean even more changes to other templates which were already calling it. But now I've gone this far I might as well.

slorek: Mmm, this was copied from the prototype and I know @edwardhorsford is very particular about his markup ;)

nasirkhanbeis:
@slorek  I'm testing this manually, please don't merge until I confirm :)

nasirkhanbeis:
Re: navigations, to make it consistent would be nice if we could fix back link on view product and business pages to return back to respective tab instead returning to case overview page.

frankieroberto:
@nasirkhanbeis re:

> navigations, to make it consistent would be nice if we could fix back link on view product and business pages to return back to respective tab instead returning to case overview page.

I don't think this is related to this PR at all? Can you raise that as a separate issue?

psd-web/app/views/investigations/_pages_top.html.erb
```diff
@@ -1,15 +1,34 @@
+<% content_for :action_bar do %>
+  <div class="hmcts-identity-bar app--case-bar">
+    <div class="hmcts-identity-bar__container">
+      <div class="hmcts-identity-bar__details app--case-bar-details">
```
frankieroberto, line 4: It kinda feels weird and unexpected to me that we’re mashing up `hmcts-identity-bar` and `app--case-bar`. I think we should either use `app-identity-bar` for our override classes, or copy-paste the `hmcts-identity-bar` css and rename it to `app-case-bar` and have it all in a single file?

psd-web/app/views/investigations/_pages_top.html.erb
edwardhorsford: In this case I'm pretty sure I did it as a hack to avoid writing css. Suggest doing however is the "correct" way.

psd-web/app/views/investigations/_pages_top.html.erb
```diff
@@ -1,15 +1,34 @@
...
+        <span class="hmcts-identity-bar__title app--case-bar-title"><%= @investigation.pretty_id %> &nbsp;&nbsp; <%= @investigation.decorate.title %></span>
+      </div>
+    </div>
+  </div>
+<% end %>
+
 <% content_for :after_header do %>
   <div class="govuk-grid-row">
```
edwardhorsford, line 12: This currently produces an empty grid row for pages that don't have a back link.

psd-web/package.json
edwardhorsford: I think we should consider bringing this within our app. The HMCTS repo looks abandoned - so we're not going to get any updates from it. The things we've used are rather minor, so it feels worth taking ownership of them - which allows us to update as we want to.

psd-web/app/views/investigations/_pages_top.html.erb
slorek: I felt that in our context "identity bar" doesn't really describe what it is, so I went with something more descriptive. I don't want to copy and paste their code unless we need to change it significantly.

psd-web/app/views/investigations/_pages_top.html.erb
slorek: OK, I'll add some spacing with CSS. This will be better markup anyway.

psd-web/app/views/investigations/_pages_top.html.erb
frankieroberto: The Design System convention for ["small modifications to components"](https://design-system.service.gov.uk/get-started/extending-and-modifying-components/#small-modifications-to-components) is to use the same css class name, but with the prefix swapped to `app-` and a modifier at the end, eg we’d have to use `app-identity-bar__title--truncated`.

My hunch is that we'll end up changing the component significantly though, in which case copying-and-pasting would make more sense. But we could always postpone that until it happens.

psd-web/app/views/investigations/alerts/about_alerts.html.erb
slorek: Happy to change it, though it was just converted from the Slim.

frankieroberto: Ah yeah, sorry – forgot this was a conversion job. We can always do a second pass at improving those later.

psd-web/app/views/investigations/_pages_top.html.erb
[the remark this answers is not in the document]
frankieroberto: 👍

psd-web/app/views/investigations/_pages_top.html.erb
slorek: Thanks for the link - I'll change the name in that case then.

slorek: The link you provided also provides guidance about forking when making large modifications - so I think we're using he right approach here 👍

psd-web/app/assets/stylesheets/case-bar.scss
slorek: OK, I'll change the other files as well so this is clear.

psd-web/package.json
slorek: I think unless we are making significant changes we should import the code from the official library.

psd-web/app/views/investigations/_pages_top.html.erb
slorek: Good spot!

slorek:
> se overview page.

I

slorek:
> Re: navigations, to make it consistent would be nice if we could fix back link on view product and business pages to return back to respective tab instead returning to case overview page.

I'm going to leave this for another change as it's not straightforward.

slorek:
I've reduced the space between the badges and sub nav following conversation with @edwardhorsford

.github/workflows/overrides/case_bar.env
```diff
@@ -0,0 +1 @@
+export DB_NAME=case_bar
```
frankieroberto, line 1: ✂️

slorek: Will delete it in a follow up as otherwise it'll take >20 mins to re-deploy.

PR #739, opened 2020-06-24
Allow all users to view case images

slorek:
https://trello.com/c/PZ79IE49/634-images-are-not-supposed-to-be-restricted-for-outside-team-members

## Description
Fixes a bug where users on teams not added to the case can't view case images.

The affected template is now only used on case images and business attachments, so there is no need for any access control on the template. Access to restricted cases is already guarded by the controller.

There were no specific tests for this functionality so it was not necessary to modify any existing tests. I don't think we need to add a specific test that the images are visible - generally we only need to add specific tests for things which the user shouldn't be able to see. I added a test for the activity timeline as the behaviour is different depending on whether the attachment is an image or not.

## Checklist:

### DIFF
--- a/psd-web/app/views/documents/_document_card.html.erb
+++ b/psd-web/app/views/documents/_document_card.html.erb
@@ -1,37 +1,31 @@
 <%
-  restricted = parent.is_a?(Investigation) && !policy(parent).view_protected_details?
-  title = restricted ? "Attachment restricted" : document.title
   safe = document.metadata["safe"]
   document = document.decorate
 %>
 
 <div class="govuk-grid-row govuk-!-padding-bottom-6">
   <div class="govuk-grid-column-one-quarter">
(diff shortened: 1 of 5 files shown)


psd-web/app/views/documents/_document_card.html.erb
```diff
@@ -1,37 +1,31 @@
...
+    <p class="govuk-body"><%= document.description %></p>
 
-      <% if safe %>
-        <%= link_with_hidden_text_to "View #{pretty_type_description(document)}",
-           "(#{title}, opens in a new window)",
-           document,
-           class: "govuk-link app-block-link govuk-!-margin-right-3",
-           target: '_blank',
-           rel: 'noopener' %>
+    <% if safe %>
```
slawosz, line 19: This is probably not related to this `PR`, but `safe` is too short variable name, as it does not describe what it does.

PR #740, opened 2020-06-25
Use isolated database for each review app

slorek:
## Description
Following discussion with the team it was agreed we would change the default behaviour for review apps to use a new isolated database instance. This will increase the time needed to deploy a branch for the first time, but will help resolve issues with data changes in different branches. I have retained the override functionality so that if needed, a shared database instance could be used if one was created or maintained. It also allows us to continue overriding other variables as needed.

### DIFF
--- a/.github/workflows/review-apps.yml
+++ b/.github/workflows/review-apps.yml
@@ -54,7 +54,7 @@ jobs:
         cf7 auth
         cf7 target -o 'beis-opss' -s $SPACE
         export APP_NAME=psd-pr-$PR_NUMBER
-        export DB_NAME=psd-db-$BRANCH
+        export DB_NAME=psd-db-pr-$PR_NUMBER
         echo "Deploying review app. If fails, use command below to deploy manually:"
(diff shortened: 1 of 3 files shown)


.github/workflows/review-apps.yml
```diff
@@ -53,19 +53,10 @@ jobs:
         cf7 api api.london.cloud.service.gov.uk
         cf7 auth
         cf7 target -o 'beis-opss' -s $SPACE
-        export DB_VERSION=`cat psd-web/db/schema.rb | grep 'ActiveRecord::Schema.define' | grep -o '[0-9_]\+'`
         export APP_NAME=psd-pr-$PR_NUMBER
-        export DB_NAME=psd-db-$DB_VERSION
+        export DB_NAME=psd-db-$BRANCH
```
frankieroberto, line 57: I’d suggest using the `$PR_NUMBER` rather than the branch name, as it's possible to have really long branch names containing characters that probably aren't allowed by PaaS (eg the dependabot ones which contain slashes).

slorek: Good shout.

PR #744, opened 2020-06-26
Fix seeds

StupidCodeFactory:
## CHANGES:

Recent changes meant that none of the seed users/team in review app have
accesses to any cases


StupidCodeFactory:
Closing this PR as it seems that it is not an issue not to have access to any cases or any teams in PR.

PR #745, opened 2020-06-26
Remove ability to add test requests

frankieroberto:
This refactors the test result controllers and view, as they no longer need to handle "test requests", which we no longer support.

## Implementation notes

The main thing this does is to merge `Investigations::TestResultsController` with `Investigations::TestsController` as they now both refer only to "test results" and not to "test requests".  To do this meant removing the WickedWizard pattern, as that uses the `show` action for the form pages, but this is now needed for the "view tests result" page.

As well as the standard Rails methods for CRUD actions, I've added these two which handle the "check your answers" flow:

* `POST  #create_draft` -> this accepts all the params, but saves them to the session instead of the database.
* `GET #confirm` -> this renders the "check your answers" style page, reading the values from the session.

The `#create` method now converts the test result saved in the session to an actual database record.

The tests are largely unchanged, except to replace input name references with label text references, as the input name prefixes have changed from `test[attribute_name]` to `test_result[attribute_name]`.


psd-web/app/controllers/investigations/test_results_controller.rb
```diff
@@ -1,20 +1,71 @@
 class Investigations::TestResultsController < ApplicationController
+  include FileConcern
+  set_attachment_names :file
+  set_file_params_key :test_result
+
+  def new
+    @investigation = investigation_from_params
+    authorize @investigation, :update?
+    @test_result = build_test_result_from_params
+    set_attachment
```
slorek, line 10: Suggestion: I realise you don't like setting global variables in `before_filter` methods, but this block of code is repeated in each of the first four actions, so I think there is a case for extracting it.

I also realise that `set_attachment` is an existing name, but it doesn't clearly describe what it does. So there's an opportunity if we extract this to a method to come up with a descriptive name.

psd-web/app/controllers/investigations/test_results_controller.rb
```diff
@@ -1,20 +1,71 @@
 class Investigations::TestResultsController < ApplicationController
+  include FileConcern
+  set_attachment_names :file
+  set_file_params_key :test_result
+
+  def new
+    @investigation = investigation_from_params
```
slorek, line 7: Please change: we should either wipe `session[:test_result]` here, or use a more unique session key (e.g. `"test_result_#{@investigation.id}"`. The problem this causes is when the user tries to add a test result to a second case - the form will be pre-filled if there is existing data in the session. Using a more unique key will mean users will not lose any work but can work on multiple cases.

slorek: Worth noting there is an opportunity here to warn the user if there are no products added to the case, rather than showing the form. Something for another time maybe.

frankieroberto: @slorek we can't wipe the session from the `new` action, as clicking "Edit details" from the confirm view takes you back to the form, with the expectation that it's pre-filled with your current values... 

I like the idea of using `test_result_#{@investigation.id}` as the key, although that doesn’t completely solve the issue as if you attempt to create a second test result for the same investigation, it'll still contain any values from a previous un-completed test result for that case. But that feels like a reasonable compromise in the short term.  Longer term, I think we could either avoid using the session store for these flows, and save partial records to the database (which would allow users to continue a flow in a separate session), or generate a random ID at the start of each flow, and use that as a key within the session store.

frankieroberto: @slorek oh an yes, the form still rendering (but being unable to save) if there are no products on the case is a bit unhelpful! @edwardhorsford pointed this out too. I think we should create a backlog item to rethink this, if there's not something there already.

psd-web/app/controllers/investigations/test_results_controller.rb
frankieroberto: @slorek I've renamed `set_attachment` to `attach_file_from_params`, to (hopefully) better describe what it does. Mostly that method delegates to `load_file_attachments` from the `FileConcern`, which is even more confusing, but I didn't want to touch that concern just yet... 😬 

I'm not yet convinced it’s worth extracting the first four lines from the  `new`, `create_draft`, `confirm` and `create` methods. It'd have to be called something like `set_investigation_and_test_result_from_params_and_authorize` which is a bit of a mouthful. Open to suggestions though (and no doubt we'll end up refining these controllers again later as we add more update flows).

One downside I find with the `before_action` filters is that you can end up with the actual method containing no lines at all, which always confuses me momentarily, eg:

```ruby
before_action :set_thing
before_action :authorize

def show; end

def edit; end
```

psd-web/app/controllers/investigations/test_results_controller.rb
slorek: I think the Wicked Wizard has an advantage here, in that `new` immediately redirects to another action (`show`) with a unique URL slug for the page, so it solves a couple of problems (resetting the session at the start, and unique URLs). If we generate a UUID for the test result at the start it will mean the user loses their work if they navigate away, though. On balance I think having one per investigation is fine, as the only issues will be caused if the user has multiple tabs open.

frankieroberto: Agreed. Ultimately, I'd be in favour of saving to the database (or some other persistent store), at least for longer flows, as it'd mean we could offer "continue where you left off" type features. However it's a fair bit more work...

PR #746, opened 2020-06-26
Ensure only admin users can export cases

slorek:
## Description
This change ensures only users with the `psd_admin` role are able to download the cases export spreadsheet. This was always intended by conditionally showing the link, but the actual endpoint was not protected.

## Checklist:

### DIFF
--- a/psd-web/app/policies/user_policy.rb
+++ b/psd-web/app/policies/user_policy.rb
@@ -1,0 +1,5 @@
+class UserPolicy < ApplicationPolicy
+  def export_cases?
+    record.is_psd_admin?
+  end
+end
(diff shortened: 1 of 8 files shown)


psd-web/app/policies/user_policy.rb
```diff
@@ -0,0 +1,5 @@
+class UserPolicy < ApplicationPolicy
+  def export_cases?
+    record.is_psd_admin?
+  end
+end
```
frankieroberto, line 5: I wonder if this would make slightly more sense as part of `InvestigationPolicy`, as it seems to be _about_ the investigations, and pundit policies get passed the current user anyway?

eg something like:

```suggestion
class InvestigationPolicy < Struct.new(:user, :export)
  def export?
      user.is_psd_admin?
    end
end
```

and then in the controller it looks like you can pass the class instead of an instance:

```ruby
 authorize Post, :export?
````

(I had to read [the docs](https://github.com/varvet/pundit#policies) to realise this was an option)

slorek: Thanks, I was looking for this.

PR #748, opened 2020-06-26
Fix HTML in project description on case created activity

slorek:
https://trello.com/c/St4Ek73j/642-html-tag-p-showing-up-in-activity-logs-for-opss-projects

## Description
Fixes HTML `<p>` tags in the description of the audit activity for project created.

This was caused by what I believe to be unnecessary use of `decorate` on the investigation before it's created. None of the other creation flows use this, and the corresponding feature spec did not check the activity page, so it was not spotted.

I've improved the feature spec to check the activity page.

I have checked on production and the last project was created in April, before metadata was introduced. Therefore there are no affected cases on production, so no data migration is needed to fix anything.

## Checklist:

### DIFF
--- a/psd-web/app/controllers/investigations/project_controller.rb
+++ b/psd-web/app/controllers/investigations/project_controller.rb
@@ -59,7 +59,7 @@ private
   end
 
   def set_investigation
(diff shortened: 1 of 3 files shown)


nasirkhanbeis:
looks good to me!

PR #737, opened 2020-06-24
WIP: Sort supporting information first version

slawosz:

PR #749, opened 2020-06-26
demonstrate how to improve sort supporting information PR

StupidCodeFactory:
This is a PR to demonstrate improvements that could be made to [this](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/737/files) original PR

The trick here is to define an explicit single interface to handle the sorting.
I've also moved the sort method which returns an array as the public method allowing the remove the method delegation in the presenter and make the presenter leaner.


slorek:
Ideas have been incorporated into https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/737

PR #751, opened 2020-06-29
Convert test mode tests to rspec

frankieroberto:
This converts a couple more minitest tests to rspec specs.


psd-web/spec/factories/tests.rb
```diff
@@ -1,8 +1,8 @@
 FactoryBot.define do
   factory :test do
-    date_day    { date.day }
-    date_month  { date.month }
-    date_year   { date.year }
+    date_day    { date&.day }
+    date_month  { date&.month }
+    date_year   { date&.year }
```
frankieroberto, line 5: Required so that we can set the date to `nil` when testing validations.

psd-web/spec/models/test/test_result_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      it "is invalid and includes an error message", :aggregate_failures do
+        expect(test_result).not_to be_valid
+        expect(test_result.errors.messages[:result]).to include("Select result of the test")
+      end
+    end
+  end
+
+  describe "saving" do
+    let!(:product) { create(:product) }
+    let!(:investigation) { create(:allegation) }
```
frankieroberto, line 53: These use the `let!` method so that the activity entry for the investigation is created during setup, rather than as part of the `test_result.save!` transaction.

psd-web/spec/models/test/test_result_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    let(:test_result) { build(:test_result, investigation: investigation, product: product) }
+
+    it "creates an activity entry" do
+      expect { test_result.save! }.to change(Activity, :count).by(1)
+    end
```
frankieroberto, line 58: This is pretty loose (copied straight from the existing test), but I'm not sure it's worth improving it right now, as we'll hopefully soon move this to a `CreateTestResult` service object class.

PR #754, opened 2020-06-30
Set investigation Risk Level

scruti:
[Trello ticket](https://trello.com/c/LpaF7UMw/287-3-add-case-risk-level)
Allow users to set the risk level for the investigation through a radio button selection between a list of risk levels.

## Description
Users will be able to either choose one of the default risk levels or to set a customised risk level for the case.
Once set, the risk level can be changed or removed.

**The risk level will be displayed in the case Overview. In this case is not set, an will display a link allowing the user to set it**
Screenshot from 2020-07-02 23-41-27

**The risk level page will have the current risk level pre-selected**
Screenshot from 2020-07-02 23-41-45

**After selecting a risk level an submitted, the user will be sent to the case overview and will see a flash message informing them about the change**
Screenshot from 2020-07-10 11-49-11

**The overview page will display the current risk level and a link to change it**
Screenshot from 2020-07-10 11-49-16

**Users will be able to set their own risk level**
Screenshot from 2020-07-10 11-53-46

**Flash message will also be displayed when the risk level changes**
Screenshot from 2020-07-10 11-53-24

**Risk level can also be removed from the case**
Screenshot from 2020-07-10 12-02-48
Screenshot from 2020-07-10 12-03-11

**The audit activity log for the case will register every change on the risk level**
Screenshot from 2020-07-02 23-52-15

## Checklist:


nasirkhanbeis:
Hi @scruti  I'm testing this manually so please don't merge until I confirm, thanks.

scruti:
Thanks for the heads up @nasirkhanbeis

psd-web/app/models/audit_activity/investigation/update_risk_level.rb
```diff
@@ -0,0 +1,42 @@
+class AuditActivity::Investigation::UpdateRiskLevel < AuditActivity::Investigation::Base
+  I18N_SCOPE = "audit_activity.investigation.update_risk_level".freeze
+
+  def self.from(investigation, action:, source:)
```
slorek, line 4: Please change: The `from` interface is a legacy of the callbacks on `Investigation`, which we're actively trying to remove (once all the callbacks are gone). Just call `create!` in the service instead. See https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/models/audit_activity/investigation/update_owner.rb#L3 for an example.

psd-web/app/models/audit_activity/investigation/update_risk_level.rb
```diff
@@ -0,0 +1,42 @@
...
+    create!(
+      source: source,
+      investigation: investigation,
+      metadata: metadata
+    )
+  end
+
+  private_class_method def self.build_metadata(investigation, action)
+    {
+      risk_level: investigation.risk_level,
```
slorek, line 15: Suggestion It would be useful for audit purposes and other potential future use to store the old risk level as well as the new one, as @frankieroberto did on `AuditActivity::Test::TestResultUpdated`.

psd-web/app/models/audit_activity/investigation/update_risk_level.rb
```diff
@@ -0,0 +1,42 @@
...
+    I18n.t(".title.#{metadata['action']}", level: metadata["risk_level"].downcase, scope: I18N_SCOPE)
+  end
+
+private
+
+  def subtitle_slug
+    "Risk level changed"
+  end
+
+  def notify_relevant_users
```
slorek, line 30: Please change: Move this to the `ChangeCaseRiskLevel` service and override this method, as it relies upon ActiveRecord callbacks and we are trying to remove them. See https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/models/audit_activity/investigation/update_owner.rb#L32 for an example

psd-web/app/controllers/investigations/risk_level_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+      @risk_level_form = RiskLevelForm.new(params.require(:investigation).permit(:risk_level, :risk_level_other))
+      return render :show unless @risk_level_form.valid?
+
+      result = ChangeCaseRiskLevel.call(investigation: @investigation, user: current_user, risk_level: @risk_level_form.risk_level.presence)
+      if result.success?
+        flash[:success] = I18n.t(".success", scope: "investigations.risk_level", action: result.change_action, case_type: @investigation.case_type) if result.change_action.present?
+        redirect_to investigation_path(@investigation)
+      else
+        flash[:error] = I18n.t(".error", scope: "investigations.risk_level", action: result.change_action)
+        render :show
```
frankieroberto, line 21: Could we ✂️ this `else` condition? I think in theory the service call should always be successful, as you've already checked the form is valid?

psd-web/app/decorators/investigation_decorator.rb
```diff
@@ -38,13 +38,18 @@ def product_summary_list
     h.render "components/govuk_summary_list", rows: rows, classes: "govuk-summary-list--no-border"
   end
 
-  def display_risk_and_issues_list?
-    object.hazard_type.present? || object.non_compliant_reason.present?
+  def risk_level
+    object.risk_level.to_s
   end
```
frankieroberto, line 43: suggestion: rename this as something like `risk_level_description` and move the logic to return "Not set" if null here?

```suggestion
  def risk_level_description
    object.risk_level || "Not set"
  end
```

psd-web/app/models/audit_activity/investigation/update_risk_level.rb
```diff
@@ -0,0 +1,42 @@
...
+    entities_to_notify.each do |entity|
+      email = entity.is_a?(Team) ? entity.team_recipient_email : entity.email
+      NotifyMailer.case_risk_level_updated(
+        email: email,
+        name: entity.name,
+        investigation: investigation,
+        action: metadata["action"],
+        level: metadata["risk_level"]
+      ).deliver_later
```
frankieroberto, line 39: Please change: I think we've been moving away from doing this from the audit activity classes, and doing it within the service classes instead (eg see `ChangeCaseOwner` or `UpdateTestResult`).

psd-web/app/models/audit_activity/investigation/update_risk_level.rb
```diff
@@ -0,0 +1,42 @@
...
+  I18N_SCOPE = "audit_activity.investigation.update_risk_level".freeze
+
+  def self.from(investigation, action:, source:)
+    metadata = build_metadata(investigation, action)
+    create!(
+      source: source,
+      investigation: investigation,
+      metadata: metadata
+    )
+  end
```
frankieroberto, line 11: Please change: I think we're deprecating this method (eg see `AuditActivity::Test::TestResultUpdated` and `AuditActivity::Investigation::UpdateOwner`)

psd-web/app/models/audit_activity/investigation/update_risk_level.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  private_class_method def self.build_metadata(investigation, action)
+    {
+      risk_level: investigation.risk_level,
+      action: action
+    }
```
frankieroberto, line 17: In `AuditActivity::Test::TestResultUpdated` we ended up deciding to store both the old and previous values in the metadata, in a pattern like:

```ruby
{
  updates: {
   risk_level: ["Medium", "High"]
  }
}
```

...I wonder if that's worth adopting here?  It’d allow the activity entry to say something like "Risk level reduced from High to Medium". Would also mean you wouldn’t have to store the "action", I think?

psd-web/app/models/audit_activity/investigation/update_risk_level.rb
```diff
@@ -0,0 +1,42 @@
+class AuditActivity::Investigation::UpdateRiskLevel < AuditActivity::Investigation::Base
```
frankieroberto, line 1: Suggestion: rename to `AuditActivity::Investigation::RiskLevelUpdated` to better describe the event/action that has  happened.

psd-web/app/views/investigations/activities/investigation/_update_risk_level.html.erb
```diff
@@ -0,0 +1 @@
+<%= markdown simple_format(activity.body) %>
```
frankieroberto, line 1: I don't think this is doing anything, as the activity body isn't set? Also I think ideally we'd be generating the activity body from the metadata instead of storing HTML in the body.

psd-web/config/locales/en.yml
```diff
@@ -74,6 +74,14 @@ en:
         no_label: "No, this is business as usual"
         hint: "For example, cases relating to hand sanitiser or personal protective equipment (PPE)"
       success: "Coronavirus status updated on %{case_type}"
+    risk_level:
+      show:
+        title_set: "Set case risk level"
+        title_change: "Change case risk level"
```
frankieroberto, line 80: Suggestion: structure this so that it's a bit clearer that the two titles are alternatives to each other, eg:

```suggestion
      show:
        title:
          set: "Set case risk level"
          change: "Change case risk level"
```

psd-web/db/migrate/20200626142459_add_risk_level_to_investigations.rb
```diff
@@ -0,0 +1,8 @@
+class AddRiskLevelToInvestigations < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def change
+    add_column :investigations, :risk_level, :string
```
frankieroberto, line 5: I was _kinda_ expecting this to be an enum, with the "other" text value as a separate column. The main advantage would be that we could change the display values of the enum (eg "High Risk" to "High risk" 😄 ) without doing a data migration. However perhaps this is unlikely enough that it’s not worth worrying about.

psd-web/app/views/investigations/tabs/_overview.html.erb
```diff
@@ -23,11 +23,9 @@
...
     <% end %>
 
-    <% if @investigation.display_risk_and_issues_list? %>
-      <%= govuk_hr %>
-      <h2 class="govuk-heading-m"><%= "Risks and issues" %></h2>
-      <%= @investigation.risk_and_issues_list %>
-    <% end %>
+    <%= govuk_hr %>
+    <h2 class="govuk-heading-m"><%= "Risks and issues" %></h2>
+    <%= @investigation.risk_and_issues_list(policy(@investigation).update?(user: current_user)) %>
```
frankieroberto, line 28: I think you can drop passing the user in here, as Pundit does that automatically?

```suggestion
    <%= @investigation.risk_and_issues_list(policy(@investigation).update?) %>
```

psd-web/app/views/investigations/risk_level/show.html.erb
```diff
@@ -0,0 +1,42 @@
...
+                   classes: "govuk-!-width-one-third",
+                   label: { text: "Other risk level", classes: "govuk-visually-hidden" } %>
+      <% end %>
+
+      <% radio_items = Investigation::STANDARD_RISK_LEVELS.map do |level|
+          { text: level, value: level }
+         end
+         radio_items.push(text: "Other", value: "other", conditional: { html: other })
+         radio_items.push divider: "or"
+         radio_items.push(text: "Not set", value: "")
```
StupidCodeFactory, line 28: I would do this as a method on the form object. This clean up a bit this view.

psd-web/db/migrate/20200626142459_add_risk_level_to_investigations.rb
StupidCodeFactory: I agree with @frankieroberto

psd-web/spec/features/set_investigation_risk_level_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+    email = delivered_emails.last
+    expect(email.recipient).to eq creator_user.email
+    expect(email.personalization).to include(
+      name: creator_user.name,
+      verb_with_level: "set to medium risk",
+    )
+  end
+
+  scenario "Changing risk level for an investigation" do
+    investigation.update!(risk_level: "Medium Risk")
```
StupidCodeFactory, line 52: maybe user the service class?

psd-web/app/forms/risk_level_form.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+private
+
+  def set_other_from_risk_level?(risk_level)
+    risk_level.present? && !standard_risk_level?(risk_level)
+  end
+
+  def standard_risk_level?(risk_level)
+    Investigation::STANDARD_RISK_LEVELS.include? risk_level
```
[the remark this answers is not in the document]
scruti: Definitely. Was in my wish to list and ended forgetting to add it. Thanks for bringing it up.

psd-web/app/controllers/investigations/risk_level_controller.rb
slorek: If you do this, I would consider using `call!` instead of `call`, as failure will be suppressed otherwise. Although it should theoretically not fail, this will at least send an exception via Sentry and alert the user something went wrong.

psd-web/app/views/investigations/activities/investigation/_update_risk_level.html.erb
slorek: Yes, `body` and `title` are deprecated columns.

psd-web/db/migrate/20200626142459_add_risk_level_to_investigations.rb
slawosz: I think it will be more complicated to have 2 columns for risk level. @frankieroberto idea could be solved by storing risk levels as the hash (basically what enum does under the hood):

```ruby
RISK_LEVELS = {
  low: "Low risk level",
  high: "High risk level",
}
```
and so on.

psd-web/app/views/investigations/activities/investigation/_update_risk_level.html.erb
scruti: You're right, this is a copy&paste setting up the files when setting up the activity. Deleting it.

psd-web/app/views/investigations/risk_level/show.html.erb
scruti: Seems a good idea to build the array of options in the form object, but then I realised we have this bit
`radio_items.push(text: "Other", value: "other", conditional: { html: other })` loading html for the optional input. So stopped me from moving it there.

psd-web/app/models/audit_activity/investigation/update_risk_level.rb
scruti: What made me not follow the `create!` approach you linked is that the communication between the service class and the audit class becomes unnecessarily  bi-directional.
```ruby
  def create_audit_activity_for_case_owner_changed
    metadata = activity_class.build_metadata(owner, rationale) # enquiry metadata from the audit class

    activity_class.create!(
      source: UserSource.new(user: user),
      investigation: investigation,
      title: nil,
      body: nil,
      metadata: metadata # pass the data to the same class that generated it
    )
  end
```
The service class does not need to know about `metadata` existence at all, as its just internal information/behaviour belonging to the audit class, but we call it in order to be able to use "create!" method directly. I think this "generate metadata and create" knowledge belongs to an abstraction in the audit class, not in the service.

I  did override  `.from` as an encapsulation for this logic. And this method does not rely at the parent method (it completely overrides it). Maybe  moving it from `.from` to `.create_for` method can be clearer?
Going to rename it for now while I wait for your thoughts.

psd-web/config/locales/en.yml
```diff
@@ -91,6 +99,11 @@ en:
         body:
           true: "The case is related to the coronavirus outbreak."
           false: "The case is not related to the coronavirus outbreak."
+      update_risk_level:
+        title:
+          set: "Case risk level set to %{level}"
+          changed: "Case risk changed to %{level}"
```
edwardhorsford, line 105:
```suggestion
          changed: "Case risk level changed to %{level}"
```

psd-web/spec/features/set_investigation_risk_level_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+    click_button "Set risk level"
+
+    expect_confirmation_banner("Risk level changed on allegation")
+    expect_to_be_on_case_page(case_id: investigation.pretty_id)
+    expect(page).to have_summary_item(key: "Case risk level", value: "Mildly risky")
+
+    # Risk level change reflected in audit activity log
+    click_link "Activity"
+    expect_to_be_on_case_activity_page(case_id: investigation.pretty_id)
+    expect(page).to have_css("h3", text: "Case risk changed to mildly risky")
```
edwardhorsford, line 80:
```suggestion
    expect(page).to have_css("h3", text: "Case risk level changed to mildly risky")
```

psd-web/db/migrate/20200626142459_add_risk_level_to_investigations.rb
edwardhorsford: In our backlog we've got several more fields we want to expand so when you pick 'other' you get a freetext - so it would be good to have a consistent pattern. Months ago we discussed this with @slorek and I think there was week consensus on having the 'other' as an extra column.

psd-web/app/controllers/investigations/risk_level_controller.rb
```diff
@@ -0,0 +1,25 @@
+module Investigations
+  class RiskLevelController < ApplicationController
+    def show
```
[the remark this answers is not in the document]
scruti: Started doing this but found that the Coronavirus Related page was basically the same case as the Risk Level page and was implemented with `show`. So decided to stick with what was already implemented for this kind of controllers/pages.

psd-web/app/controllers/investigations/risk_level_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+      @investigation = Investigation.find_by!(pretty_id: params.require(:investigation_pretty_id)).decorate
```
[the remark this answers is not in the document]
slawosz: @StupidCodeFactory why? I am with @frankieroberto on this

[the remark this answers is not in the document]
scruti: The reason behind me favouring the `require` is that raises `ActionController::ParameterMissing` when the parameter is not there, and that is more idiomatic and revealing that an internal method raising a random exception for a call over `nil`.

I agree that calling directly the param is safe here, but strong params brigs us more than the protection against mass assignment.

psd-web/app/forms/risk_level_form.rb
```diff
@@ -0,0 +1,76 @@
...
+
+private
+
+  def set_attributes
+    attrs = {}
+    if original_level == OTHER
+      attrs[:risk_level] = original_level_other
+    elsif set_other_from_risk_level?
+      attrs[:risk_level_other] = original_level
+      attrs[:risk_level] = OTHER
```
slorek, line 31: Aren't these two conditions in the reverse order? The first will always set the `risk_level` if "other" is selected, so I don't think the second condition will ever be matched.

psd-web/app/forms/risk_level_form.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  attribute :risk_level, :string, default: nil
+  attribute :risk_level_other, :string, default: nil
+
+  validate :risk_level_allowed_validation
+  validate :risk_level_other_present_validation
+
+  def initialize(attributes)
+    @original_level = normalise(attributes[:risk_level])
+    @original_level_other = normalise(attributes[:risk_level_other])
```
slorek, line 18: Nitpick: I find the naming of `original_level` quite confusing. I thought it meant the risk level originally on the case, but actually it's the new form input. I'm not quite sure we need to juggle things in this way.

psd-web/app/forms/risk_level_form.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  def risk_level_other_present_validation
+    if original_level == OTHER && original_level_other.blank?
+      errors.add(:risk_level_other, :blank)
+      restore_originals
+    end
+  end
+
+  def set_other_from_risk_level?
```
slorek, line 52: I guess this is to catch edge cases where a user has somehow submitted a value that's not presented on the form?

psd-web/app/forms/risk_level_form.rb
```diff
@@ -0,0 +1,76 @@
...
+    original_level.present? && !matching_level_in_list(original_level, Investigation::STANDARD_RISK_LEVELS)
+  end
+
+  def restore_originals
+    self.risk_level = original_level
+    self.risk_level_other = original_level_other
+  end
+
+  def equal_levels?(first_level, second_level)
+    first_level.delete(" ").casecmp(
```
slorek, line 62: Nitpick: we probably want to strip all whitespace characters, not just spaces. `strip` is probably better as I'm not sure why we'd need to delete spaces in the middle of the string.

psd-web/app/models/audit_activity/investigation/risk_level_updated.rb
```diff
@@ -0,0 +1,38 @@
+class AuditActivity::Investigation::RiskLevelUpdated < AuditActivity::Investigation::Base
+  I18N_SCOPE = "audit_activity.investigation.risk_level_updated".freeze
+  SUBTITLE_SLUG = "Risk level changed".freeze
+
+  def self.from(*)
+    raise "Deprecated - use .create_for! instead"
+  end
+
+  def self.create_for!(investigation, action:, source:)
```
slorek, line 9: Nitpick: `action` is quite generic and already has other meanings - I had to look at other code to find what it meant (`set` or `change`).

psd-web/app/models/audit_activity/investigation/risk_level_updated.rb
```diff
@@ -0,0 +1,38 @@
...
+    I18n.t(".title.#{metadata['action']}", level: metadata["new_risk_level"]&.downcase, scope: I18N_SCOPE)
+  end
+
+private
+
+  def subtitle_slug
+    SUBTITLE_SLUG
+  end
+
+  # Do not send investigation_updated mail. This is handled by the ChangeCaseOwner service
```
slorek, line 36:
```suggestion
  # Do not send investigation_updated mail. This is handled by the ChangeCaseRiskLevel service
```

scruti: ouch copy paste :see_no_evil:

psd-web/db/migrate/20200626142459_add_risk_level_to_investigations.rb
slorek: I don't recall arriving at a consensus for this pattern, but I think it makes sense. i would also argue that we should not be storing strings like `"Severe risk"` in the database, but rather a key which is translated using `I18n` on presentation.

psd-web/app/forms/risk_level_form.rb
scruti: Good point, I added complexity through feedback/requirements interactions and ended not being very clear.
The data massaging is caused by:
- The need to transform the form input into the attributes.
- The fact that this happens in 2 ways: The form input from the user when submitting the form and the form input from the db values when loading the form.
- The need of re-render the form with errors when the "other" input value has a validation error. This is the main cause of having to store the original input so we can re-load the form with the original elements selected/filled.

Let me re-word it and add a couple comments so it becomes less confusing (I hope).

psd-web/app/forms/risk_level_form.rb
scruti: is actually to display the form with the "other" input field value from the "risk_level" in DB when displaying the form.
Going to add a comment, thanks for raising this.

psd-web/app/forms/risk_level_form.rb
scruti: I was thinking... Maybe someone introduces `high   Risk` or similar. But yeah going to make it simpler and switch ti `strip`

psd-web/app/models/audit_activity/investigation/update_risk_level.rb
slorek: I can understand your reasoning. I'd like to be consistent with the other services, so either call `build_metadata` from the service or implement `create_for` in the other services/activity models as well.

psd-web/db/migrate/20200626142459_add_risk_level_to_investigations.rb
scruti: Now that "It works", going to try to take this approach. I suspect it will make the `DB` <--> `Form` mapping simpler.

psd-web/app/models/audit_activity/investigation/update_risk_level.rb
scruti: Makes sense, going to follow up with a PR unifying the interface for the services.

psd-web/app/models/investigation.rb
```diff
@@ -24,6 +31,7 @@ class Investigation < ApplicationRecord
   validates :description, length: { maximum: 10_000 }
   validates :non_compliant_reason, length: { maximum: 10_000 }
   validates :hazard_description, length: { maximum: 10_000 }
+  validates :custom_risk_level, absence: true, if: -> { risk_level.present? }
```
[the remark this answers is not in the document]
scruti: Ended turning this into validating the presence when the risk level is "other" and the absence when the risk level is anything else.

psd-web/app/mailers/notify_mailer.rb
```diff
@@ -159,4 +160,25 @@ def team_deleted_from_case_email(message:, investigation:, team_deleted:, user_w
...
   end
+
+  def case_risk_level_updated(email:, name:, investigation:, update_verb:, level:)
+    set_template(TEMPLATES[:case_risk_level_updated])
+
+    verb_with_level = case update_verb
+                      when "set" then "set to #{level.downcase}"
+                      when "changed" then "changed to #{level.downcase}"
+                      when "removed" then "removed"
+                      end
```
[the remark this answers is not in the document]
scruti: brilliant!

PR #761, opened 2020-07-02
Add buttons to case bar

frankieroberto:
This adds some contextual buttons to the case bar. 

The content and destination of these buttons varies according to the status of the case, and the role of the user’s team in relation to it.

A future story will enhance buttons with javascript to a dropdown menu, avoiding the need to visit the second page.

### 1. User’s team is the owner of the case
 
Screenshot 2020-07-02 at 11 11 33

### 2. User’s team is not involved with the case at all

Screenshot 2020-07-02 at 11 11 43

### 3. User’s team is not involved with the case at all, but user is from OPSS and can send email alerts

Screenshot 2020-07-02 at 11 12 02

### 4. User’s team has been added to the case (but isn't the owner)

Screenshot 2020-07-02 at 11 14 33

### 5. User’s team has been added to the case (but isn't the owner), and user is from OPSS so can send email alerts

Screenshot 2020-07-02 at 11 14 27

### 1. User is not from OPSS

Screenshot 2020-07-02 at 11 32 16

### 2. User is from OPSS

Screenshot 2020-07-02 at 11 32 43

### DIFF
--- a/psd-web/app/assets/stylesheets/_button.scss
+++ b/psd-web/app/assets/stylesheets/_button.scss
@@ -25,5 +25,5 @@
 .app-button--case-bar-secondary:link,
 .app-button--case-bar-secondary:visited {
   color: $govuk-text-colour;
-  background-color: #dee0e2;
+  background-color: $old-mid-grey;
 }
--- a/psd-web/app/assets/stylesheets/_button.scss
+++ b/psd-web/app/assets/stylesheets/_button.scss
@@ -21,7 +21,7 @@
 }
 
 // An alternative secondary button, for use on the case bar (which has the
-// same backgorund colour as the regular secondary button)
+// same background colour as the regular secondary button)
 .app-button--case-bar-secondary:link,
 .app-button--case-bar-secondary:visited {
   color: $govuk-text-colour;
--- a/psd-web/app/assets/stylesheets/_button.scss
+++ b/psd-web/app/assets/stylesheets/_button.scss
@@ -19,3 +19,11 @@
   min-height: auto;
   justify-content: center;
 }
+
+// An alternative secondary button, for use on the case bar (which has the
+// same backgorund colour as the regular secondary button)
+.app-button--case-bar-secondary:link,
+.app-button--case-bar-secondary:visited {
+  color: $govuk-text-colour;
+  background-color: #dee0e2;
+}
--- a/psd-web/app/assets/stylesheets/main.scss
+++ b/psd-web/app/assets/stylesheets/main.scss
@@ -6,11 +6,14 @@ $govuk-assets-path: '~govuk-frontend/govuk/assets/';
 $hmcts-page-width: $govuk-page-width;
 $hmcts-gutter: $govuk-gutter;
 $hmcts-gutter-half: $govuk-gutter-half;
+$hmcts-images-path: '~@hmcts/frontend/assets/images/';
 @import "~@hmcts/frontend/objects/width-container";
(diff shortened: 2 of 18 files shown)


psd-web/spec/forms/investigation_actions_form_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+      it "is not valid" do
+        expect(form).not_to be_valid
+      end
+    end
+  end
+
+  describe "#actions" do
+    context "when the user is the owner of the investigation but not an OPSS user" do
+      let(:user) { create(:user, :activated) }
+      let(:investigation) { create(:allegation, owner_team: user.team, owner_user: nil) }
```
StupidCodeFactory, line 25: So you should use the trait `creator: user` and this call the `CreateCaseOwner` in the background and have a realistic setup. If you want to change the the case owner call the `ChangeCaseOwner`. That's how it is done everywhere.

psd-web/app/assets/stylesheets/_button.scss
```diff
@@ -19,3 +19,11 @@
   min-height: auto;
   justify-content: center;
 }
+
+// An alternative secondary button, for use on the case bar (which has the
+// same backgorund colour as the regular secondary button)
```
slorek, line 24: Typo:
```suggestion
// An alternative secondary button, for use on the case bar (which has the
// same background colour as the regular secondary button)
```

nasirkhanbeis:
Hi @frankieroberto  I'm looking at it now

psd-web/app/assets/stylesheets/main.scss
```diff
@@ -6,11 +6,14 @@ $govuk-assets-path: '~govuk-frontend/govuk/assets/';
 $hmcts-page-width: $govuk-page-width;
 $hmcts-gutter: $govuk-gutter;
 $hmcts-gutter-half: $govuk-gutter-half;
+$hmcts-images-path: '~@hmcts/frontend/assets/images/';
 @import "~@hmcts/frontend/objects/width-container";
 @import "~@hmcts/frontend/components/hmcts-badge/badge";
 @import "~@hmcts/frontend/components/hmcts-banner/banner";
 @import "~@hmcts/frontend/components/hmcts-identity-bar/identity-bar";
 @import "~@hmcts/frontend/components/hmcts-sub-navigation/sub-navigation";
+@import "~@hmcts/frontend/components/hmcts-menu/menu";
```
slorek, line 15: Suggestion: Order the dependencies alphabetically as currently.

psd-web/spec/forms/investigation_actions_form_spec.rb
slorek: Agreed, please change this.

scruti: I must be blind but I don't find the trait you mentioned :man_shrugging:

StupidCodeFactory: @scruti my apologies it is just a transient attribute https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/spec/factories/investigations.rb#L99

frankieroberto: Done, thanks.

frankieroberto:
@slorek the removal `click_link "Send email alert"` from the feature specs is because that link has now been removed, and you now have to use the button-link in the case bar for that. The link from the overview page was in an awkward location which I think was only meant to be temporary until we did the case bar...

PR #767, opened 2020-07-06
Remove unused data

slawosz:
Removing data related to the investigation, which is no longer needed.
`UserSource` where `type = Investigation` is replaced by `Collaboration` where type is `Collboration::CreatorTeam` and `Collaboration::CreatorUser`. `owner_id` and `owner_type` were replaced by `Collaboration::Access::User` and `Collaboration::Access::Team`.


psd-web/app/models/concerns/investigation_elasticsearch.rb
```diff
@@ -16,20 +16,19 @@ module InvestigationElasticsearch
     end
 
     def as_indexed_json(*)
-      as_json(
+      data = as_json(
```
StupidCodeFactory, line 19: I guess you could now remove `data` and return directly from `as_json`

PR #770, opened 2020-07-06
Add corrective action edit form

StupidCodeFactory:
# Description
Allows for updating a corrective action
I did not got along with the usual `form`, performing validation, along side a `service` object saving the object due to the DateConcern and FileConcern being incredibly hard and un-flexible to work with.

Some problems will be addressed in a later PR:
1. choose whether a file is attached or not is not pre-selected. At last demo is was pointed out it should not be, while the demo we made today it was pointed out that it should be.
2. when selecting "No"  to "Is there a file related to the corrective action?" it removes the existing attached documents but does not create an audit log. This will also be addressed in a later PR


nasirkhanbeis:
Found couple of blocking issues
1. Activity log is throwing error after I edit corrective actions
2. Edit form doesn't show file replace when I didn't upload file at first time and apparently throwing error when I submit edit form

psd-web/app/controllers/investigations/corrective_actions_controller.rb
```diff
@@ -1,7 +1,48 @@
...
   def show
-    @investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
+    @investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id]).decorate
+    authorize @investigation, :view_non_protected_details?
+    @corrective_action = @investigation.corrective_actions.find(params[:id]).decorate
+  end
+
+  def edit
+    @investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id]).decorate
     authorize @investigation, :view_non_protected_details?
```
frankieroberto, line 10: Please change: users need to have update permissions to be able to load this view.

```suggestion
    authorize @investigation, :update?
```

psd-web/app/controllers/investigations/corrective_actions_controller.rb
```diff
@@ -1,7 +1,48 @@
...
     @corrective_action = @investigation.corrective_actions.find(params[:id]).decorate
+    @file_blob = @corrective_action.documents_blobs.first || @corrective_action.documents_blobs.new
+  end
+
+  def update
+    @investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
+    authorize @investigation, :view_non_protected_details?
```
frankieroberto, line 17: Please change:

```suggestion
    authorize @investigation, :update?
```

psd-web/app/views/investigations/corrective_actions/show.html.erb
```diff
@@ -40,4 +40,8 @@
       <% end %>
     </div>
   <% end %>
+
+  <% if policy(@investigation).update? %>
+    <p class="govuk-body"><%= link_to "Edit corrective action", edit_investigation_action_path(@investigation, @corrective_action), class: "govuk-link" %></p>
+  <% end %>
```
frankieroberto, line 46: Please change:

This needs to be within a column div (eg the one above), to avoid it appearing outdented:

Screenshot 2020-07-16 at 12 36 02

nasirkhanbeis:
@yann found couple more minor issues
1. update attachment description doesn't get recorded in activity log
2. business field is not pre-filled in edit form, user can easily miss this and save corrective without business

nasirkhanbeis:
One more minor issue-  
Audit log doesn't show view corrective action in audit log if it doesn't have file attached
see screen shot  latest one shows link if there's a file but one below doesn't show 
image

nasirkhanbeis:
Just summarise, we have two main issues remaining to sort
1. Audit log changes: does always show attachment and attachment description even though I didn't changed
2. Corrective action from field, are there any files doesn't throw validation message when nothing selected and apparently on form submit it throws application error.

psd-web/app/services/update_corrective_action.rb
```diff
@@ -0,0 +1,136 @@
...
+
+    context.fail! if corrective_action.invalid?
+  end
+
+  def investigation
+    corrective_action.investigation
+  end
+
+  def send_notification_email(activity)
+    activity.entities_to_notify.each do |recipient|
```
slorek, line 120: Please change: We're working towards removing `entities_to_notify` from `Activity` and move them into the relevant service classes. This is because the recipients tend to be different and it relied on class overrides before. Please define the recipients in the service for clarity. See https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/services/change_case_owner.rb#L68 and https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/psd-web/app/services/add_team_to_case.rb#L28 for examples.

We would also then not need to pass `activity` in as an argument so we can make the `call` method clearer by calling `create_audit_activity_for_corrective_action_update` and `send_notification_email` sequentially in line with the other services.

psd-web/app/services/update_corrective_action.rb
```diff
@@ -0,0 +1,136 @@
...
+      email = recipient.is_a?(Team) ? recipient.team_recipient_email : recipient.email
```
slorek, line 121: Please change: We have now aliased `Team#email` for `Team#team_recipient_email ` so this line is not necessary.

StupidCodeFactory:
[(5) Let corrective actions be edited](https://trello.com/c/AE2mYEHl/501-5-let-corrective-actions-be-edited)

psd-web/app/services/update_corrective_action.rb
```diff
@@ -0,0 +1,134 @@
...
+  end
+
+  def entities_to_notify
+    return [] if user == investigation.owner_user
+    return [investigation.owner_user, investigation.owner_team] if investigation.owner_team.email?
+
+    User
+      .active
+      .where(team_id: investigation.teams_with_access.map(&:id))
+      .where.not(id: user.id)
```
frankieroberto, line 132: I think we want to only send notifications to active users in the team which owns the case (rather than any team with access) - and also to not send a notification to the user which did the action.

```suggestion
    investigation.owner_team.users.active - [user]
```

PR #776, opened 2020-07-08
Proposition to simplify risk level

StupidCodeFactory:
## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/psd-web/app/models/investigation.rb
+++ b/psd-web/app/models/investigation.rb
@@ -15,10 +15,11 @@ class Investigation < ApplicationRecord
   }
 
   enum risk_level: {
-    serious: 10,
-    high: 20,
-    medium: 30,
-    low: 40,
+    serious: "serious",
+    high: "high",
+    medium: "medium",
+    low: "low",
+    other: "other"
   }
 
   before_validation { trim_line_endings(:user_title, :description, :non_compliant_reason, :hazard_description) }
--- a/psd-web/app/models/investigation.rb
+++ b/psd-web/app/models/investigation.rb
@@ -32,7 +32,8 @@ class Investigation < ApplicationRecord
   validates :description, length: { maximum: 10_000 }
   validates :non_compliant_reason, length: { maximum: 10_000 }
   validates :hazard_description, length: { maximum: 10_000 }
-  validates :custom_risk_level, absence: true, if: -> { risk_level.present? }
+  validates :custom_risk_level, absence: true, if: -> { risk_level != "other" }
+  validates :custom_risk_level, presence: true, if: -> { risk_level == "other" }
 
   after_update :create_audit_activity_for_status,
                :create_audit_activity_for_visibility,
--- a/psd-web/app/models/investigation/risk_level_change.rb
+++ b/psd-web/app/models/investigation/risk_level_change.rb
@@ -3,7 +3,6 @@
 # risk level and custom risk level, providing information about what action
 # (if any) will be resulting from the change.
 class Investigation::RiskLevelChange
-
   def initialize(investigation)
     self.investigation = investigation
   end
--- a/psd-web/app/models/investigation/risk_level_change.rb
+++ b/psd-web/app/models/investigation/risk_level_change.rb
@@ -5,7 +5,7 @@
 class Investigation::RiskLevelChange
   attr_reader :current_level, :current_custom, :new_level, :new_custom
 
-  def initialize(investigation, new_level, new_custom)
+  def initialize(investigation, changes)
     @current_level = investigation.risk_level
     @current_custom = investigation.custom_risk_level
     @new_level = new_level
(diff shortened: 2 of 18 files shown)


psd-web/app/models/investigation.rb
```diff
@@ -15,10 +15,11 @@ class Investigation < ApplicationRecord
...
   enum risk_level: {
-    serious: 10,
-    high: 20,
-    medium: 30,
-    low: 40,
+    serious: "serious",
+    high: "high",
+    medium: "medium",
+    low: "low",
+    other: "other"
```
scruti, line 22: I'm not sure about switching to DB enum and including "other" too. 
From one of the discussions in the PR about how to store this, @slorek  mentioned how relying in the default int for enums and stet some ranges between different levels would allow us to have them ordered by severity and  future interject new risk levels between others.
(https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/754#discussion_r450761347) 

From what I see here maybe the order/severity will be taken from the enum definition here instead of from the DB values? I may have missed that behaviour on the enums.

About inclusion of "other" I thought we do not want to store it as a level since is just FE representation and we would have to exclude it when dealing with the risk levels beyond the frontend bit. But storing it as a risk level seems to help reducing the form complexity.

psd-web/app/models/investigation/risk_level_change.rb
```diff
@@ -3,13 +3,9 @@
...
 class Investigation::RiskLevelChange
-  attr_reader :current_level, :current_custom, :new_level, :new_custom
 
-  def initialize(investigation, new_level, new_custom)
-    @current_level = investigation.risk_level
-    @current_custom = investigation.custom_risk_level
-    @new_level = new_level
-    @new_custom = new_custom
+  def initialize(investigation)
+    self.investigation = investigation
```
scruti, line 8: Very nice refactor. Yep relying in Dirty values seems the way to go here :clap:

psd-web/db/migrate/20200708184946_change_risk_level_to_enum.rb
```diff
@@ -0,0 +1,21 @@
+class ChangeRiskLevelToEnum < ActiveRecord::Migration[5.2]
+  def up
+    safety_assured do
+      execute <<-SQL
+        CREATE TYPE risk_levels AS ENUM ('serious', 'high', 'medium', 'low', 'other');
```
scruti, line 5: Not sure about the upsides of this (besides being able to understand the DB value) against relying on the string -> integer mapping (https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/754#discussion_r450761347). 
In this case, when adding new levels we will need to do a DB migration right? While when using integers we could just add a new element to the enum in the Investigation.

psd-web/app/models/investigation.rb
StupidCodeFactory: that's a fair point, I think though it is reasonably easy to solve if when storing as a struct:
```SQL
SELECT * FROM investigations ORDER BY CASE risk_level WHEN serious THEN 1 WHEN high THEN 2 WHEN medium THEN 3 WHEN low THEN 4 WHEN other ELSE 5 END
```
It's also easy enough to filter no risk level. Also "other" instead of NULL effectively would be the same thing. "other" carries a semantic meaning whereas "NULL" doesn't.

psd-web/db/migrate/20200708184946_change_risk_level_to_enum.rb
StupidCodeFactory: that's true we'll need a migration, but a rather simple one and it's most likely going to happen rarely, in fat I'm pretty this not going to change for the next year

StupidCodeFactory: it's consistent with how we did `reported_reason` one

psd-web/spec/forms/risk_level_form_spec.rb
```diff
@@ -10,29 +10,22 @@
...
-
         it "keeps custom risk level " do
-          expect(form.custom_risk_level).to eq custom_risk_level
+          expect(form.attributes).to eq(custom_risk_level: custom_risk_level, risk_level: Investigation.risk_levels[:other])
         end
       end
 
       context "with a custom risk level that matches with space/capital variations a standard level" do
-        let(:custom_risk_level) { standard_level_text.upcase + "  " }
+        let(:custom_risk_level) { standard_level_key.upcase + "  " }
```
scruti, line 24: This will change the behaviour as now (from what I get from the form class and here) will only match the custom level with the risk level if the custom level is something like `Medium`, `High`... instead of matching the displayed text for the risk levels. 
As the form became simpler and there wasn't a requirement about the string matching between fields, I'm more than happy to go with this simpler version, but thought was worth to mention the behaviour change.

scruti:
@StupidCodeFactory thanks again for doing this. Made some small changes to get it ready to merge against the parent PR. Now everything should be ready to go! :raised_hands:

PR #784, opened 2020-07-16
Fixes and refactoring for adding team to a case

slorek:
Related to: https://trello.com/c/uxlRUoKo/559-5-let-case-owners-give-view-only-rights-to-teams

## Description
This is the first refactoring step needed before we can implement view-only collaborations on a case. It does not implement any new functionality yet but it is now possible to add this relatively easily.

* Refactored the code to use a form object rather than model validations in order to be consistent with our established patterns related to the use of form/service objects. This allowed for removal of the `include_message` attribute which was only used for form validation.
* Renamed the service to be consistent with naming of other services.
* Improved the interface to the service to be more object-oriented
* Removed the redundant `NotifyTeamAddedToCaseJob` which wasn't adding value and was not tested, to improve consistency with other service implementations.
* Fixed a bug where a team could be added to a case more than once.
* Fixed bugs which would allow multiple case owners.
* Added more test coverage.

In a follow-up PR I will address some issues with `EditInvestigationCollaboratorForm`.

## Checklist:


nasirkhanbeis:
I'm testing the bug fixes, please wait till I confirm.

PR #791, opened 2020-07-24
Allow case owner to give a team view only permissions on a case

slorek:
https://trello.com/c/uxlRUoKo/559-5-let-case-owners-give-view-only-rights-to-teams

## Description
This allows a member of a case's owner team to add teams to a case as collaborators with view only permissions, or change an existing collaborator to view only.

* Added radio options to allow the user to set the team's permission level when adding a team to a case.
* Renamed `EditInvestigationCollaboratorForm` to `EditCaseCollaboratorForm` for consistency.
* Refactored `EditCaseCollaboratorForm` to remove the `save` functionality which removed the team from a case - separated into a service class `RemoveTeamFromCase` for consistency and to facilitate changes of permission level.
* Added new view only option to change collaborator form.
* Added new `ChangeCasePermissionLevelForTeam` service.
* Wrapped some existing service database interactions in transactions for safety.
* Added additional validations on `EditCaseCollaboratorForm`.
* Added new Notify email template for when a team's permissions change on a case: https://www.notifications.service.gov.uk/services/b1dab0f5-4651-4af8-9f15-fa8143ff338d/templates/772f8eb6-2aa2-4ed3-92f2-78af24548303
* Converted all existing instances of `AuditActivity::Investigation::TeamAdded` and `AuditActivity::Investigation::TeamDeleted` to use metadata instead of title/body attributes via a data migration. I have tested the migration should work on both staging and production with existing data. This allows us to avoid adding redundant fallback code and displays the permission level on 'team added' activities, as well as a proper reference to the team which was added/removed.
* Fixed a race condition whereby the 'team added' notification email job could fail if the team had already been removed or had their permission changed when the email was delivered (due to the `collaboration` object being destroyed).
* Improved test coverage.

I have not tested what happens when a team views a case with view only permissions as I believe that has already been tested separately.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis:
looks good to me, just a minor validation issue. see below it is showing error message twice in error summary
image
steps to replicate, 
Go to teams page, click add team, select team and submit the form.
see error summary

psd-web/app/models/collaboration/access/owner.rb
```diff
@@ -1,6 +1,9 @@
 class Collaboration < ApplicationRecord
   class Access < Collaboration
-    class Owner < Edit; end
-    attribute :include_message, :boolean, default: false
+    class Owner < Edit
+      def self.can_be_changed?
+        false
+      end
```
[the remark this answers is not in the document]
slorek: Yes I struggled with the naming as well.

psd-web/app/services/add_team_to_case.rb
```diff
@@ -1,22 +1,26 @@
 class AddTeamToCase
   include Interactor
 
-  delegate :collaboration, :user, :investigation, :team, :message, to: :context
+  delegate :collaboration, :user, :investigation, :team, :collaboration_class, :message, to: :context
```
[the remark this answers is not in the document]
slorek: I did start with this approach, but I ended up with a lot of hard-coded "readonly" and "edit" strings/symbols all over the place and ended up doing some refactoring to arrive at this (and the use of I18n to remove much of it which felt a little awkward as well). I felt it was a more object-oriented approach but happy to change it back if you don't agree.

psd-web/app/views/collaborators/edit.html.erb
```diff
@@ -33,20 +33,29 @@
         },
         items: [
           {
-            id: "permission_level",
-            text: "Edit full case",
-            value: EditInvestigationCollaboratorForm::PERMISSION_LEVEL_EDIT,
+            id: "permission_level_view",
```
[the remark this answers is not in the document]
slorek: 👍 didn't know about this. Thanks.

psd-web/db/migrate/20200724135733_convert_activity_metadata_for_teams_on_case.rb
```diff
@@ -0,0 +1,27 @@
+class ConvertActivityMetadataForTeamsOnCase < ActiveRecord::Migration[5.2]
+  def change
+    AuditActivity::Investigation::TeamAdded.where.not(title: nil).each do |activity|
+      team_name = activity.attributes["title"].match(/(.*) added to/).captures.first.strip
+      team = Team.find_by!(name: team_name)
```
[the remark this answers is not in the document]
slorek: I have tested it should work ;)

psd-web/app/forms/add_team_to_case_form.rb
```diff
@@ -3,14 +3,27 @@ class AddTeamToCaseForm
...
   validates :message, presence: true, if: :include_message
   validates :include_message, inclusion: { in: [true, false] }
 
   def team
     @team ||= Team.find(team_id)
   end
+
+  def collaboration_class
+    Collaboration::Access.class_from_human_name(permission_level)
+  end
```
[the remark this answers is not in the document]
slorek: See above comment. This was the original implementation.

psd-web/app/services/add_team_to_case.rb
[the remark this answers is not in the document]
slorek: Also applies to `ChangeCasePermissionLevelForTeam`.

psd-web/app/models/collaboration/access/owner.rb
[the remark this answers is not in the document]
slorek: For context I added this because I ended up with a lot of places which were I had to whitelist the classes which were edit/read only. And this allows us to add other permission levels in future.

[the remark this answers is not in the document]
StupidCodeFactory: you'd have to have other extra checks the other way around, whether `Owner` is or is not a subclass of `Edit`. 
I think you could rejig the class hierarchy or more the association on the investigation to avoid this method.

psd-web/app/services/add_team_to_case.rb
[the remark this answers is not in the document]
StupidCodeFactory: I think it is cleaner to inject/delegate the class rather than having the service knows about the internals of the each of the collaborations class. I think we should keep it that way.

psd-web/app/services/remove_team_from_case.rb
```diff
@@ -0,0 +1,60 @@
+class RemoveTeamFromCase
+  include Interactor
+
+  delegate :collaboration, :user, :message, to: :context
+
+  def call
+    context.fail!(error: "No collaboration supplied") unless collaboration.is_a?(Collaboration::Access)
+    context.fail!(error: "Collaboration type cannot be removed") unless collaboration.can_be_changed?
```
StupidCodeFactory, line 8: Suggestion: So it seems the `#can_be_changed?` are only really used inside the services class. It feels to me that these are the one that should be the oracle for that data rather than each of the classes. That would remove all this methods `#can_be_changed?`

psd-web/app/views/collaborators/edit.html.erb
```diff
@@ -33,20 +33,29 @@
...
+            text: "View full case",
+            value: Collaboration::Access::ReadOnly.model_name.human,
```
StupidCodeFactory, line 38: ❤️

psd-web/spec/forms/edit_case_collaborator_form_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+      it "returns true" do
+        expect(form.delete?).to be true
+      end
+    end
+
+    context "when permission_level is a valid access level" do
+      let(:permission_level) { Collaboration::Access::Edit.model_name.human }
+
+      it "returns false" do
+        expect(form.delete?).to be false
```
StupidCodeFactory, line 108:
```suggestion
        expect(form).not_to be_delete
```

psd-web/spec/forms/edit_case_collaborator_form_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+      it "returns the existing level" do
+        expect(form.permission_level).to eq(Collaboration::Access::Edit.model_name.human)
+      end
+    end
+  end
+
+  describe "#delete?" do
+    context "when permission_level is delete" do
+      it "returns true" do
+        expect(form.delete?).to be true
```
StupidCodeFactory, line 100:
```suggestion
        expect(form).to be_delete
```

psd-web/spec/models/collaboration/access/edit_spec.rb
```diff
@@ -0,0 +1,9 @@
+require "rails_helper"
+
+RSpec.describe Collaboration::Access::Edit do
+  describe ".can_be_changed?" do
+    it "returns true" do
+      expect(described_class.can_be_changed?).to be true
```
StupidCodeFactory, line 6:
```suggestion
      expect(described_class).to be_can_be_changed
```

psd-web/spec/models/collaboration/access/owner_team_spec.rb
```diff
@@ -7,6 +7,12 @@
   let(:team) { investigation.owner_team }
   let(:user) { investigation.owner_user }
 
+  describe ".can_be_changed?" do
+    it "returns false" do
+      expect(described_class.can_be_changed?).to be false
```
StupidCodeFactory, line 12:
```suggestion
      expect(described_class).not_to be_can_be_changed
```

psd-web/spec/models/collaboration/access/owner_user_spec.rb
```diff
@@ -0,0 +1,9 @@
+require "rails_helper"
+
+RSpec.describe Collaboration::Access::OwnerUser do
+  describe ".can_be_changed?" do
+    it "returns false" do
+      expect(described_class.can_be_changed?).to be false
```
StupidCodeFactory, line 6:
```suggestion
      expect(described_class).not_to be_can_be_changed
```

psd-web/spec/models/collaboration/access/read_only_spec.rb
```diff
@@ -0,0 +1,9 @@
+require "rails_helper"
+
+RSpec.describe Collaboration::Access::ReadOnly do
+  describe ".can_be_changed?" do
+    it "returns true" do
+      expect(described_class.can_be_changed?).to be true
```
StupidCodeFactory, line 6:
```suggestion
      expect(described_class).to be_can_be_changed
```

frankieroberto: "be_can_be_changed" is a bit of a mouthful! Perhaps we could use `changeable?` or `editable?` as the method name, making the matcher `be_changeable`/`be_editable`?

psd-web/app/services/remove_team_from_case.rb
slorek: There are also [three](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/791/files#diff-0774823e544b2266d7823806a775a8ddR42) [instances](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/791/files#diff-0774823e544b2266d7823806a775a8ddR51) [where](https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/791/files#diff-03c53b56d420e530333ca2c6dbb3da23R25) we need to retrieve records scoped to only edit/read-only to prevent removal or duplicate case owners. I do think it makes sense for the classes to determine if they may be transformed to/from an `Owner`. As @frankieroberto observed this all stems from the inheritance and positioning in the `Collaboration::Access` namespace of the `Owner`. While this does make sense in respect of the level of access an owner has, it is causing us to have to specifically exclude the owner due to business logic which dictates a case must always have exactly one owner.

slorek:
Following discussion we have agreed to rename `can_be_changed` to `changeable` and I have now implemented the other suggested changes and the error bug fix.

PR #803, opened 2020-08-05
Render custom 404 page if the route is invalid

frankieroberto:
This makes sure we render our 404 page rather than a 500 page if there's a routing error.

### DIFF
--- a/psd-web/app/controllers/application_controller.rb
+++ b/psd-web/app/controllers/application_controller.rb
@@ -18,6 +18,7 @@ class ApplicationController < ActionController::Base
 
   helper_method :nav_items, :secondary_nav_items, :previous_search_params, :current_user
 
+  rescue_from ActionController::RoutingError, with: :render_404_page
   rescue_from Wicked::Wizard::InvalidStepError, with: :render_404_page
 
   def set_current_user


psd-web/app/controllers/application_controller.rb
```diff
@@ -18,6 +18,7 @@ class ApplicationController < ActionController::Base
 
   helper_method :nav_items, :secondary_nav_items, :previous_search_params, :current_user
 
+  rescue_from ActionController::RoutingError, with: :render_404_page
```
StupidCodeFactory, line 21: I'm not certain this is what would solve https://sentry.io/organizations/beis/issues/1637006347/?project=1329381&referrer=slack

In this case I think a 500 is what ought to be renderer. 
Also I'm not convinced we went to render as 404 all the routing error, this could hide other potential problems

frankieroberto: Rails renders 404 pages if there's a routing error by default, but that's getting blocked in our app because it’s still trying to fetch the current user for some reason. This probably isn't the best fix, but I didn't want to spend too long looking into it initially (also wanted to test this in a review app, as it behaves different locally).

PR #804, opened 2020-08-06
add a custom ActiveModel::Type::GovUKDate

StupidCodeFactory:
## add a custom ActiveModel::Type::GovUKDate

This makes it simpler to add the govuk date coming from the govuk multiple input form.

now in form object just declare the type in at the attribute level:

```ruby
class AFormWithAGovUKDate
  attribute :date_receive, :govuk_date
end
```

### DIFF
--- a/psd-web/app/models/date_parser.rb
+++ b/psd-web/app/models/date_parser.rb
@@ -11,26 +11,24 @@ class DateParser
   def date
     return nil if @date.nil?
     return @date if @date.is_a?(Date)
-    return nil if @date[:year].blank? && @date[:month].blank? && @date[:day].blank?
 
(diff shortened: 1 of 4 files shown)


psd-web/app/models/date_parser.rb
```diff
@@ -11,27 +11,25 @@ def initialize(date)
...
-        struct_from_hash
-      end
-    else
+    date_values = @date.values_at(:year, :month, :day).map do |date_part|
+      Integer(date_part)
+    rescue StandardError
+      nil
+    end
+
+    return nil if date_values.any?(&:blank?)
```
frankieroberto, line 21:
```suggestion
    return nil if date_values.all?(&:blank?)
```

PR #819, opened 2020-08-11
Fix https://sentry.io/organizations/beis/issues/1828948260/?environme…

StupidCodeFactory:
## CHANGES: 

Added the missing #own! method and extract the changeable interface to a module
to be included in all the changeable/swappable to owner teams.


StupidCodeFactory:
[Allow users to change case owner to a team which currently has read-only access](https://trello.com/c/txGMZCGz/692-allow-users-to-change-case-owner-to-a-team-which-currently-has-read-only-access)

PR #823, opened 2020-08-12
Add risk level badges

StupidCodeFactory:
**CHANGES:**
1. badge on top of investigation page
2. badge on case card
Screenshot 2020-08-12 at 16 42 40
Screenshot 2020-08-12 at 16 42 31
Screenshot 2020-08-12 at 16 42 21
Screenshot 2020-08-12 at 17 06 21


StupidCodeFactory:
[(2)New tags to be added `serious risk and high risk` for cases](https://trello.com/c/rQYHdiVo/665-2new-tags-to-be-added-serious-risk-and-high-risk-for-cases)

app/decorators/investigation_decorator.rb
```diff
@@ -129,6 +129,12 @@ def owner
     object.owner&.decorate
   end
 
+  def risk_level
+    return if self[:risk_level].blank?
+
+    @risk_level ||= ActiveSupport::StringInquirer.new(self[:risk_level])
+  end
```
frankieroberto, line 136: Any reason not to use the `enum` methods from the investigation (`investigation.serious?` etc)? Not sure this needs a decorator...

app/helpers/investigations/display_text_helper.rb
```diff
@@ -152,6 +152,6 @@ def report_summary_list(investigation)
   end
 
   def has_badges?(investigation)
-    investigation.is_private? || investigation.is_closed? || investigation.coronavirus_related?
+    investigation.is_private? || investigation.is_closed? || investigation.coronavirus_related? || investigation.risk_level&.serious? || investigation.risk_level&.high?
```
frankieroberto, line 155: Suggest using the methods provided by the investigation enum:

```suggestion
    investigation.is_private? || investigation.is_closed? || investigation.coronavirus_related? || investigation.serious? || investigation.high?
```

(It might be worth considering adding [a prefix](https://api.rubyonrails.org/classes/ActiveRecord/Enum.html) to the method names to make this slightly more readable, eg `risk_level_serious?`

app/views/investigations/_case_badges.html.erb
```diff
@@ -1,18 +1,59 @@
...
-    classes: "app-badge--coronavirus #{local_assigns[:classes]}"
-  %>
+<% badges = [] %>
+<% badges << capture do %>
+  <% if investigation.coronavirus_related? %>
+    <%= render "components/hmcts_badge",
+               text: t(:coronavirus_related, scope: "case.badges"),
+               classes: "app-badge--coronavirus #{local_assigns[:classes]}"
+    %>
+  <% end %>
```
frankieroberto, line 8: Suggestion, move the `badges << capture do` inside the condition, to avoid having to do `badges.compact!` at the end:

```suggestion
<% if investigation.coronavirus_related? %>
  <% badges << capture do %>
    <%= render "components/hmcts_badge",
               text: t(:coronavirus_related, scope: "case.badges"),
               classes: "app-badge--coronavirus #{local_assigns[:classes]}"
    %>
  <% end %>
```

app/views/investigations/_case_badges.html.erb
```diff
@@ -1,18 +1,59 @@
...
+        <% grouped_badge.each do |badge| %>
+          <%= badge %>
+        <% end %>
+      </div>
+    </div>
+  <% end %>
+<% else %>
+  <% badges.each do |badge| %>
+    <%= badge %>
+  <% end %>
```
frankieroberto, line 58: Is this needed? I just tried editing the HTML so that 3 tags were all together, and it seemed to wrap onto two line ok anyway?

We could avoid this wrapping mid-tag using either non-breaking spaces within the tag, or `white-space: nowrap` in the CSS:

Screenshot 2020-08-12 at 18 31 17

StupidCodeFactory: I'm not sure I quite understand this but will give it a go. I'm pretty CSS dumb :)

StupidCodeFactory: ah ok gotcha!

PR #825, opened 2020-08-12
Add prompt to update case risk level (if different from risk assessment)

frankieroberto:
This adds an additional question, when adding a risk assessment to a case which contains a different risk level from the existing case risk level, asking whether or not the user wants to set the case risk level to match the risk assessment.

Note that this happens after the risk assessment has already been saved to the case, and so if the user doesn’t answer, then the risk assessment is still saved. If the user does choose to update the case risk level, then an additional activity entry is added to the case, and an additional email notification will be sent to the case owner.

## Screenshot

update-case-risk-level


app/controllers/investigations/risk_assessments_controller.rb
```diff
@@ -29,7 +29,11 @@ def create
           })
         )
 
-        redirect_to investigation_risk_assessment_path(@investigation, result.risk_assessment)
+        if (result.risk_assessment.risk_level == @investigation.risk_level) && (result.risk_assessment.custom_risk_level == @investigation.custom_risk_level)
```
StupidCodeFactory, line 32: Suggestion: wrap this check in a method in the model and use of early return to avoid extra/nested if/else for consistency with the patterns used in other controllers:

```ruby
  # ...
  def risk_level_unchaged?
    risk_level_was == risk_level && custom_risk_level_was == custom_risk_level
  end
  # ...
end

class RiskAssessmentsController < ApplicationController
  def update
   #...
    return redirect_to investigation_risk_assessment_path(@investigation, result.risk_assessment) if @investigation.risk_level_unchaged?
    redirect_to investigation_risk_assessment_update_case_risk_level_path(@investigation, result.risk_assessment)
   end
end
```

app/controllers/investigations/update_case_risk_level_from_risk_assessment_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+        )
+      end
+
+      redirect_to investigation_risk_assessment_path(@investigation, @risk_assessment)
+    end
+
+  private
+
+    def form_params
+      params.fetch(:update_risk_level_from_risk_assessment_form, {}).permit(:update_case_risk_level_to_match_investigation)
```
StupidCodeFactory, line 38: I'm a bit confused as to why the params wouldn't contain the `update_risk_level_from_risk_assessment_form` key? is it because of the `render :show` if the form is invalid and you try to avoid the case where the user would refresh the page on the render show with error?

frankieroberto: It’s because if you submit the form with no radio selected, then the key is missing from the params as the browser doesn’t submit anything. I did consider a dummy hidden field under the same key to avoid this, but the fetch method seemed a bit cleaner and was what was suggested here: https://guides.rubyonrails.org/action_controller_overview.html#more-examples - open to other ideas though.

StupidCodeFactory: My preference would have been hidden field as that's what rails does by default, we've added a few hidden fields in other parts to solve this problem. I'm not strongly opposed to this approach either. 
It's probably something to keep in mind (add default hidden fields that is) when we will wrap the govuk helpers for our form builders!

PR #827, opened 2020-08-13
Add content to set case risk level page

frankieroberto:
This adds some additional content to the "Set case risk level" page to remind users about any risk assessments that have been added to the case, and their risk levels.

### When there are no risk assessments added to the case

Screenshot 2020-08-14 at 11 56 21

### When there is one risk assessment added to the case

Screenshot 2020-08-14 at 11 57 23

### When there is more than one risk assessments added to the case

Screenshot 2020-08-14 at 11 58 02

https://trello.com/c/wOCDXNxu/605-2-add-risk-assessment-content-to-case-risk-level-page

### DIFF
--- a/app/decorators/investigation_decorator.rb
+++ b/app/decorators/investigation_decorator.rb
@@ -1,6 +1,6 @@
 class InvestigationDecorator < ApplicationDecorator
   delegate_all
-  decorates_associations :complainant, :documents_attachments, :creator_user, :owner_user, :owner_team, :activities
+  decorates_associations :complainant, :documents_attachments, :creator_user, :owner_user, :owner_team, :activities, :risk_assessments
 
   PRODUCT_DISPLAY_LIMIT = 6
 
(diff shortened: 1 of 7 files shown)


app/decorators/investigation_decorator.rb
```diff
@@ -24,6 +24,10 @@ def display_product_summary_list?
     products.any?
   end
 
+  def risk_assessment_risk_levels
+    risk_assessments.collect(&:risk_level_description).uniq
```
StupidCodeFactory, line 28:
```suggestion
    risk_assessments.distinct(:risk_level_description)
```

StupidCodeFactory: not sure this adds adds a lot of value so feel free to discard

frankieroberto: It doesn't work, as `risk_level_description` is a method on the decorator, not the model.

PR #850, opened 2020-08-21
replace corrective action summary

StupidCodeFactory:
https://trello.com/c/7yOKSipX/278-5-replace-summary-of-action-on-corrective-action-with-type-of-action-taken

## CHANGES: 

This PR substitute free text "summary" field on corrective action for structured a list of predetermined actions. This list of actions is displayed as radio box.
This also has addressed the trading standard flow! 

## Deployment notes:
since the `summary` field has been removed for `action` and `other_action`. We will need to reindex all the investigation upon deploy.
```shell
$ cf7 run-task psd-pr-839 --command "export \$(./infrastructure/env/get-env-from-vcap.sh) && bin/rails runner  \"Investigation.__elasticsearch__.import(force: true)\""
```

## Testing note
Check that all  previous corrective actions summaries have all been migrated successfully. 
  @nasirkhanbeis I will re-open the PR against a branch reflecting master. You will need to *first* create corrective action and then check later on that this is still displayed as the title after merging this branch in and then also verify that new and old corrective actions are searchable.


nasirkhanbeis:
migration seems looking good, corrective actions are not searchable yet. 
Steps to replicate
1. Add a corrective action ( e.g. import rejected at border)
2. Now to cases list page and search by keyword "import rejected at border" 
returns 0 results, expected to return case against above corrective action added

README.md
```diff
@@ -1,4 +1,4 @@
-# Product safety database
+# Product safety database (trigger build)
```
frankieroberto, line 1: ✂️

app/controllers/investigations/record_corrective_actions_controller.rb
```diff
@@ -31,6 +31,7 @@ def create
     respond_to do |format|
       update_attachment
       if corrective_action_saved?
+        @investigation.__elasticsearch__.index_document
```
frankieroberto, line 34: Curious why this is needed here...

app/helpers/application_form_builder.rb
```diff
@@ -105,7 +108,7 @@ def govuk_file_upload(attribute, label:, hint: nil)
     )
   end
 
-  def govuk_select(attribute, label:, items:, hint: nil)
+  def govuk_select(attribute, label:, label_classes: nil, items:, hint: nil)
```
frankieroberto, line 111: Default to empty string to avoid the need to call `to_s` later?

```suggestion
  def govuk_select(attribute, label:, label_classes: "", items:, hint: nil)
```

app/models/corrective_action.rb
```diff
@@ -28,12 +40,17 @@ class CorrectiveAction < ApplicationRecord
   validates :duration, inclusion: { in: DURATION_TYPES }, if: -> { duration.present? }
   validates :geographic_scope, presence: true
   validates :geographic_scope, inclusion: { in: Rails.application.config.corrective_action_constants["geographic_scope"] }, if: -> { geographic_scope.present? }
+  validates :action, inclusion: { in: actions.keys }
+  validates :other_action, presence: true, length: { maximum: 10_000 }, if: -> { action == "other" }
```
frankieroberto, line 44: Use the enum method here?

```suggestion
  validates :other_action, presence: true, length: { maximum: 10_000 }, if: -> { other? }
```

app/models/corrective_action.rb
```diff
@@ -28,12 +40,17 @@ class CorrectiveAction < ApplicationRecord
...
 
-  validates :summary, length: { maximum: 10_000 }
   validates :details, length: { maximum: 50_000 }
 
   after_create :create_audit_activity
 
+  def action_label
+    self.class.actions[action]
+  end
```
frankieroberto, line 52: This kinda feels like it shouldn't be part of the model? Could you lookup the label string from the locale file from wherever calls this method instead?

app/models/corrective_action.rb
frankieroberto, line 44: Might also be good to have a validation to ensure that the `other_action` attribute is `nil` if an action is specified?

```ruby
validates :other_action, absence: true, unless: -> { other? }
```

app/controllers/investigations/record_corrective_actions_controller.rb
StupidCodeFactory: This is needed since we investigation does not change it is not reindexed but we need to for a reindex so that the new corrective action's value are picked up by the search

app/models/corrective_action.rb
StupidCodeFactory: I've put it in the model because this is used below to by the reindex module https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/850/files#diff-5d0bacf0b49eb27a259dda1a2f862fd9R58

app/controllers/investigations/record_corrective_actions_controller.rb
slorek: If it is necessary it would be worth adding a comment here to explain why. I would have probably tried to remove it if I refactored this code.

app/controllers/investigations/ts_investigations_controller.rb
```diff
@@ -87,6 +87,7 @@ def new
 
   def create
     if records_saved?
+      @investigation.__elasticsearch__.index_document
```
slorek, line 90: Suggestion: Add a comment here to explain why it's necessary.

app/decorators/audit_activity/corrective_action/update_decorator.rb
```diff
@@ -1,4 +1,8 @@
 class AuditActivity::CorrectiveAction::UpdateDecorator < AuditActivity::CorrectiveAction::BaseDecorator
+  def new_action
+    CorrectiveAction.actions[metadata.dig("updates", "action", 1)]
```
slorek, line 3: Suggestion: It could be worth storing the value in the metadata as well, as the `CorrectiveAction.actions` list could change over time.

app/models/corrective_action.rb
```diff
@@ -15,8 +15,20 @@ class CorrectiveAction < ApplicationRecord
 
   date_attribute :date_decided
 
-  before_validation { trim_line_endings(:summary, :details) }
-  validates :summary, presence: { message: "Enter a summary of the corrective action" }
+  enum action: {
```
slorek, line 18: Suggestion: I don't think we need to translate all the values in the model. This can be done on the view layer when necessary.

app/helpers/form_options_helper.rb
```diff
@@ -16,4 +16,18 @@ def product_categories
   def corrective_action_geographic_scope
     Rails.application.config.corrective_action_constants["geographic_scope"]
   end
+
+  def corrective_action_summary_radio_items(form)
+    CorrectiveAction.actions.map do |value, text|
+      item = { text: text, value: value }
```
slorek, line 22: Suggestion: Translate the action here rather than retrieve from the model.

app/services/update_corrective_action.rb
```diff
@@ -20,13 +20,16 @@ def call
       actvity = create_audit_activity_for_corrective_action_updated!(@previous_attachment)
 
       send_notification_email(actvity)
+
+      investigation.reload.__elasticsearch__.index_document
```
slorek, line 24: Suggestion: Add an explanatory comment here.

slorek (approved):
Code looks good. Just a couple of minor suggestions. I haven't tested the review app.

app/models/corrective_action.rb
StupidCodeFactory: it is translated there because I need the translated labels to index the translated values for each key

tobyhughes:
Could we amend these fields

image
image

to align with the 'Under which legislation?' formatting?

tobyhughes:
image

think the naming convention in the table should be:

`{truncated action}: {product name}`
or if 'Other'
`{free text}: {product name}`

PR #853, opened 2020-08-25
Add product categories to export

StupidCodeFactory:
On the case overview page we solve this by collecting the categories of each product added to the case, combining this with the investigation category, and uniquing the result.

Suggest we do similar for export - but output a comma separated string rather than separate lines for each.

Screenshot 2020-08-28 at 12 32 10


StupidCodeFactory:
[Product categories do not get exported in categories when exported](https://trello.com/c/ZunEBpc9/491-product-categories-do-not-get-exported-in-categories-when-exported)

slorek (approved):

PR #872, opened 2020-09-08
Fix error when displaying activity for risk assessment updated

slorek:
https://trello.com/c/Zx9Q2pIm/724-edit-risk-assessment-leave-activity-tab-in-error-state

## Description
This fixes [an exception](https://sentry.io/organizations/beis/issues/1858324804/?environment=staging&project=1329381&query=is%3Aunresolved) when displaying the audit activity for an updated risk assessment where the assessed on date had changed. The view was expecting a `Date` object but the model was returning a `String`.

Generally we are missing unit test coverage on this activity model which needs further backfilling. I've added unit tests for the relevant method, and modified the feature spec to test changing the date as well.

Steps to reproduce:

1. Add a risk assessment to a case
1. Edit the risk assessment and change the assessed on date
1. Go to the case activity page

### DIFF
--- a/app/models/audit_activity/risk_assessment/risk_assessment_updated.rb
+++ b/app/models/audit_activity/risk_assessment/risk_assessment_updated.rb
@@ -43,7 +43,10 @@ class AuditActivity::RiskAssessment::RiskAssessmentUpdated < AuditActivity::Base
   end
 
   def new_assessed_on
-    updates["assessed_on"]&.second
+    date = updates["assessed_on"]&.second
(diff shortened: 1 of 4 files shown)


StupidCodeFactory (approved):
:shipit:

nasirkhanbeis (approved):
nice work, looks good as expected.
Tested all possible scenarios works fine.
Btw there's seems a pre-existing bug which is separate issue so will raise a ticket for that.

PR #888, opened 2020-09-14
Remove unused zip gem

frankieroberto:
This doesn't appear to be used anywhere (searching the codebase for `zip`).


StupidCodeFactory (approved):
We might unzip zips soon :)

PR #893, opened 2020-09-16
Upgrade node from 10.20 to 12.18

frankieroberto:
12.18.x is the current "long term support" (LTS) version of Node.js, so we should upgrade to it.

10.20 will no longer be maintained from mid-2021.


frankieroberto:
Closing as this version of Node isn't yet supported by the CloudFoundry ruby buildpacks.

PR #894, opened 2020-09-16
Run brakeman static analysis for security warnings

frankieroberto:
This adds a new Github Action to run the [Brakeman](https://brakemanscanner.org) static analysis tool, which checks for potential security issues within the Rails code.

For the sake of speed, the gem is installed directly, rather than as part of the application bundle, as no other gems are required. This will also ensure we always run the latest version.

There was one false-positive caused by `Business.find(params[:id]).decorate`, but this was easy to resolve by postponing the call to `decorate` to later within the controller (which is best practice anyway).


StupidCodeFactory (approved):

slorek (approved):
Nice, that's a good result! Thanks for doing this.

PR #903, opened 2020-09-23
Fix sentry deprecation warning

StupidCodeFactory:
removes annoying deprecation warning about configuration change around the breadcrumbs logger.


slorek (approved):
Thanks!

frankieroberto (approved):

PR #906, opened 2020-09-24
Zip file report

StupidCodeFactory:
## Description
print a table with the blob id, average file size and list of extensions

## Run the report 

In the console: 

```irb
Rails.application.load_tasks
Rake::Task["data:zip_file_report"].invoke
```

Screenshot 2020-09-28 at 12 27 31

Screenshot 2020-09-28 at 12 43 33


nasirkhanbeis:
@StupidCodeFactory  I think we need to extract cases too which as .zip files?

slorek (approved):

PR #907, opened 2020-09-24
Convert record email correspondence flow to 1 page form

frankieroberto:
This converts the "email correspondence" flow from being a 2-page flow with a confirmation screen, to a 1-page form with no confirmation screen. This is to make it more consistent with the other flows (eg risk assessments), and to enable an editing feature to be implemented more easily (ie without sending users back through a multi-page flow).

## Implementation notes

I've removed Wicked Wizard, and simplified the controller and the routes to just using the `:new`, `:create` and `:show` actions. A form class has been introduced, and the view has been switched to use the new `ApplicationFormBuilder` helpers. The actual saving of the email model has been moved into a service class.

One tricky bit is making sure that any uploaded file is saved, even if there’s a validation error on the form. This seems to be the default behaviour in Rails 5, although there’s no built-in way within Rails 5 to associate the saved file with the re-rendered form, but in Rails 6 this will no longer happen automatically. 

The current code uses the `FileConcern` for this, and saves the reference to the file in the session data. However this has caused us some bugs in the past, and causes issues with edge cases like filling in the same form in two different browser tabs.

So instead I've introduced some code in the controller to implement this, by calling `ActiveStorage::Blob.create_after_upload!` with the file, and then associating the `signed_id` with the form using a separate attribute / hidden field (re-using the same attribute didn’t work as you then have 2 fields with the same name, which seemed to confuse Capybara at least). This is part of the controller for now, as it didn't feel it should be part of either the form class or the service object class - but we can refactor this into some shared behaviour.

The saving of the audit entry and the sending of the notification emails isn't yet part of the service class, as I didn't want to refactor too much at once, but I've left a `TODO` note, and hope to do this as part of a follow-up PR.

### Before (2 pages)

record-email-before

### After (1 page)

record-email-single-form


app/controllers/investigations/record_emails_controller.rb
```diff
@@ -1,75 +1,94 @@
...
+
+      @email_correspondence_form.existing_email_attachment_id = @email_correspondence_form.email_attachment.signed_id
+
+    # Lookup existing file using signed ID if present (for example if previously uploaded
+    # but validation failed)
+    elsif @email_correspondence_form.email_attachment.blank? && @email_correspondence_form.existing_email_attachment_id.present?
+
+      @email_correspondence_form.email_attachment = ActiveStorage::Blob.find_signed(@email_correspondence_form.existing_email_attachment_id)
+
+    end
```
frankieroberto, line 54: This is duplicated (for the two different attributes) for now. But it'd be nice to extract this somehow. It's also part of the controller as it doesn't feel like it belongs to the form model or the service class. Suggestions welcome!

app/assets/javascripts/investigations/attachment_description.js
```diff
@@ -1,7 +1,7 @@
 import $ from 'jquery'
 
 $(document).ready(() => {
-  const attachmentFileInput = $('#attachment-file-input')
+  const attachmentFileInput = $('#attachment-file-input, #email_attachment')
```
StupidCodeFactory, line 4: should we change this to  more a more generic class `js-attachmentFileInput` ?

app/controllers/investigations/record_emails_controller.rb
```diff
@@ -1,75 +1,94 @@
...
-    get_attachment_metadata_params(:email_file)
-        .merge(common_file_metadata)
-        .merge(
-          description: "Original email as a file"
-        )
-  end
+    # Upload file to S3 if present, so that it can persist even if validation fails
+    if @email_correspondence_form.email_file.present?
+
+      @email_correspondence_form.email_file = ActiveStorage::Blob.create_after_upload!(
```
StupidCodeFactory, line 21: I'd move this in a lazy method inside the form. that would find or upload and setting `existing_email_file_id`

app/controllers/investigations/record_emails_controller.rb
StupidCodeFactory, line 54: same for this if/else block

app/services/add_email_to_case.rb
```diff
@@ -0,0 +1,31 @@
...
+      email_address: email_address,
+      email_direction: email_direction,
+      overview: overview,
+      details: details,
+      email_subject: email_subject,
+      email_file: email_file,
+      email_attachment: email_attachment
+    )
+
+    # Update email attachment description
```
StupidCodeFactory, line 22: suggestion: remove the comment and create a method `#update_email_attachment_description`

StupidCodeFactory (changes requested):
Looks good, just a couple thing:
1. There was a 500 when entering record email date in the future
2. I think it is fine to handle the file upload, like I described in my comments in the form object and un-clutter the controller with a couple `if/else` but up to you if you want to address  that or not

frankieroberto:
@StupidCodeFactory good catch! Didn't spot the validation as it was on the super class. Validation added to the form object in 1fa670f2ba06cc582b71787cbed786cbaf21eda5

app/assets/javascripts/investigations/attachment_description.js
frankieroberto: Yeah we should, but I didn't want to break/change the existing code for now.  I'd also quite like to refactor it to avoid jQuery as we barely use it.

app/controllers/investigations/record_emails_controller.rb
frankieroberto: It felt a bit unexpected to me that a form object would save a file or look up a record using an ID. But maybe it's ok. Think I'll leave it for now until the pattern becomes clearer as other forms need to do this.

StupidCodeFactory (approved):

app/controllers/investigations/record_emails_controller.rb
slorek: At the moment the form object's role is to sanitise user input - I don't think it would be too unexpected for it to ensure the uploaded file was processed (and therefore sent for AV scanning), especially as we need that to happen to display the form back to the user. But I'm not bothered whether it's there or in the controller. However, this `create` method is rather long and so it could do with breaking up a bit. We have a number of forms where the file upload is not persisted when there are errors, so a re-usable/repeatable pattern would be ideal.

app/controllers/investigations/record_emails_controller.rb
slorek: Perhaps something like a `FileConcern` :trollface:

app/services/add_email_to_case.rb
```diff
@@ -0,0 +1,36 @@
...
+      email_subject: email_subject,
+      email_file: email_file,
+      email_attachment: email_attachment
+    )
+
+    if email.email_attachment.attached? && attachment_description.present?
+      update_attachment_description!
+    end
+
+    # TODO: refactor into this class
```
slorek, line 26: I think we should do this as part of the change (or immediately after) as we have done previously. These audit activity models need to be migrated one by one so we can remove `User.current`.

slorek (commented):
Nice work, thanks 👏 

There are a couple of comments on things I would like us to do now, but if you prefer to do these as an immediate follow-up PR we can merge this.

app/services/add_email_to_case.rb
frankieroberto: Agreed. Will do this as a follow-up PR.

nasirkhanbeis (approved):
looks good as expected, also it doesn't clear the attachment on other field validation errors.

PR #915, opened 2020-09-30
Bugfix: Allow people to enter dates with leading zeros

frankieroberto:


slorek (approved):
Not your fault; but I wonder if converting to `Integer` is unnecessary and likely to lead to other unexpected results.

PR #920, opened 2020-10-05
merge email correspondence

StupidCodeFactory:
## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -148,7 +148,7 @@ Rails.application.routes.draw do
 
     resources :correspondence, controller: "investigations/correspondence_routing", only: %i[new create]
     resources :emails, controller: "investigations/record_emails", only: %i[new create]
-    resources :phone_calls, controller: "investigations/record_phone_calls", only: %i[show new create update]
+    resources :phone_calls, controller: "investigations/record_phone_calls", only: %i[new create], path: "record-phone-calls"
     resources :meetings, controller: "investigations/record_meetings", only: %i[show new create update]
     resources :alerts, controller: "investigations/alerts", only: %i[show new create update]
 
--- a/spec/features/record_phone_call_activity_spec.rb
+++ b/spec/features/record_phone_call_activity_spec.rb
@@ -38,7 +38,9 @@ RSpec.feature "Adding a record phone call activity to a case", :with_stubbed_ela
 
     # Test date validation
     fill_in "Day", with: "333"
-    click_button "Continue"
+    save_and_open_page
+
+    # byebug
 
     expect(page).to have_error_messages
     expect(page).to have_error_summary "Correspondence date must include a month and year"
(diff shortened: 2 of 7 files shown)


StupidCodeFactory:
[Change the "record phone call" flow to be one page.](https://trello.com/c/Ss1WA3P0/746-change-the-record-phone-call-flow-to-be-one-page)

config/routes.rb
```diff
@@ -148,7 +148,7 @@
 
     resources :correspondence, controller: "investigations/correspondence_routing", only: %i[new create]
     resources :emails, controller: "investigations/record_emails", only: %i[new create]
-    resources :phone_calls, controller: "investigations/record_phone_calls", only: %i[show new create update]
+    resources :phone_calls, controller: "investigations/record_phone_calls", only: %i[new create], path: "record-phone-calls"
```
frankieroberto, line 151: I think we could match the path to the one used by the `show` action (defined separately above), with the intention that we'll soon be able to merge those controllers, and add edit/update actions?

```suggestion
    resources :phone_calls, controller: "investigations/record_phone_calls", only: %i[new create], path: "phone-calls"
```

spec/features/record_phone_call_activity_spec.rb
```diff
@@ -31,53 +31,33 @@
     expect_to_be_on_record_phone_call_page
 
     # Test required fields
-    click_button "Continue"
+    click_on "Add phone call"
```
frankieroberto, line 34: I think it's helpful to assert that it's a button rather than a link?

```suggestion
    click_button "Add phone call"
```

spec/features/record_phone_call_activity_spec.rb
```diff
@@ -31,53 +31,33 @@
...
-    click_button "Continue"
-
-    expect_to_be_on_record_phone_call_details_page
-
-    within_fieldset "Details" do
-      expect(page).to have_css("a", text: File.basename(file))
-    end
-
-    click_button "Continue"
+    attach_file "phone_call_correspondence_form[transcript]", file
```
frankieroberto, line 55: You can use the label text as a selector here:

```suggestion
    attach_file "Upload a file", file
```

frankieroberto (changes requested):
Looks good.

I think we should have at least a super basic unit test for the new service and form classes, personally - if only because they're likely to get more complicated in future and it's helpful to have the test skeleton in place.

config/environments/test.rb
```diff
@@ -6,7 +6,7 @@
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
-  config.cache_classes = true
+  config.cache_classes = false
```
StupidCodeFactory, line 9: this avoids problems with spring and classes being loaded twice in the class hierarchy.

frankieroberto (approved):
Looks good. 👍 

Just one required change to one of the tests.

app/forms/email_correspondence_form.rb
```diff
@@ -24,22 +24,16 @@ class EmailCorrespondenceForm
   validates :correspondence_date,
             presence: true,
             real_date: true,
-            complete_date: true
+            complete_date: true,
+            in_future: true
```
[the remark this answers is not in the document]
StupidCodeFactory: yes agreed that's confusing!

nasirkhanbeis (approved):
looks good as expected!

PR #923, opened 2020-10-06
Ability to edit emails

frankieroberto:
This allows "email correspondence" to be edited.

The form largely looks the same as when creating the email correspondence in the first place, except for the file upload fields, which, if there is currently a file attached, present 3 radio options, allowing the user to keep the existing file, remove it, or replace it.

## Screenshots

Screenshot 2020-10-28 at 16 09 37

If you select the option to replace the file, but don't select a different file, you get a validation error:

Screenshot 2020-10-28 at 16 10 42

If you select a replacement file, and then submit the form, but there's a validation error elsewhere, then the file upload field is replaced by a link to the file you just uploaded, but you still have the option to revert to the "current" file, or to remove both files:

Screenshot 2020-10-28 at 16 11 52


nasirkhanbeis:
Had a quick look yesterday and found a bug

nasirkhanbeis:
wondering when user want to change the uploaded file then there's no way unless they select remove and edit again 
Suggestion: we still need choose file input next to uploaded file
image

frankieroberto:
@nasirkhanbeis the bug should have been fixed - thanks for spotting that.

You're right that there's currently no way to select a different file if after selecting a replacement file but getting a validation error. Whilst we could keep the file input below the currently selected file, I was worried that this would introduce confusion, and there would also be no way, once selecting a 3rd file, to change your mind and stick with the one you uploaded before. Another option I considered was adding a 4th radio button, like this:

```
Replace or remove file

( ) Keep current file
( ) Remove file
( ) Use replacement file: filename.pdf
( ) Upload a different file
```

...but that might be even more confusing!

It's a tricky edge case! I suggest that we keep this as-is for now, as it seems _fairly_ unlikely that a user would want to swap the file again after getting a validation error, and if that happens, they can always re-edit the record anyway.

nasirkhanbeis:
> Had a quick look yesterday and found a bug, submit edit email form with empty body and subject field with Remove email file option, it is throwing 422 error

Hi @frankieroberto  Found another bug around this behaviour . steps to replicate

1. email with no body or subject, click edit
2. select remove radio and submit, throws expected error 
3. Now select radio keep and submit , it still throws error.

frankieroberto:
@nasirkhanbeis thanks, that bug has been fixed now too.

nasirkhanbeis:
looks all good now, just few minor issues around activity log otherwise all good. Re: design I'm still not convinced from user point however Toby's call. 
No activity log on below field changes 
Email attachment with no attachment description
Email Summary 
Email subject 
Email body

app/controllers/investigations/record_emails_controller.rb
```diff
@@ -15,61 +17,68 @@ def create
...
 
-    # Lookup existing file using signed ID if present (for example if previously uploaded
-    # but validation failed)
-    elsif @email_correspondence_form.existing_email_file_id.present? && @email_correspondence_form.email_file.blank?
+      @email = @investigation.emails.new
+      @email_correspondence_form.cache_files!
 
-      @email_correspondence_form.email_file = ActiveStorage::Blob.find_signed(@email_correspondence_form.existing_email_file_id)
+      @email_correspondence_form.email_file_action = "keep" if @email_correspondence_form.email_file.present?
+      @email_correspondence_form.email_attachment_action = "keep" if @email_correspondence_form.email_attachment.present?
```
slorek, line 41: Suggestion: this logic should probably be encapsulated by the form object.

app/models/audit_activity/correspondence/email_updated.rb
```diff
@@ -0,0 +1,78 @@
+class AuditActivity::Correspondence::EmailUpdated < AuditActivity::Base
```
slorek, line 1: Please change: I'm a bit confused why we aren't inheriting from `AuditActivity::Correspondence::Base`. Especially as that defines the `correspondence` relationship. Perhaps we need an explanatory comment.

app/models/audit_activity/correspondence/email_updated.rb
```diff
@@ -0,0 +1,78 @@
...
+    updated_values["email_attachment_filename"]
+  end
+
+  def new_attachment_description
+    updated_values["attachment_description"]
+  end
+
+  def self.from(*)
+    raise "Deprecated - use UpdateEmail.call instead"
+  end
```
slorek, line 40: I think this is not needed if we are inheriting from `AuditActivity::Base` as we have not inherited a `from` method.

app/models/audit_activity/correspondence/email_updated.rb
slorek, line 1: Please change: I think we need to override the private `notify_relevant_users` method as we have in other activity models. Otherwise the `after_save :notify_relevant_users` callback inherited from `Activity` will generate extra emails which have an empty body. Our tests should also be checking that no extra emails are being generated. It will be nice when we can finally eliminate this mess!

app/services/update_email.rb
```diff
@@ -0,0 +1,119 @@
...
+        investigation.pretty_id,
+        recipient.name,
+        email_address,
+        "#{user_source.show(recipient)} edited an email on the #{investigation.case_type}.",
+        "Email edited for #{investigation.case_type.upcase_first}"
+      ).deliver_later
+    end
+  end
+
+  def entities_to_notify
```
slorek, line 113: Please change: Use `EntitiesToNotify#email_recipients_for_case_owner` from #987

spec/features/edit_email_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+    expect_to_be_on_email_page(case_id: investigation.pretty_id, email_id: email.id)
+
+    # Page should show updated details
+    expect(page).to have_summary_item(key: "Date of email", value: "2 April 2020")
+
+    expect(page).to have_h1("Note received from manufacturer")
+
+    expect(page).to have_text("email_attachment.txt")
+    expect(page).to have_text("Manufacturer safety document")
+  end
```
slorek, line 134: Suggestion: We could also lightly test that emails have been sent here.

spec/services/update_email_spec.rb
```diff
@@ -0,0 +1,172 @@
+require "rails_helper"
+
+RSpec.describe UpdateEmail, :with_stubbed_elasticsearch, :with_stubbed_mailer, :with_stubbed_antivirus, :with_test_queue_adapter do
+  let!(:investigation) { create(:allegation) }
+  let(:product) { create(:product_washing_machine) }
+
+  let(:team) { create(:team, name: "Team 2") }
+  let(:user) { create(:user, name: "User 2", team: team) }
```
slorek, line 8: Suggestion: I guess this should have an `activated` trait.

spec/services/update_email_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+
+        it "notifies the team", :aggregate_failures do
+          expect { result }.to have_enqueued_mail(NotifyMailer, :investigation_updated).with(a_hash_including(args: [
+            investigation.pretty_id,
+            investigation.owner_user.name,
+            investigation.owner_user.email,
+            "User 2 (Team 2) edited an email on the allegation.",
+            "Email edited for Allegation"
+          ]))
+        end
```
slorek, line 141: Please change: Please use the `it_behaves_like "a service which notifies the case owner"` shared examples from #987

slorek (changes requested):
Nice! Just a few minor comments.

app/models/audit_activity/correspondence/email_updated.rb
slorek, line 1: Please change: define a `title` method for the activity which is displayed on the timeline.
Screenshot 2020-10-30 at 16 32 30

slorek (changes requested):
Bug: Attachment description is always showing in the activity summarywhen it has not changed (remained blank).
Screenshot 2020-10-30 at 16 32 30

app/services/update_email.rb
```diff
@@ -0,0 +1,119 @@
...
+      (attachment_description == @previous_attachment_description.to_s)
+  end
+
+  def update_attachment_description!
+    context.email.email_attachment.blob.metadata[:description] = attachment_description
+    context.email.email_attachment.blob.save!
+  end
+
+  def create_audit_activity
+    AuditActivity::Correspondence::EmailUpdated.create!(
```
slorek, line 75: Please change: i think for consistency with other audit activity models we would want to add the `correspondence` object here. Otherwise there's no reference to the actual email that was changed.

slorek (changes requested):

app/services/update_email.rb
frankieroberto: Done, thanks. I forgot about the foreign key to correspondence (the activity model has foreign keys to some models but not others... 🤷 )

app/models/audit_activity/correspondence/email_updated.rb
frankieroberto: Done, thanks.

app/services/update_email.rb
```diff
@@ -0,0 +1,113 @@
...
+
+private
+
+  def no_changes?
+    !email.changed? &&
+      (email_file_action == "keep" || (
+        email_file_action.nil? && !email.email_file.attached?)) &&
+      (email_attachment_action == "keep" || (email_attachment_action.nil? && !email.email_attachment.attached?)) &&
+      (attachment_description == @previous_attachment_description.to_s)
+  end
```
StupidCodeFactory, line 69: multiple interactor would make this a lot easier to understand.

nasirkhanbeis:
looks like there's  no way to change uploaded file in below edit scenario
1. given email which has subject and body fields entered with no attachment, click edit
2. clear subject and body, upload file, change date to invalid
3. Submit form, throws error on the date field which is expected, however returned page doesn't give option to remove, change uploaded file. 
Expected to show choose input field so if needed user can re-upload new file. 
image

nasirkhanbeis:
Also when nothing changed, it still records empty log
Click edit email, don't change anyting but just click update
go to activity log, it shows empty log

frankieroberto:
> looks like there's no way to change uploaded file in below edit scenario

This should be fixed, thanks.

nasirkhanbeis (approved):
Looks all good as expected.

slorek (approved):

PR #924, opened 2020-10-06
Fix test result file not being reset and conflicting with other forms

slorek:
https://trello.com/c/xqSWK6kR/677-test-results-form-doesnt-clear-the-attachment-and-description-on-subsequent-test-result-form

## Description
Fixes the add test result form having a file attached already from a previous session.

Also fixes the risk assessment files apparently not being virus checked in the Trading Standards flow.

When adding a test result, the form will save the attachment filename in the session under a generic `file` key name (set in the controller). This key is used by several other forms as well, so if you abandon filling in one of these forms their attachment could appear to be pre-filled on the test result form.

Normally the session key is reset to `nil` on the `new` action. Because of the different way the test result controller was constructed, this doesn't happen, so the form would appear to be pre-filled with the last attachment used on any form sharing the `file` attachment name.

To mitigate this, I have set a unique name for the file so it is not shared with any other forms. I have also ensured it is cleared on `create` so it will be blank if the user adds a second test result.

This does not resolve another issue where if the user has multiple tabs open on the same form, it will create a conflict in the input values. This is because the form sets its value in the session using a generic key. This is another separate issue.

In the Trading Standards Flow it re-uses the same form view template, which caused some issues with the new params namespace, so I had to make some minor changes to the file upload handling. The `FileConcern#load_file_attachments` implementation needs reworking to do this properly, and we have already decided to rework this separately. Also, we really should create a service object to create test results as we do with risk assessments, but this work would be out of scope for this minor fix. So I've followed a similar pattern to the risk assessment files and not used the `load_file_attachments` method from `FileConcern` as it is much easier this way.

## Checklist:


StupidCodeFactory (approved):

spec/features/add_test_results_spec.rb
```diff
@@ -93,6 +98,22 @@ def fill_in_test_result_submit_form(legislation:, date:, test_result:, file:)
     expect(page).to have_css("h1", text: "Confirm test result details")
   end
 
+  def expect_test_result_form_to_be_blank
+    expect(page).to have_field("Against which legislation?", with: nil)
+    expect(page.find_field("Day").text).to be_blank
+    expect(page.find_field("Month").text).to be_blank
+    expect(page.find_field("Year").text).to be_blank
```
frankieroberto, line 105: Any reason not to use the same capybara method here as above?

```suggestion
    expect(page).to have_field("Day", with: "")
    expect(page).to have_field("Month", with: "")
    expect(page).to have_field("Year", with: "")
```

frankieroberto (approved):
Nice!

Although, the sooner we can kill `FileConcern` the better 😄

spec/features/add_test_results_spec.rb
slorek: Yes - it doesn’t work 😂

nasirkhanbeis (approved):
Tested on review app, looks good as expected.

PR #931, opened 2020-10-09
Convert minitest tests to feature specs for business locations

frankieroberto:
This converts a minitest controller test for creating, editing and deleting business locations into 3 separate feature specs.

Also fixes a bug I found, which is that the "Town or city" field wasn't being saved as the wrong (old?) attribute name of `:locality` was whitelisted, instead of `:city`.


spec/features/add_location_to_a_business_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    fill_in "Town or city", with: "New Town"
+    fill_in "County", with: "Townshire"
+    fill_in "Postcode", with: "AB1 2CDE"
+    select "United Kingdom", from: "Country"
+
+    click_button "Save"
+
+    expect_to_be_on_business_page(business_id: business.id, business_name: "Acme Ltd")
+
+    expect(page).to have_text("Headquarters")
```
StupidCodeFactory, line 29: Minor: we've been using have_css|have_selector everywhere I'd keep using have_css it's a bit more descriptive as to where/how the piece of text is displayed on the page.

StupidCodeFactory (approved):

spec/features/add_location_to_a_business_spec.rb
frankieroberto: Ha, well I've been avoiding using css selectors as I don't like tying the feature specs to the CSS implementation – same reason for using the label text as the field selectors rather than the ID / name attributes. There are pros and cons to it though.

I think ideally I'd like to do something like:

```ruby
within_section("Locations") do 
  expect(page).to have_text("Headquarters")
end
```

...where `within_section` scopes the query to a `<section>` element which has the heading text given.  Might try and create that as a helper at some point.

PR #939, opened 2020-10-12
fix  zip analysis script

StupidCodeFactory:
## Description

Fixes the extraction of the case associated with a zip file when the zip file is attached to a product.


StupidCodeFactory:
[Investigate ZIP files in prod](https://trello.com/c/XHZLyfwd/727-investigate-zip-files-in-prod)

frankieroberto (approved):
There's a typo in the (existing) variable name: `attachement`

PR #943, opened 2020-10-12
Remove model tests

frankieroberto:
This removes the final two minitest tests, which were tests on the models.

The functionality was largely already covered by the rspec suite, although I added some quick tests for activity generation on adding and removing a business, as that hasn't been converted to a service class yet.

As these are the last minitest tests remaining, the final test removes all the minitest set up and support.


test/models/investigation_test.rb
```diff
@@ -1,82 +0,0 @@
...
-    owner.own!(@new_investigation)
-
-    assert(policy(@new_investigation).view_non_protected_details?(user: owner))
-  end
-
-  test "past owners should be computed" do
-    user = users(:southampton_bob)
-    ChangeCaseOwner.call!(investigation: @investigation, owner: user, user: user)
-    assert_includes @investigation.past_owners, user
-  end
```
slorek, line 63: Looks like we need to implement a unit test for `past_owners`.

test/models/investigation_test.rb
```diff
@@ -1,82 +0,0 @@
...
-    assert_equal(policy(@new_investigation).view_non_protected_details?(user: user), true)
-  end
-
-  test "visible to owner organisation" do
-    create_new_private_case(users(:southampton))
-    owner = users(:southampton_steve)
-    owner.own!(@new_investigation)
-
-    assert(policy(@new_investigation).view_non_protected_details?(user: owner))
-  end
```
slorek, line 57: Looks like we need a unit test on `view_non_protected_details?` in investigation_policy_spec.rb.

test/models/investigation_test.rb
```diff
@@ -1,82 +0,0 @@
...
-  test "people out of current owner's team should not be able to change the case owner" do
-    investigation = create_new_case(users(:southampton))
-    assert_not policy(investigation).change_owner_or_status?(user: users(:luton))
-  end
-
-  test "people not in the team that is the case owner should not be able to change the case owner" do
-    investigation = create_new_case
-    User.current.team.own!(investigation)
-    assert_not policy(investigation).change_owner_or_status?(user: users(:luton))
-  end
```
slorek, line 74: Looks like we need a unit test on `change_owner_or_status?` in investigation_policy_spec.rb.

test/models/investigation_test.rb
```diff
@@ -1,82 +0,0 @@
...
-    end
-  end
-
-  test "should create an activity when status is updated on investigation" do
-    @investigation = create_new_case
-    assert_difference "Activity.count" do
-      @investigation.is_closed = !@investigation.is_closed
-      @investigation.save!
-    end
-  end
```
slorek, line 43: I think we need a feature spec for closing a case. We have a request spec for this but it doesn't cover activity generation.

test/models/notification_test.rb
```diff
@@ -1,114 +0,0 @@
...
-  test "should not notify creator when case is closed or reopened by the creator" do
-    @investigation.update!(creator_user: users(:southampton))
-    users(:southampton_bob).own!(@investigation)
-    @investigation.reload
-    mock_investigation_updated(who_will_be_notified: [users(:southampton_bob).email])
-    @investigation.update!(is_closed: !@investigation.is_closed)
-    assert_equal 1, @number_of_notifications
-    @investigation.update!(is_closed: !@investigation.is_closed)
-    assert_equal 2, @number_of_notifications
-  end
```
slorek, line 69: Same comment as before relating to closing cases - we aren't testing notifications or activity as far as I can see.

test/models/notification_test.rb
```diff
@@ -1,114 +0,0 @@
...
-    assert_equal @number_of_notifications, 0
-  end
-
-  test "should notify all team members when the owner is a team and there is any change done by outsiders" do
-    opss_enforcement = teams(:opss_enforcement)
-    opss_enforcement.own!(@investigation)
-    mock_investigation_updated(who_will_be_notified: [users(:southampton_bob).email])
-    make_generic_change
-    assert_equal @number_of_notifications, 0
-  end
```
slorek, line 48: We should add some tests for email notifications to the feature spec as well.

slorek (changes requested):
Thanks for doing this. I think there is a bit more work to do to cover the tests being removed.

test/models/notification_test.rb
frankieroberto: Interesting. We've / I've generally not been testing emails from feature specs, and leaving that to service specs instead.  Although I see that we do have a few feature specs which do test emails.

Personally I find that things like this:

```ruby
notification_email = delivered_emails.last

expect(notification_email.personalization[:updater_name]).to eq("Bob Jones (Portsmouth Trading Standards)")
expect(notification_email.personalization[:optional_message]).to eq("Message from Bob Jones (Portsmouth Trading Standards):\n\n^ Thanks for collaborating on this case with us.")
expect(notification_email.personalization[:investigation_url]).to end_with("/cases/#{investigation.pretty_id}")
```

to be a bit implementation-specific for a feature spec. It'd be nice if we could write feature specs which just make assertions on the rendered content of the email instead, eg something like:

```ruby
expect_email(
  to: "john@example.com",
  subject: "Investigation updated",
  body: "Message from Bob"
)
```

But I'm not sure how we'd do that with Notify... 🤔

test/models/investigation_test.rb
frankieroberto: Thanks. Done in c08e41326871e8cfb472b3a920f1f1da9d7d35e9.

test/models/investigation_test.rb
frankieroberto: Thanks, done. See 21453fd059e33521bb17d7e70954161e4d65e2d2.

test/models/investigation_test.rb
frankieroberto: I've added a new feature spec for this here: 26ca8f8db9e71a1dbeb944933f0f9569bec7cfc1 – also covers activity generation (via checking the Activity page) and emails.

test/models/notification_test.rb
frankieroberto: I've created an email expectation which abstracts around the Notify mocking here: https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/943/commits/26ca8f8db9e71a1dbeb944933f0f9569bec7cfc1#diff-84f9b40234b7b60c0d51bbced9fd49a9610d55b3a557f48b62245460d97178d1 – let me know what you think.

test/models/notification_test.rb
frankieroberto: Done now in 26ca8f8db9e71a1dbeb944933f0f9569bec7cfc1

test/models/notification_test.rb
slorek: Nice!

slorek (approved):
Thanks, good work on the changes!

PR #954, opened 2020-10-16
Add a new validated barcode field, and rename the existing one

frankieroberto:
This change renames the existing "Barcode or serial number" field as "Other product identifiers", in recognition of the reality that this field currently contains a mixture of barcodes, serial numbers, model numbers, ASINs, eBay item numbers, and other unidentified codes.

It also adds a new "Barcode number" field which accepts a single barcode number (of type [UPC](https://en.wikipedia.org/wiki/Universal_Product_Code), [EAN](https://en.wikipedia.org/wiki/International_Article_Number) or [GTIN](https://en.wikipedia.org/wiki/Global_Trade_Item_Number)). This number is validated to make sure that it is of the right format (usually 12 or 13 digits) and has a valid [check digit](https://en.wikipedia.org/wiki/Check_digit).  If valid, it is stored in a standardised 13 digit format (12 digit barcodes are padded with a leading `0`).

Multiple barcodes for a single product are not currently supported, as this requires further work to understand how best to support product variants.

The barcodes are not checked for uniqueness, as we do not yet allow a single product to be associated with multiple cases.

A separate rake task moves any existing valid barcodes within the "Barcode or serial number" field to the new barcode field.

Trello card: https://trello.com/c/R1dgyUc3/764-3-add-barcode-field

## Before

Screenshot 2020-10-16 at 09 51 25

## After

Screenshot 2020-10-21 at 11 01 27

### Validation message

Screenshot 2020-10-21 at 11 02 04


slorek (commented):
Looks good, but the naming `gtin` might be confusing as a full GTI is 14 digits, padding with leading zeroes. Perhaps we should make it 14 characters and pad it, or call it `ean`? This is a bit pedantic, but our implementation differs from the official description.

slorek (approved):

frankieroberto:
@slorek as I understand it, GTIN refers to the whole system, and the numbers can be 8, 12, 13 or 14 digits long! 😱   Perhaps we should call the attribute `gtin13` to be super explicit though?

I did consider standardising on the full 14 digits, but it seems that the only use case for the 14 digit version is to be able to also encompass various types of containers of products (eg crates or shipping containers), which I don't think we need to support – and 13 digits is what you'd normally see on packaging, so that seems less confusing?

Have learnt a lot about barcodes these past 2 days!

slorek:
> @slorek as I understand it, GTIN refers to the whole system, and the numbers can be 8, 12, 13 or 14 digits long! 😱 Perhaps we should call the attribute `gtin13` to be super explicit though?
> 
> I did consider standardising on the full 14 digits, but it seems that the only use case for the 14 digit version is to be able to also encompass various types of containers of products (eg crates or shipping containers), which I don't think we need to support – and 13 digits is what you'd normally see on packaging, so that seems less confusing?
> 
> Have learnt a lot about barcodes these past 2 days!

`gtin13` would be better I think. If you Google `gtin` the first summaries result says "The GTIN is a globally unique 14-digit number" so this is where the confusion could arise.

spec/models/product_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      end
+    end
+
+    context "when setting a 12 digit UPC-A code" do
+      before do
+        product.gtin = "012345678912"
+      end
+
+      it "gets converted to a 13 digit code on validation", :aggregate_failures do
+        product.valid?
```
StupidCodeFactory, line 44: is it expected to be valid? not obvious, it would be clearer to add an exepct(product).to be_valid

app/helpers/sanitization_helper.rb
```diff
@@ -6,4 +6,21 @@ def trim_line_endings(*keys)
...
+    end
+  end
+
+  def convert_gtin_to_13_digits(*keys)
+    keys.each do |key|
+      gtin = Barkick::GTIN.new(read_attribute(key)&.strip)
+      write_attribute(key, gtin.gtin13) if gtin.valid?
+    # 8 digit codes raise an ArgumentError, as a type (UPC-E or EAN-8) needs
+    # to be specified in order to convert the code to a valid GTIN
+    rescue ArgumentError
```
StupidCodeFactory, line 22: Please change: When this happens try parse it as a EAN-8 and if this doesn't work then cary on with `next`
```ruby
gtin = Barkick::GTIN.new("01234565", type: :ean8)
```

Not very hard to implement and even if we don't have a lot for now, the future may be different.

app/validators/gtin_validator.rb
```diff
@@ -0,0 +1,19 @@
...
+#
+class GtinValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    return if value.nil?
+
+    begin
+      gtin = Barkick::GTIN.new(value)
+      # 8 digit codes raise an ArgumentError, as a type (UPC-E or EAN-8) needs
+      # to be specified in order to convert the code to a valid GTIN
+    rescue ArgumentError
```
StupidCodeFactory, line 11: Same as above ^^

StupidCodeFactory (dismissed):

slorek:
It would be worth adding this feature to the change log.

app/helpers/sanitization_helper.rb
frankieroberto: I'm slightly wary of doing this, as it seems there are some 8 digit UPC codes which are also valid EAN8 codes - but represent a different 13 digit code? ie there are some 8 digit codes which are genuinely ambiguous without knowing what type of code they are...  not 100% certain though, as it’s hard to research this stuff.

spec/models/product_spec.rb
frankieroberto: This was added here to trigger the callback tbh, but we may as well also turn it into an expectation. 👍

nasirkhanbeis (approved):
looks good as expected.

tobyhughes (approved):

PR #957, opened 2020-10-19
Make phone call editable

StupidCodeFactory:
## make phone call logging editable

1. I've remove `DateConcern` from `Correspondence` and add it to only `Correspondence::Email` and `Correspondence::Meeting`.
2. Changed the `has_one :activity` association on the correspondence to `has_many` since `PhoneCallUpdated` is new activity that can be linked to a correspondence.

Other than that it's follows the established patterns. Not the handling of the uploaded file done in the same fashion as in the create action


StupidCodeFactory:
[Make the phone call correspondence editable](https://trello.com/c/tn6KfVvP/747-make-the-phone-call-correspondence-editable)

app/decorators/audit_activity/correspondence/phone_call_updated_decorator.rb
```diff
@@ -0,0 +1,30 @@
...
+      end
+
+      def new_phone_number
+        metadata.dig("updates", "phone_number", 1)
+      end
+
+      def new_correspondence_date
+        Date.parse(metadata.dig("updates", "correspondence_date", 1)).to_s(:govuk)
+      rescue StandardError # rubocop:disable Lint/SuppressedException
+      end
```
frankieroberto, line 15: Can we avoid the rescue by checking if the value is nil first?

```suggestion
      def new_correspondence_date        
        value = metadata.dig("updates", "correspondence_date", 1)
        return if value.nil?
        Date.parse(value).to_s(:govuk)
      end
```

app/views/investigations/record_phone_calls/_form_fields.html.erb
```diff
@@ -0,0 +1,51 @@
...
+      <%= govukDetails(summaryText: "Replace this file", html: file_upload_field) %>
+    <% else %>
+      <%= file_upload_field %>
+    <% end %>
+
+  </div>
+
+  <%= form.govuk_text_area :details, label: "Notes", label_classes: "govuk-!-width-two-thirds", attributes: { maxlength: 50_000 } %>
+
+<% end %>
```
frankieroberto, line 51: Is this partial used anywhere? Looks like `_form.html.erb` is being used instead?

app/views/investigations/record_phone_calls/_form.html.erb
```diff
@@ -0,0 +1,60 @@
...
+            <%= file_upload_field %>
+          <% end %>
+
+        </div>
+
+        <%= form.govuk_text_area :details, label: "Notes", label_classes: "govuk-!-width-two-thirds", attributes: { maxlength: 50_000 } %>
+
+      <% end %>
+
+      <%= form.submit form.object.persisted? ? "Update phone call" : "Add phone call", class: "govuk-button" %>
```
frankieroberto, line 57: Please change: switch to the `govukButton` helper (which adds the class automatically plus some extra html attributes).

```suggestion
      <%= govukButton text: (form.object.persisted? ? "Update phone call" : "Add phone call") %>
```

frankieroberto (changes requested):
Looking good! A couple of minor issues to fix.

app/controllers/investigations/record_phone_calls_controller.rb
```diff
@@ -29,6 +29,38 @@ def create
...
+  end
+
+  def update
+    investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
+    authorize investigation, :update?
+
+    phone_call = Correspondence::PhoneCall.find(params[:id])
+    correspondence_form = PhoneCallCorrespondenceForm.new(phone_call_params.merge(id: phone_call.id))
+    correspondence_form.cache_file!
+    correspondence_form.load_transcript_file
```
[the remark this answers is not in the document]
StupidCodeFactory: Steve was happy to handle this inside the form. It seemed the obvious place since we need to do this dance to be able to persist the file across multiple form submissions. Hence I went ahead in my previous PR and implemented this inside the form object.

app/decorators/audit_activity/correspondence/phone_call_updated_decorator.rb
StupidCodeFactory: The value can still raise an exception and I'm not certain this is more readable so I'm going to keep it this way.

app/forms/phone_call_correspondence_form.rb
```diff
@@ -17,6 +17,28 @@ class PhoneCallCorrespondenceForm
   attribute :details
   attribute :transcript
   attribute :existing_transcript_file_id
+  attribute :id
```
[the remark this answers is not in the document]
StupidCodeFactory: I don't think it provides a lot of value
1. You'd have to prefix every attributes to be consistant and remain clear what attributes are for what model
2. _prefix_\_id is usually used to denote an associated object to the one in the current scope. 
In this class it's pretty obvious we are editing a phone call object.

app/forms/phone_call_correspondence_form.rb
```diff
@@ -17,6 +17,28 @@ class PhoneCallCorrespondenceForm
...
+
+  def self.from(phone_call)
+    new(phone_call.serializable_hash(only: ATTRIBUTES_FROM_PHONE_CALL, methods: :transcript)).tap do |form|
+      form.existing_transcript_file_id = phone_call.transcript.signed_id
+    end
+  end
+
+  def initialize(*args)
+    super
+    details&.strip!
```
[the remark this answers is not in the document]
StupidCodeFactory: I'll add a comment that's indeed not self obvious.

Basically that was done in to remove the line feed that's sent by some browser for text in textareas which would then be picked up as a change even if the field has not changed.

app/services/edit_phone_call.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def email_update_text
+    "Phone call details updated on the #{investigation.case_type.upcase_first} by #{activity.source.show}."
+  end
+
+  def check_valid_arguments!
+    context.fail!(error: "No user supplied") unless user.is_a?(User)
+    context.fail!(error: "No phone call supplied") unless correspondence.is_a?(Correspondence::PhoneCall)
+  end
```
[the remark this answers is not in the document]
StupidCodeFactory: I've renamed the method so that it is clearer what he requirements are.

app/controllers/investigations/record_phone_calls_controller.rb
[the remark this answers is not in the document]
frankieroberto: 👍  if we're happy to have this logic in the form, it'd be good to port this across to the similar email controller/form at some point soon.

app/forms/phone_call_correspondence_form.rb
```diff
@@ -17,6 +17,28 @@ class PhoneCallCorrespondenceForm
...
+    phone_number
+    overview
+    transcript
+  ].freeze
+
+  def self.from(phone_call)
+    new(phone_call.serializable_hash(only: ATTRIBUTES_FROM_PHONE_CALL, methods: :transcript)).tap do |form|
+      form.existing_transcript_file_id = phone_call.transcript.signed_id
+    end
+  end
```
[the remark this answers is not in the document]
StupidCodeFactory: yes it is a very useful pattern to abstract away the instantiation of an object. It's used a lot everywhere even in Rails , basically when you do `ActiveRecord::Record.create({})` that's a factory pattern.

https://en.wikipedia.org/wiki/Factory_method_pattern

Another good alternative is the builder pattern, for a more complex object setup.

app/controllers/investigations/record_phone_calls_controller.rb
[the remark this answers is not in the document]
slorek: To reiterate I think it's fine because we will need to do certain validations and processing on the attachment as we would any other user input.

app/models/audit_activity/correspondence/phone_call_updated.rb
```diff
@@ -0,0 +1,30 @@
...
+    "Phone call updated"
+  end
+
+private
+
+  def subtitle_slug
+    "Phone call updated"
+  end
+
+  # no-op sending of email is done by the service AddPhoneCallToCase
```
slorek, line 28:
```suggestion
  # no-op sending of email is done by the service UpdatePhoneCall
```

app/models/audit_activity/correspondence/phone_call_updated.rb
```diff
@@ -0,0 +1,30 @@
+class AuditActivity::Correspondence::PhoneCallUpdated < AuditActivity::Correspondence::Base
+  belongs_to :correspondence, class_name: "Correspondence::PhoneCall"
```
slorek, line 2: Please change: Override the `from` method inherited from the parent class (see `AuditActivity::Correspondence::AddMeeting` for the consistent example used elsewhere), or we might be able to remove it completely from the parent class.

app/services/update_phone_call.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  def email_update_text
+    "Phone call details updated on the #{investigation.case_type.upcase_first} by #{activity.source.show}."
+  end
+
+  def require_user_and_phone_call!
+    context.fail!(error: "No user supplied") unless user.is_a?(User)
+    context.fail!(error: "No phone call supplied") unless correspondence.is_a?(Correspondence::PhoneCall)
+  end
```
slorek, line 65: Suggestion: This is inconsistent with the other service classes which raise these errors inline. When I tried this before it didn't reduce the CodeClimate complexity score. I don't have a strong opinion but would prefer consistency.

app/services/update_phone_call.rb
```diff
@@ -0,0 +1,70 @@
...
+
+private
+
+  def investigation
+    correspondence.investigation
+  end
+
+  def send_notification_email
+    entities_to_notify.each do |entity|
```
slorek, line 38: Suggestion: `entities_to_notify` is often different across different business processes, sometimes it's the case creator - in this case it's all teams with access to the case except the current user. If we want to extract this out to reduce duplication I'd suggest renaming `entities_to_notify` to something more specific e.g. `email_recipients_for_team_with_access` (I realise this was done in a previous PR).

app/services/update_phone_call.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    correspondence.assign_attributes(
+      correspondence_date: correspondence_date,
+      phone_number: phone_number,
+      correspondent_name: correspondent_name,
+      overview: overview,
+      details: details
+    )
+
+    return unless any_changes?
```
slorek, line 18: Please change: Add a unit test for this behaviour.

app/services/update_phone_call.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    Correspondence.transaction do
+      if transcript && correspondence.transcript_blob != transcript
+        correspondence.transcript.attach(transcript)
+      end
```
slorek, line 23: Please change: unit tests only cover adding a new file, but this code indicates should test when no file is supplied (should it remove the existing one or keep it?), when the file is identical and when it is different.

slorek (changes requested):
Looks good, just some minor changes needed which I have noted. Thanks.

app/controllers/investigations/record_phone_calls_controller.rb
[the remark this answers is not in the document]
frankieroberto: I think it's fine too. It's _maybe_ a bit unexpected for the form object to be persisting files and creating associated models, which was why I hesitated, but equally it's also a lot of boilerplate code to have in the controller otherwise.

frankieroberto (approved):

slorek (approved):

app/decorators/audit_activity/correspondence/phone_call_updated_decorator.rb
```diff
@@ -1,12 +1,14 @@
 module AuditActivity
   module Correspondence
     class PhoneCallUpdatedDecorator < AuditActivity::Correspondence::BaseDecorator
+      REMOVED = "Removed".freeze
```
slorek: Suggestion: I18n would be a better place to manage the text.

app/decorators/audit_activity/correspondence/phone_call_updated_decorator.rb
```diff
@@ -1,12 +1,14 @@
...
+
       def new_correspondent_name
-        metadata.dig("updates", "correspondent_name", 1)
+        metadata.dig("updates", "correspondent_name", 1).presence || REMOVED
```
slorek: Suggestion: We can probably wrap this in a reusable method or replace `nil` in values once at the end.

slorek (approved):
Just a couple of suggestions.

PR #960, opened 2020-10-20
Upgrade Elasticsearch to 7.x

slorek:
Related to https://trello.com/c/1UONIGDB/757-investigate-indexing-on-most-recently-created-sort

## Description
PaaS now supports Elasticsearch 7.x, so we can now upgrade

I've created new Elasticsearch 7.x instances in `int`, `staging`, and `prod` spaces in preparation. We'll just need to manually re-create the indexes on the new instances (our seed task doesn't do this currently if there's data in the database - an improvement we could consider making in future). So this deployment will need to be done out of hours.


StupidCodeFactory (approved):
Approved but you're missing the version bump in the `docker-compose.yml`

frankieroberto (approved):

PR #963, opened 2020-10-21
Fix case ordering caused by default_scope override

slorek:
https://trello.com/c/1UONIGDB/757-investigate-indexing-on-most-recently-created-sort

## Description
Fixes case ordering on the case list pages. This affected the ordering of all results within a page, forcing them to be reordered by `updated_at` regardless of relevance or other sort options selected. Usage of `default_scope` is generally discouraged because of gotchas like this, and in fact was redundant anyway because of the implementation of search in the controller.

I added some tests around case sorting and fixed the export controller action so that it ordered correctly, removing some redundant code.

## Checklist:

### DIFF
--- a/spec/features/filter_investigations_spec.rb
+++ b/spec/features/filter_investigations_spec.rb
@@ -212,7 +212,7 @@ RSpec.feature "Case filtering", :with_elasticsearch, :with_stubbed_mailer, type:
 
   describe "sorting" do
     let(:default_filtered_cases) { Investigation.where(is_closed: false) }
-    let(:case_ids) { default_filtered_cases.order(updated_at: :desc) }
+    let(:cases) { default_filtered_cases.order(updated_at: :desc) }
 
     def select_sorting_option(option)
       within_fieldset("Sort by") { choose(option) }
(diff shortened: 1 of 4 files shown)


spec/features/filter_investigations_spec.rb
```diff
@@ -208,4 +209,48 @@
 
     expect(page).not_to have_link(restricted_case.title, href: "/cases/#{restricted_case.pretty_id}")
   end
+
+  describe "sorting" do
+    let(:default_filtered_cases) { Investigation.where(is_closed: false) }
+    let(:case_ids) { default_filtered_cases.order(updated_at: :desc) }
```
StupidCodeFactory, line 215: this would return cases not case ids it seems

StupidCodeFactory (approved):

spec/features/filter_investigations_spec.rb
slorek: Ah thanks, I did change the behaviour at the last minute but forgot to change the name.

frankieroberto (approved):

PR #966, opened 2020-10-22
Add product brand as a new field

frankieroberto:
This adds a new field to capture the 'brand' of a product, so that this can be captured separately from the product name.

This is in anticipation of the brand becoming a separate lookup table, so that a single brand can be associated with many products. This may help us to better detect duplicate products in future, as well as making the data easier to analyse.

https://trello.com/c/2zvJmakk/763-3-add-field-for-brand-on-product-page

### Before

Screenshot 2020-10-22 at 12 41 53

### After

Screenshot 2020-10-22 at 12 41 42

## Implementation notes

In order to avoid having a mixture of blanks and `nulls` in the database for this column, empty (or whitespace-only) string values for this field are converted to `nil` before saving.


slorek (approved):

nasirkhanbeis (approved):
looks good!

PR #967, opened 2020-10-22
Remove record meeting feature

slorek:
https://trello.com/c/tIUj1fqn/748-3-remove-record-meeting-feature

## Description
This feature was rarely used (8 meetings in total registered since go live, 3 of which were erroneously meetings instead of phone calls or comments), and has not been used by anyone since January 2020. Therefore we have decided to deprecate the feature.

Existing data will continue to work for now.

## Checklist:


frankieroberto (approved):
👍 

Should we remove the rspec factory for the model at this point too?

slorek:
> 👍
> 
> Should we remove the rspec factory for the model at this point too?

I think we still need this as we still need to support the display of meeting data.

PR #972, opened 2020-10-23
Fix Date parser: reject negative dates

StupidCodeFactory:
## Description
We discovered it was possible to save negative values in the govuk date,  now prevents it.


spec/models/date_parser_spec.rb
```diff
@@ -27,6 +27,12 @@
     it { is_expected.to eql(Date.new(2020, 1, 9)) }
   end
 
+  context "when given a hash containing a negative year" do
+    let(:input) { { year: "-2020", month: "1", day: "32" } }
+
+    it { is_expected.to eq(OpenStruct.new(year: "-2020", month: "1", day: "32")) }
+  end
```
frankieroberto, line 34: Suggest changing the day to a valid one, as otherwise this would pass anyway as the date is invalid.  Also suggest making the month or day a negative number instead of the year, as negative years are valid. It's unlikely that we'd need to support BC dates, but these can be validated through a separate validator (eg we already have `validate :assessed_on_cannot_be_older_than_1970` on the risk assessment form).

```suggestion
  context "when given a hash containing a negative year" do
    let(:input) { { year: "2020", month: "-2", day: "30" } }

    it { is_expected.to eq(OpenStruct.new(year: "2020", month: "-2", day: "30")) }
  end
```

app/models/date_parser.rb
```diff
@@ -22,6 +22,7 @@ def date
 
     return nil if date_values.all?(&:blank?)
     return struct_from_hash if date_values.any?(&:blank?)
+    return struct_from_hash if date_values.any?(&:negative?)
```
frankieroberto, line 25: Think we only need to check whether the month or day is negative, not the year, as `Date.new(-2020,1,1)` is valid.

```suggestion
    return struct_from_hash if date_values[1].negative? || date_values[2].negative?
```

frankieroberto (changes requested):

spec/models/date_parser_spec.rb
```diff
@@ -27,6 +27,12 @@
     it { is_expected.to eql(Date.new(2020, 1, 9)) }
   end
 
+  context "when given a hash containing a negative year" do
```
frankieroberto, line 30:
```suggestion
  context "when given a hash containing a negative month" do
```

frankieroberto (approved):

PR #978, opened 2020-10-27
Refactor owner form

slorek:
Technical debt remediation

## Description
Removes some usage of `User.current` which we are actively eliminating due to various unexpected behaviours. This leaves only some of the activity models using it, which are actively being refactored.

Removes some code which is apparently redundant relating to selecting previous owners of a case on the change ownership form. This code does not appear to be integrated anymore (see 4772e88d0f617669fb8204c07b20b522c2575504).

### DIFF
--- a/app/controllers/investigations/ownership_controller.rb
+++ b/app/controllers/investigations/ownership_controller.rb
@@ -7,6 +7,9 @@ class Investigations::OwnershipController < ApplicationController
 
   def show
     @potential_owner = form.owner&.decorate
+
+    get_potential_assignees if step == :"select-owner"
+
     render_wizard
   end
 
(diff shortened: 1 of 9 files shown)


app/controllers/investigations/ownership_controller.rb
```diff
@@ -69,4 +72,12 @@ def session_params
   def form
     @form ||= ChangeCaseOwnerForm.new(form_params)
   end
+
+  def get_potential_assignees
+    @selectable_users = [@investigation.owner_user&.object, current_user].uniq.compact
+    @team_members = User.get_team_members(user: current_user)
```
frankieroberto, line 78: I know this is an existing method, but I'd be tempted to refactor it to this, and remove the class method from `User`.

```suggestion
    @team_members = current_user.team.users.active
```

frankieroberto (approved):

PR #980, opened 2020-10-27
Fix Sidekiq::Web showing Forbidden errors

slorek:
Resolves a long-standing issue with Sidekiq::Web returning `Forbidden` errors when performing operations such as kill/delete.

Our application still uses the outdated secrets approach to storing the secret key, so the previous code always returned `nil`. This is the correct approach to read the secret key as described in the [documentation](https://api.rubyonrails.org/classes/Rails/Application.html#method-i-secret_key_base).

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,7 +1,7 @@
 require "sidekiq/web"
 require "sidekiq/cron/web"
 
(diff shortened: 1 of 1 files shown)


StupidCodeFactory (approved):
Good find! 👏

PR #982, opened 2020-10-28
Fix images not saved for enquiry

StupidCodeFactory:
## Description

The `DateConcern` was causing the `Enquiry` model all contain changes. When attaching an document Rails would never call `save` on the object detecting that it had other changes and hence on changing the attributes in memory.

```ruby
# this initialises other :date_received_day, :date_received_month, :date_recieved_year
# on on every object instantiation causing enquiry#changes? to be true  
date_attributes :date_received 
```

I've removed the `DateConcern` from the `Enquiry` and fixed the Javascript that set date to today or tomorrow. 

Fixed the "Other received type" which was not accessible by adding a label text and applying the "govuk-visually-hidden" class.
This field was also not retained when the form re-displays for validation errors.

## Notes 
It was impossible to write an integration test to reproduce the bug as it would save the attachement correctly 🤷‍♂️


slorek (approved):

nasirkhanbeis (approved):
looks good as expected

config/locales/en.yml
```diff
@@ -350,6 +350,14 @@ en:
       collaboration/access/read_only: "readonly"
     errors:
       models:
+        investigation/enquiry:
+          attributes:
+            date_received:
+              blank: Enter the date the enquiry was received
+              must_be_real: Date received of enquiry must be a real date
+              incomplete: Date received of enquiry must include a %{missing_date_parts}
+              in_future: Date received of enquiry must be today or in the past
```
frankieroberto, line 359: Think these need a bit of re-wording...

Suggestion:

```suggestion
              must_be_real: Date enquiry received must be a real date
              incomplete: Date enquiry received must include a %{missing_date_parts}
              in_future: Date enquiry received must be today or in the past
```

frankieroberto (changes requested):
Looks good, thanks!

Just needs some tweaks to the error messages, I think.

config/locales/en.yml
StupidCodeFactory: cheers! yes does read much better indeed.

frankieroberto (approved):

PR #983, opened 2020-10-28
Refactor code for updating coronavirus status

slorek:
## Description
Refactors the code around updating a case's coronavirus status to reflect the new established patterns:

* Service object for updating the status
* Audit activity generating metadata rather than hardcoded text
* Remove dependency on `User.current`, support for which is deprecated
* Refactored services to use `EntitiesToNotify#email_recipients_for_case_owner` where appropriate, removing dependencies on the definitions in `Activity` which will be removed once all activities are similarly refactored.

I've created a data migration to move the old activity to metadata. Although the existing data does not store the new/previous status values, we can infer these from the title/body text stored on the activity. There are over 360 instances in production currently.


app/controllers/investigations/coronavirus_related_controller.rb
```diff
@@ -9,14 +9,13 @@ def show
...
       @coronavirus_related_form = CoronavirusRelatedForm.new(params.require(:investigation).permit(:coronavirus_related))
       return render :show unless @coronavirus_related_form.valid?
 
-      @investigation.coronavirus_related = @coronavirus_related_form.coronavirus_related
+      result = ChangeCaseCoronavirusStatus.call!(investigation: @investigation, status: @coronavirus_related_form.coronavirus_related, user: current_user)
 
-      if @investigation.coronavirus_related_changed?
-        @investigation.save!
-        AuditActivity::Investigation::UpdateCoronavirusStatus.from(@investigation)
+      if result.updated_coronavirus_status
```
frankieroberto, line 18: Is there any reason why `updated_coronavirus_status` needs to be set? Could this just be `if result` or `if result.success?`?

app/services/change_case_coronavirus_status.rb
```diff
@@ -0,0 +1,68 @@
...
+    return if investigation.changes.none?
+
+    ActiveRecord::Base.transaction do
+      investigation.save!
+      create_audit_activity_for_coronavirus_status_changed
+    end
+
+    send_notification_email
+
+    context.updated_coronavirus_status = status
```
frankieroberto, line 22: Bit unclear why this needs to be set.

app/services/change_case_coronavirus_status.rb
```diff
@@ -0,0 +1,68 @@
+class ChangeCaseCoronavirusStatus
```
frankieroberto, line 1: Suggest renaming this to be consistent with the other update services and the wording in the UI (the button). Although we do have some other services prefixed with `Change` so maybe it doesn’t matter.

```suggestion
class UpdateCaseCoronavirusStatus
```

Screenshot 2020-10-28 at 15 39 48

app/services/concerns/entities_to_notify.rb
```diff
@@ -13,4 +13,15 @@ def email_recipients_for_team_with_access
     end
     entities.uniq - [context.user]
   end
+
+  # Notify the case owner, unless it is the same user/team performing the change
+  def email_recipients_for_case_owner
+    entities = [investigation.owner] - [user, user.team]
```
frankieroberto, line 19: I think we should avoid using `investigation.owner`, as the intent is a bit ambiguous if there's both an owner_user and a owner_team.

As I understand it, the logic is _meant_ to be something like "if there's an owner user, notify them unless they're the user who did the action. If there's no owner user, notify the team email if there is one, otherwise notify all users of that team except the current user"?

But I'm not 100% clear!

frankieroberto (approved):
Looks good - another step towards dropping `User.current`...

Left some minor comments and questions.

app/controllers/investigations/coronavirus_related_controller.rb
slorek: This is needed because the flash message should not be displayed if there was no change. The service will return `success` even if there were no changes. I borrowed the pattern from `ChangeCaseRiskLevel` - if you have a better suggestion I'm happy to consider it.

app/services/change_case_coronavirus_status.rb
slorek: We have as many services using the `Change` prefix currently. Agree we should standardise the terminology (as we should with investigation/case), but that can be done separately.

app/services/concerns/entities_to_notify.rb
slorek: I've borrowed and consolidated a different implementation of very similar logic from another service which hopefully is easier to follow, albeit more lines of code.

app/controllers/investigations/coronavirus_related_controller.rb
frankieroberto: Ahh, that makes sense.

Possibly something like `context.changes_made = true` would be more obvious, and might work as a pattern elsewhere, as the `coronavirus_status` attribute is already in the `context` anyway?

Happy to leave as-is though.

app/services/concerns/entities_to_notify.rb
frankieroberto: Thanks - that's a bit easier to follow for my brain anyway.

I'd forgotten about the "don't notify anyone if the case is only owned by a team, and I'm a member of that team" clause... 🤦

app/controllers/investigations/coronavirus_related_controller.rb
slorek: I've refactored as I think that is slightly better. Thanks!

PR #985, opened 2020-10-28
Fix randomly failing test

frankieroberto:
These tests sometimes fail due to `Faker::Hipster.word` sometimes randomly producing the same word multiple times, which then get removed by the call to `uniq!` in the model.

Calling the [`.unique`](https://github.com/faker-ruby/faker#ensuring-unique-values) method on the Faker classes fixes this.


StupidCodeFactory (changes requested):
Please change: use `.unique` method https://github.com/faker-ruby/faker#ensuring-unique-values

spec/models/investigation_spec.rb
```diff
@@ -174,38 +174,34 @@
   end
 
   describe "#categories" do
-    let(:product_category) { Faker::Hipster.word }
```
StupidCodeFactory, line 177: To guaranty uniqueness faker introduced this concept: `Faker::Hipster.unique.word`. 
I see the argument for having plain string to improve readability of the test. 
I prefer the randomness and assigning the result into a `let` variable with a descriptive enough name and use it in the test. Also if you change one the test will break. you'll have to fix the other in a separate test run, it seems more brittle to me.

StupidCodeFactory: But I dunno what do you prefer @slorek?

slorek: I prefer to refer to variables by name rather than value, too. Shall we try using `unique`? I haven't seen that before.

frankieroberto: ok, done. @StupidCodeFactory thanks for pointing out the `.unique` method!

StupidCodeFactory (approved):
@frankieroberto thanks for the change!

PR #987, opened 2020-10-29
Refactor service notifications

slorek:
This refactors various service objects and their specs to standardise the methodology for notifying the case owner of a change, and the related specs for each of the various scenarios. This ensures that services behave consistently and are well tested.

Also fixes a bug in the `ChangeCaseOwner` service when passed a stale investigation object. This could result in the wrong user/team being notified of a change, but in practise only really affected usage in specs.

Fixed some issues with the `UpdateTestResult` specs which reported false positives, caused by the callbacks on the tested models.


spec/support/shared_examples/service_which_notifies_the_case_owner.rb
```diff
@@ -0,0 +1,83 @@
...
+          expected_email_body("#{user.name} (#{user.team.name})"),
+          expected_email_subject
+        ]))
+      end
+
+      context "when the owner team does not have an email address" do
+        let(:other_team) { create(:team, team_recipient_email: nil) }
+
+        # Create an inactive user to test email is not delivered to them
+        before { create(:user, team: other_team, organisation: other_team.organisation) }
```
frankieroberto, line 69: Might be worth explicitly using the `:inactive` trait here, in case we change the default for the user factory in future?

Could also include a deleted user?

```suggestion
        before do
          create(:user, :inactive, team: other_team, organisation: other_team.organisation)
          create(:user, :deleted, team: other_team, organisation: other_team.organisation)
        end
        
```

frankieroberto (approved):
Nice work, thanks!

spec/services/update_test_result_spec.rb
```diff
@@ -136,42 +123,34 @@
         end
 
         let(:result) { described_class.call(test_result: test_result, new_attributes: new_attributes, user: user, new_file_description: "Updated description") }
+        let(:activity_timeline_entry) { test_result.investigation.activities.where(type: AuditActivity::Test::TestResultUpdated.to_s).order(:created_at).last }
```
StupidCodeFactory, line 126:
```suggestion
        let(:activity_timeline_entry) { test_result.investigation.activities.where(type: "AuditActivity::Test::TestResultUpdated").order(:created_at).last }
```

StupidCodeFactory (approved):

nasirkhanbeis (approved):
Tested notifications briefly, looking good as expected. nice work!!

PR #993, opened 2020-11-04
Bug fix: Accessing a MSA case creation wizard step page before starting from the beginning

slorek:
https://trello.com/c/Wyv846Fk/694-interim-journey-urls-when-accessed-directly-without-session-data-then-it-should-return-to-dashboard

## Description
This fixes a bug whereby a user could access a page in the MSA case creation flow directly without having started the wizard. This could sometimes render an error if assumed session data was missing, or otherwise create a broken state with unexpected results.

This is an interim quick fix before refactoring the offending controller completely.

Rather than redirecting to the dashboard I have chosen to redirect to the first wizard page. Redirecting to the `new` action is better than redirecting to the first step using the Wicked helper because it sets up the session data in the call to `clear_session`. This is important even though it's not obvious.

## Checklist:

### DIFF
--- a/app/controllers/investigations/ts_investigations_controller.rb
+++ b/app/controllers/investigations/ts_investigations_controller.rb
@@ -24,6 +24,7 @@ class Investigations::TsInvestigationsController < ApplicationController
         :other_files,
         :reference_number
 
+  before_action :redirect_to_first_step_if_wizard_not_started, if: -> { step && (step != steps.first) }
   before_action :set_countries, only: %i[show create update]
   before_action :set_product, only: %i[show create update]
   before_action :set_investigation, only: %i[show create update]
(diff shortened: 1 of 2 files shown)


StupidCodeFactory (approved):

nasirkhanbeis:
I had a bit of play around sessions and direct access to pages, fix looks good. it puts back user to starting point of case creation which is coronavirus flag, however I think if not on home page it should put user back to /cases page . 

Also I found another related issue which is returning 422 error. I will raise that separately.

nasirkhanbeis:
Another related issue spotted,
Steps to replicate
1. As TS user, create a new case
2.Follow the journey up till case reference number question, copy the url
3.Open new browser and login as ts user (in same review app)
4.Now go to the url as copied in step 2
5. It doesn't redirect to start of the journey but prompts to enter reference number, 
6. Select option 'no' and click create 
Returns blank page.
Expected to return user to starting point of case creation journey or /cases page.  I think latter will be much better option so user can choose to create case or whatever

PR #1012, opened 2020-11-10
Fix email validation bug when whitelisting is turned on

KyleMacPherson:
## Description
Prior to this commit whitelisting emails is enabled we see errors being raised if an incorrectly formatted email is entered. We did not see this issue raised in tests, nor could we replicate it in development, because `EMAIL_WHITELIST_ENABLED` is set to false.

The issue is that Mail::Field::IncompleteParseError is raised but not handled.

This commit rescues from this error and adds the whitelisted domain error to our form.

Note: This will raise both an error for incorrectly formatted error and an error for non-whitelisted email. The other option I considered was to return from `#email_domain_must_be_whitelisted` if there were existing validation errors, but while I believe that `validates` run before `validate` I can't seem to find any concrete documentation on that, and so I do not want to rely on the validations running in order (e.g validating email formatting before this whitelisting validation.)

Another option could be to validate format of the email again, but that is obviously more expensive.

https://trello.com/c/m425Ruk8/810-email-validations-bug

Screenshots (with whitelisting turned on):
Before:
Screenshot 2020-11-10 at 13 43 02

After:
Screenshot 2020-11-10 at 13 41 09

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/forms/invite_user_to_team_form.rb
+++ b/app/forms/invite_user_to_team_form.rb
@@ -34,6 +34,9 @@ private
 
   def email_domain_must_be_whitelisted
     address = Mail::Address.new(email)
+  rescue Mail::Field::IncompleteParseError
+    errors.add(:email, :not_whitelisted, opss_enquiries_email: opss_enquiries_email)
+  else
     errors.add(:email, :not_whitelisted, opss_enquiries_email: opss_enquiries_email) unless whitelisted_domains.include?(address.domain.to_s.downcase)
   end
 
--- a/app/forms/invite_user_to_team_form.rb
+++ b/app/forms/invite_user_to_team_form.rb
@@ -35,7 +35,7 @@ private
   def email_domain_must_be_whitelisted
     address = Mail::Address.new(email)
   rescue Mail::Field::IncompleteParseError
-    return
+    # This error will be thrown when an email address is incorrectly formatted. The form will be invalid based on that so no need to do add error here.
   else
     errors.add(:email, :not_whitelisted, opss_enquiries_email: opss_enquiries_email) unless whitelisted_domains.include?(address.domain.to_s.downcase)
   end
--- a/app/forms/invite_user_to_team_form.rb
+++ b/app/forms/invite_user_to_team_form.rb
@@ -35,7 +35,7 @@ private
   def email_domain_must_be_whitelisted
     address = Mail::Address.new(email)
   rescue Mail::Field::IncompleteParseError
-    errors.add(:email, :not_whitelisted, opss_enquiries_email: opss_enquiries_email)
+    return
   else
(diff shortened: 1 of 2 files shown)


spec/forms/invite_user_to_team_form_spec.rb
```diff
@@ -26,7 +26,7 @@
       it "populates an error message" do
         form.validate
         errors.each do |property, message|
-          expect(form.errors.full_messages_for(property)).to eq([message])
+          expect(form.errors.full_messages_for(property)).to eq([message].flatten)
```
StupidCodeFactory, line 29: 🤔 this looks a bit odd. but 🤷‍♂️

StupidCodeFactory (approved):

app/forms/invite_user_to_team_form.rb
```diff
@@ -34,6 +34,9 @@ def cannot_be_existing_user_on_another_team
 
   def email_domain_must_be_whitelisted
     address = Mail::Address.new(email)
+  rescue Mail::Field::IncompleteParseError
+    errors.add(:email, :not_whitelisted, opss_enquiries_email: opss_enquiries_email)
+  else
```
slorek, line 39: Please change: I think in this case we do not need to add another error. The invalid email will be caught by the other validations on the model. As a rule, we only want to show one error message per form field, so I think in this case we would just want to show the "Enter an email address in the correct format, like name@example.com" error. We can just `return` here instead which should give us the behaviour we need.

slorek (changes requested):

app/forms/invite_user_to_team_form.rb
KyleMacPherson: @slorek what do you think should trigger the return? 

Shall we return here if errors is not empty and if so are you confident that email validations will run before this? (I am confident in this but not certain!) or would you rather we validated the format of the email again here and return if the email format is invalid?

slorek: I think the problem is `email_domain_must_be_whitelisted` is run regardless of whether `email` is present or valid. We can see that the email validator on https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/cd99249b64b0ff1ed67e82a5670e01e5571ea303/app/forms/invite_user_to_team_form.rb#L8 is run only if the email is present. We should probably have a similar condition on `email_domain_must_be_whitelisted` as well so that it doesn't run if no email is supplied.

I think it's safe to assume the other validation will run if the email is present so I think we can simply change this to

```ruby
rescue Mail::Field::ParseError
  return
else
```
which will catch some other potential issues as well.

KyleMacPherson: @slorek thanks boss!

slorek (approved):

PR #1015, opened 2020-11-10
fix risk assessment uploaded document cleared when create/edit form has a validation error

StupidCodeFactory:
## Description

This solves the problem whereby the attachment is lost upon validation error. 

I've used the same pattern we in other controller, unifying the interface: `FormObject#cache_file!` `FormObject#load_attribute_name_file` 

I've also created a `app/views/application/_upload_file_component.html.erb ` so that partial could be reused wherever we have file to upload.

## Checklist:

### DIFF
--- a/app/services/update_risk_assessment.rb
+++ b/app/services/update_risk_assessment.rb
@@ -15,7 +15,7 @@ class UpdateRiskAssessment
     @previous_attachment_filename = risk_assessment.risk_assessment_file.filename
 
     ActiveRecord::Base.transaction do
-      risk_assessment.attributes = {
+      risk_assessment.assign_attributes(
         assessed_on: assessed_on,
         risk_level: risk_level,
         custom_risk_level: custom_risk_level.presence,
@@ -24,7 +24,7 @@ class UpdateRiskAssessment
         assessed_by_other: assessed_by_other.presence,
         details: details,
         product_ids: product_ids
-      }
+      )
 
       if risk_assessment_file
         risk_assessment.risk_assessment_file.detach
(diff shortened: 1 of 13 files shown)


StupidCodeFactory:
[Risk assessment - uploaded file not to be removed on other field validations](https://trello.com/c/f2q8kga0/685-risk-assessment-uploaded-file-not-to-be-removed-on-other-field-validations)

spec/rails_helper.rb
```diff
@@ -10,10 +10,10 @@
     c.single_report_path = "coverage/lcov.info"
   end
   SimpleCov.formatter = SimpleCov::Formatter::LcovFormatter
-end
 
-SimpleCov.start "rails" do
-  enable_coverage :branch
+  SimpleCov.start "rails" do
+    enable_coverage :branch
+  end
```
StupidCodeFactory, line 16: avoid running coverage locally and polluting the dev machine hard drives.

app/services/update_risk_assessment.rb
```diff
@@ -15,7 +15,7 @@ def call
     @previous_attachment_filename = risk_assessment.risk_assessment_file.filename
 
     ActiveRecord::Base.transaction do
-      risk_assessment.attributes = {
+      risk_assessment.assign_attributes(
```
StupidCodeFactory, line 18: just for consistency

spec/rails_helper.rb
slorek: 🤔 I did research this before. It only adds a couple of seconds to the test run time and I thought it would be useful to have the HTML report (it's easier to digest as well). The existing files will get overwritten so it shouldn't cause too much drive usage.

nasirkhanbeis (approved):
looks good as expected

slorek (approved):

PR #1027, opened 2020-11-13
Show a success message when adding an attachment. Add feature specs

slorek:
https://trello.com/c/hU0jTM3S/290-no-confirmation-message-banner-after-attachment-is-added

Shows a confirmation message when an attachment is created. Note we currently have no way of knowing if the file was actually processed and not corrupt or invalid, so this is misleading in some circumstances. We have a story in the backlog to correct this.

I have also added feature specs around adding attachments to cases, products and businesses.


KyleMacPherson (approved):

StupidCodeFactory (approved):

nasirkhanbeis (approved):
looks good as expected!

PR #1029, opened 2020-11-13
Add user without opss_user role to seed file

KyleMacPherson:
This PR adds a user to our seed file that does not have `opss_user` role. We currently do not have a this kind of user and this will be useful to test the different flow available to trading standards users.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -485,8 +485,10 @@ if run_seeds
     end
   else
     organisation = Organisation.create!(name: "Office for Product Safety and Standards")
+    trading_standards = Organisation.create!(name: "Trading Standards")
     enforcement = Team.create!(name: "OPSS Enforcement", team_recipient_email: "enforcement@example.com", "organisation": organisation)
     operational_support = Team.create!(name: "OPSS Operational support unit", team_recipient_email: nil, "organisation": organisation)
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -513,6 +513,16 @@ def create_blob(filename, title: nil, description: nil)
...
       mobile_number: ENV.fetch("TWO_FACTOR_AUTH_MOBILE_NUMBER")
     )
+    user3 = User.create!(
+      name: "TS User",
+      email: "ts_user@example.com",
+      password: "testpassword",
+      password_confirmation: "testpassword",
+      organisation: organisation,
+      mobile_number_verified: true,
+      team: enforcement,
```
slorek, line 523: Please change: `organisation` and `team` should be different as otherwise they will be on the same OPSS team as the OPSS users.

slorek (changes requested):

db/seeds.rb
KyleMacPherson: @slorek This looking better now?

slorek (approved):

PR #1042, opened 2020-11-23
Add fields to csv export

KyleMacPherson:
https://trello.com/c/bKJRQ5WW/817-add-date-related-fields-to-export

## Description
Add date created, date updated and date closed to the exported csv file.
- Date created and date updated based upon the investigation's created_at/updated_at fields.
- Date closed based upon the created_at field of an UpdateStatus activity which recorded the investigation being closed

[cases (14).xlsx](https://github.com/UKGovernmentBEIS/beis-opss-psd/files/5583781/cases.14.xlsx)

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/controllers/investigations_controller.rb
+++ b/app/controllers/investigations_controller.rb
@@ -19,7 +19,7 @@ class InvestigationsController < ApplicationController
         authorize Investigation, :export?
 
         @answer = search_for_investigations
-        @investigations = @answer.records(includes: %i[complainant creator_user products owner_team owner_user])
+        @investigations = @answer.records(includes: %i[complainant creator_user products owner_team owner_user activities])
 
         @activity_counts = Activity.group(:investigation_id).count
         @business_counts = InvestigationBusiness.unscoped.group(:investigation_id).count
--- a/app/models/investigation/allegation.rb
+++ b/app/models/investigation/allegation.rb
@@ -17,9 +17,5 @@ class Investigation < ApplicationRecord
     def case_created_audit_activity_class
       AuditActivity::Investigation::AddAllegation
     end
-
-    def closing_activity
-      activities.where(title: "Allegation closed").last
-    end
   end
 end
--- a/app/models/investigation/allegation.rb
+++ b/app/models/investigation/allegation.rb
@@ -17,5 +17,9 @@ class Investigation < ApplicationRecord
     def case_created_audit_activity_class
       AuditActivity::Investigation::AddAllegation
     end
+
+    def closing_activity
+      activities.where(title: "Allegation closed").last
+    end
   end
 end
--- a/spec/requests/export_investigations_spec.rb
+++ b/spec/requests/export_investigations_spec.rb
@@ -152,10 +152,10 @@ RSpec.describe "Export investigations as XLSX file", :with_elasticsearch, :with_
         end
       end
 
-      context 'when investigation has close investigation activity' do
+      context "when investigation has close investigation activity" do
         it "exports closing date in date_closed column" do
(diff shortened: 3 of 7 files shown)


app/models/investigation/allegation.rb
```diff
@@ -17,5 +17,9 @@ def case_type
     def case_created_audit_activity_class
       AuditActivity::Investigation::AddAllegation
     end
+
+    def closing_activity
+      activities.where(title: "Allegation closed").last
```
slorek, line 22: Please change: Use `type: "AuditActivity::Investigation::UpdateStatus"` (as well as being less brittle, you can do this on the parent class and avoid the need to define the method differently on all the subclasses. It will also simplify the tests).

I'd also suggest returning `nil` if the case is not closed (`is_closed?`). This would avoid taking a 'reopened' activity as closed.

slorek (changes requested):
We'll also need to check how this affects performance as this endpoint is quite slow already.

I would suggest it might be useful to add a `date_closed` attribute to `Investigation` as part of this change to avoid this roundabout way of finding the date which will be quite slow and not 100% reliable. We just need to remember to blank it again when a case is reopened.

app/models/investigation/allegation.rb
KyleMacPherson: @slorek does `"AuditActivity::Investigation::UpdateStatus"` always relate to an activity that closes an investigation. I imagined that the status could be updated in a different way (perhaps opened again as you mention).

Good idea with regard to returning nil if closed is not closed, will do 👍

spec/requests/export_investigations_spec.rb
```diff
@@ -135,6 +135,85 @@
...
+      context "when investigation has close investigation activity" do
+        context "when investigation is an allegation" do
+          it "exports closing date in date_closed column" do
+            investigation = create(:allegation, is_closed: true)
+            investigation.activities.create!(title: "Allegation closed", type: AuditActivity::Investigation::UpdateStatus)
+
+            Investigation.import refresh: true, force: true
+            get investigations_path, params: { status_closed: "checked", status_open: "checked", format: :xlsx }
+
+            aggregate_failures do
```
StupidCodeFactory, line 163: Suggestion: we usually add the `:aggregate_failures` as metadata of the `it` block 
```ruby
it "exports closing date in date_closed column", :aggregate_failures do
  # ...
end
```

reduces the number of lines in the test

app/controllers/investigations_controller.rb
```diff
@@ -19,7 +19,7 @@ def index
         authorize Investigation, :export?
 
         @answer = search_for_investigations
-        @investigations = @answer.records(includes: %i[complainant creator_user products owner_team owner_user])
+        @investigations = @answer.records(includes: %i[complainant creator_user products owner_team owner_user activities])
```
StupidCodeFactory, line 22: Suggestion: good spot that you've thought about the n+1 queries! 👏 
This will however loads a tone of activities. I wonder if by just adding 
```ruby
  has_many :updated_status_activities`, -> { where(type: "AuditActivity::Investigation::UpdateStatus").order(created_at: :desc) }
end

# in the controller 
@answer.records(includes: %i[complainant creator_user products owner_team owner_user updated_status_activities])

```
This would generate a little more optimised query and only loads those audit records.
This export is already a bit slow and we need  to be careful that report can still be genrated within the web request's timeout.

StupidCodeFactory (approved):

PR #1045, opened 2020-11-24
Add date closed to investigation

KyleMacPherson:
Adding date_closed field to `investigations` so that we do not have to query activities and get slow/unreliable results when attempting to find out when an investigation was closed. Also add additional fields to the CSV export.

## Description
- Migration to add `date_closed` to investigations
- Change in controller to populate this field.
- Adding test coverage for this change
- Add fields to csv export
- Add tests to cover csv export changes
- Rake task to backfill field for existing closed investigations

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/lib/tasks/backfill_date_closed.rake
+++ b/lib/tasks/backfill_date_closed.rake
@@ -1,0 +1,13 @@
+namespace :investigations do
+  desc "Backfill date_closed field"
+  task backfill_date_closed: :environment do
+    closed_investigations = Investigation.includes(:activities).where(is_closed: true, date_closed: nil)
+
+    puts "Backfilling date closed for #{closed_investigations.count} closed investigations"
+
+    closed_investigations.each do |investigation|
+      date_closed = investigation.activities.where(type: "AuditActivity::Investigation::UpdateStatus").last.created_at
+      investigation.update(date_closed: date_closed)
+    end
+  end
+end
--- a/lib/tasks/backfill_date_closed.rake
+++ b/lib/tasks/backfill_date_closed.rake
@@ -7,7 +7,7 @@ namespace :investigations do
 
     closed_investigations.each do |investigation|
       date_closed = investigation.activities.where(type: "AuditActivity::Investigation::UpdateStatus").last.created_at
-      investigation.update(date_closed: date_closed)
+      investigation.update!(date_closed: date_closed)
     end
   end
 end
--- a/lib/tasks/backfill_date_closed.rake
+++ b/lib/tasks/backfill_date_closed.rake
@@ -6,7 +6,7 @@ namespace :investigations do
     puts "Backfilling date closed for #{closed_investigations.count} closed investigations"
 
     closed_investigations.each do |investigation|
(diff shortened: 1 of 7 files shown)


lib/tasks/backfill_date_closed.rake
```diff
@@ -0,0 +1,13 @@
+namespace :investigations do
+  desc "Backfill date_closed field"
+  task backfill_date_closed: :environment do
+    closed_investigations = Investigation.includes(:activities).where(is_closed: true, date_closed: nil)
+
+    puts "Backfilling date closed for #{closed_investigations.count} closed investigations"
+
+    closed_investigations.each do |investigation|
+      date_closed = investigation.activities.where(type: "AuditActivity::Investigation::UpdateStatus").last.created_at
```
slorek, line 9: Please change: activities are ordered by most recent first, so we would actually want the `first` one. See https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/app/models/investigation.rb#L50. You may want to explicitly override the order to ensure this isn't changed.

spec/requests/investigations/updating_the_status_of_a_case_spec.rb
```diff
@@ -29,10 +29,30 @@
...
-        expect(investigation.reload.is_closed).to be true
-        expect(response).to redirect_to(investigation_path(investigation))
+    context "when status is changed to closed" do
+      it "updates the investigation and redirects to the investigation page" do
+        aggregate_failures do
+          expect(investigation.reload.is_closed).to be true
+          expect(investigation.date_closed).not_to be nil
+          expect(response).to redirect_to(investigation_path(investigation))
+        end
+      end
```
slorek, line 39:
```suggestion
      it "updates the investigation and redirects to the investigation page", :aggregate_failures do
        expect(investigation.reload.is_closed).to be true
        expect(investigation.date_closed).not_to be nil
        expect(response).to redirect_to(investigation_path(investigation))
      end
```

spec/requests/investigations/updating_the_status_of_a_case_spec.rb
```diff
@@ -29,10 +29,30 @@
...
+              }
+      end
+
+      it "updates the investigation and redirects to the investigation page" do
+        aggregate_failures do
+          expect(investigation.reload.is_closed).to be false
+          expect(investigation.date_closed).to be nil
+          expect(response).to redirect_to(investigation_path(investigation))
+        end
       end
```
slorek, line 56:
```suggestion
      it "updates the investigation and redirects to the investigation page", :aggregate_failures do
        expect(investigation.reload.is_closed).to be false
        expect(investigation.date_closed).to be nil
        expect(response).to redirect_to(investigation_path(investigation))
      end
```

slorek (changes requested):
Minor changes needed but looks good.

slorek (approved):

PR #1046, opened 2020-11-24
Add which standard was the product tested against

StupidCodeFactory:
## Description

A few moving parts. Before adding the new field I've standardise create/update with form object `TestResultForm` 
coupled with two service class `AddTestResultToInvestigation`  and updated the `UpdateTestResult` to receive in the context each attribute i.e not bundled into a `Hash`. Also injected the `changes` that are directly from the form.

I was then able to add the the new field working seamlessly for the opss flow or trading standard flow.

Screenshot 2020-12-14 at 16 04 54

## Checklist:


app/helpers/tests_helper.rb
```diff
@@ -12,12 +12,19 @@ def test_result_summary_rows(test_result)
...
         value: { text: test_result.legislation }
-      },
-      {
-        key: { text: "Result" },
-        value: { text: test_result.result.upcase_first }
       }
     ]
+    if test_result.standards_product_was_tested_against.present?
+      rows << {
+        key: { text: "Standards" },
```
KyleMacPherson, line 19: Should the strings here be in the en.yml file?

app/views/investigations/test_results/_form.html.erb
```diff
@@ -22,15 +22,18 @@
 
 
 <%= form.govuk_autocomplete(:legislation, label: "Against which legislation?", label_classes: "govuk-label--m", choices: relevant_legislation, hint: "Select the relevant legislation from the list.") %>
-
+<%= form.govuk_input :standards_product_was_tested_against, hint: "For example, EN71. Use comma to separate multiple standards.", label: "Which standard was the product tested against?", value: form.object.standards_product_was_tested_against&.join(", ")  %>
```
KyleMacPherson, line 25: Think the strings here should be in en.yml too

db/migrate/20201125105949_add_standars_product_was_testing_against_to_tests.rb
```diff
@@ -0,0 +1,8 @@
+class AddStandarsProductWasTestingAgainstToTests < ActiveRecord::Migration[6.0]
+  def change
+    safety_assured do
+      # TODO: add default to empty array in later migration
```
KyleMacPherson, line 4: Any reason not to do this in this PR?

lib/active_model/types/comma_separated_list.rb
```diff
@@ -0,0 +1,12 @@
+module ActiveModel
+  module Types
+    class CommaSeparatedList < ActiveRecord::Type::Value
```
KyleMacPherson, line 3: Is this related to this change?

app/controllers/investigations/test_results_controller.rb
```diff
@@ -1,153 +1,96 @@
...
-  def test_result_attributes
-    params.require(:test_result).permit(:product_id, :legislation, :result, :details, date: %i[day month year])
-  end
+    return redirect_to investigation_test_result_path(investigation, test_result) if result.success?
 
-  def investigation_from_params
-    Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
-      .decorate
+    @investigation = investigation.decorate
+    render "edit"
```
KyleMacPherson, line 74: Seems like a bit of a smell to have 2 renders of the same view in the same controller. Can`UpdateTestResult` fail if the `test_result_form` is valid? If it does fail will errors be shown on the page render?

KyleMacPherson (changes requested):
Nice work Yann, looks like it brings the test results stuff in line with the pattern we use elsewhere which seems like a good idea! Just had a couple of comments/questions

nasirkhanbeis:
spotted few issues 
1. regression issue - cannot create allegation/enquiry as opss user - returns service error after I submit the contact details page
2. TS flow - add test result - standard field is not mandatory- when I submit with empty field then it should throw error
3. Edit test result - edit standard field doesn't record in activity log

db/migrate/20201125105949_add_standars_product_was_testing_against_to_tests.rb
StupidCodeFactory: Yeah I want to do this in a seperate PR.

app/views/investigations/test_results/_form.html.erb
StupidCodeFactory: No we only use `I18n` for error message and string inside models. I would have been happy to have everything in `I18n` but this is not what we've been doing so far.

app/helpers/tests_helper.rb
StupidCodeFactory: no we don't really use I18n in views. It is mostly error model's messages.

lib/active_model/types/comma_separated_list.rb
StupidCodeFactory: yes required here https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/1046/files#diff-6e6c15bed988a566b6e12a23f5dd5794c5da1a06ea8e4a491149b1b74fad6602R3 

and used here :

https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/1046/files#diff-47435d37d60442ab18b9dc65f83c618abeb5c69aa5d48b80925dea2a6f508feeR14

app/controllers/investigations/test_results_controller.rb
StupidCodeFactory: This is the pattern we've been using throughout the code base. I'm not particularly keen on it, particularly that we don't pass the form to service object but a bunch of random key value pairs, making the interface of service object totally unpredictable. 
It then necessary for controller to 1) validate the form, 2) know what and how to pass to the service object, 3) run the service object 4) hand the result

In my opinion the controller is doing far too much but that's the abstraction we all agreed on and the we have been following in most places.

slorek: I think this is a relatively complex method, but overall is consistent with others as Yann says.

app/models/audit_activity/risk_assessment/risk_assessment_added.rb
```diff
@@ -17,7 +17,7 @@ def self.build_metadata(risk_assessment)
     }
   end
 
-  def title(_)
+  def title(_ = nil)
```
slorek, line 20:
```suggestion
  def title(*)
```

app/models/audit_activity/test/result.rb
```diff
@@ -1,16 +1,14 @@
 class AuditActivity::Test::Result < AuditActivity::Test::Base
```
slorek, line 1: Please change: When overriding `from` you also need to override the `notify_relevant_users` callback method inherited from `Activity` to prevent a second duplicate e-mail being sent, which can also cause unexpected behaviour in tests. See `AuditActivity::Test::TestResultUpdated` for an example of this.

app/models/test.rb
```diff
@@ -1,17 +1,8 @@
...
-  include DateConcern
-  include SanitizationHelper
-
   belongs_to :investigation
-  belongs_to :product
-
-  has_many_attached :documents
-
-  date_attribute :date
+  belongs_to :product, optional: false
```
slorek, line 3: Please check on production whether there is any data which might cause a break with this change.

app/views/investigations/test_results/_form.html.erb
slorek: I think we agreed to move most strings to I18n over time, not just models. This would be ideal, but not a big deal.

db/migrate/20201125105949_add_standars_product_was_testing_against_to_tests.rb
```diff
@@ -0,0 +1,8 @@
+class AddStandarsProductWasTestingAgainstToTests < ActiveRecord::Migration[6.0]
```
slorek, line 1: Please change: typos:
```suggestion
class AddStandardsProductWasTestedAgainstToTests < ActiveRecord::Migration[6.0]
```

spec/fixtures/files/new_test_result.txt
```diff
@@ -0,0 +1 @@
+new test ressult
```
slorek, line 1:
```suggestion
new test result
```

slorek (changes requested):
Some minor changes needed from my first scan of the code. I'll test the review app next.

app/models/test.rb
StupidCodeFactory: I've checked with:
```ruby
 Test::Result.left_joins(:product).where(products: {id: nil}).count # => 0
```

so it should be fine

app/views/investigations/test_results/_form.html.erb
StupidCodeFactory: ok I wasn't aware of this! This Pr is already quite big, I'd like to avoid adding more moving part to this PR

.slim-lint.yml
```diff
@@ -24,6 +24,8 @@ linters:
       - 'app/views/layouts/_header.html.slim'
       - 'app/views/layouts/_footer.html.slim'
       - 'app/views/layouts/application.html.slim'
+      - 'app/views/investigations/creation_flow/allegation_details.html.slim'
+      - 'app/views/investigations/creation_flow/enquiry_details.html.slim'
```
StupidCodeFactory, line 28: Did not want to add noise to this big Pr. will do a follow up Pr to convert this templates and remove this from the slim linter

KyleMacPherson (approved):

slorek (changes requested):
Please change: looks like the wrong class was used for the field label compared to the others.

Screenshot 2020-12-14 at 14 58 21

On edit (but not add), I can add multiple "blank" test standards, e.g: 

Screenshot 2020-12-14 at 15 00 56

slorek (approved):

tobyhughes:
Very minor change, need to update hint text: 


image

Should be: `For example, EN71. Use a comma to separate multiple standards.`
 (it's just missing the 'a')

tobyhughes:
Activity record is looking a little strange 

image

Should reflect list of fields on form:

`Product:`
`Legislation:`
`Standards:`
`Test date:`
`Test result:`
`Further details:`
`Attached:` 

Then links to view the product and the test result (see corrective action activity log)

StupidCodeFactory:
@tobyhughes could you have another look I've fixed the audit.
Screenshot 2020-12-16 at 14 40 08

nasirkhanbeis:
Hi @StupidCodeFactory  hint text is slightly wrong,'a' in wrong place. :) 
It should read as "Use a comma to separate multiple standards " but not Use comma a to separate multiple standards

nasirkhanbeis:
Audit log looks good 
image

tobyhughes (approved):
Looks good to me!

PR #1053, opened 2020-11-26
Fix images not being attached to products on TS case creation flow on hosted environments

slorek:
This fixes a  bug reported by users when the application is running on production.

I was able to reproduce the error on staging and on review apps. It was not possible to reproduce in development or test modes locally or in CI.

The problem was that the product is instantiated as a child of `@investigation` and assigned to `@product` before it is saved in a later call to `CreateCase`, which calls `save` on the parent Investigation. See https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/app/controllers/investigations/ts_investigations_controller.rb#L615-L616. The product is therefore saved as well, but this state is not automatically updated in the controller.

Later, when images are attached to the product, it defers the saving of the attachments until the product is saved. This behaviour is different to when the product is already saved, in which case the attachments are saved immediately. That's the behaviour this code expects.

There is clearly some difference in behaviour between `development` and `production` modes related to tracking the state of associated ActiveRecord objects when the parent is saved. This is also clearly a regression since it must have worked before. It may have been introduced in the Rails 6 upgrade.

Consequently although I have written a test (and this does improve the test quality a bit) it doesn't really provide us with confidence that there will not be a regression in future. I'm open to suggestions about how we can improve this.

### DIFF
--- a/app/controllers/investigations/ts_investigations_controller.rb
+++ b/app/controllers/investigations/ts_investigations_controller.rb
@@ -697,9 +697,12 @@ private
   end
 
   def save_product_files
+    # Obtain updated product state before attaching files to ensure changes are saved immediately
+    @product.reload
+
     session[:product_files].each do |file_blob_id|
       file_blob = ActiveStorage::Blob.find_by(id: file_blob_id)
-      attach_blobs_to_list(file_blob, @product.documents)
+      @product.documents.attach(file_blob)
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/controllers/investigations/ts_investigations_controller.rb
```diff
@@ -697,9 +697,12 @@ def save_files
   end
 
   def save_product_files
+    # Obtain updated product state before attaching files to ensure changes are saved immediately
+    @product.reload
+
     session[:product_files].each do |file_blob_id|
       file_blob = ActiveStorage::Blob.find_by(id: file_blob_id)
-      attach_blobs_to_list(file_blob, @product.documents)
+      @product.documents.attach(file_blob)
```
slorek, line 705: This is just a simplification of the existing code.

StupidCodeFactory (approved):

nasirkhanbeis (approved):
looks good to me!

PR #1055, opened 2020-11-27
Fix total search results count

slorek:
https://trello.com/c/AtLllGyL/591-count-of-search-results-all-tabs-is-wrong

Fixes the total search result count shown in the top left of the cases, businesses and products pages not reflecting the total number of matches. Instead it was showing the number of results on the current page.

Also fixes the count not being output on the product search results page.

### DIFF
--- a/app/views/businesses/index.html.erb
+++ b/app/views/businesses/index.html.erb
@@ -12,7 +12,7 @@
       <% if params[:q].present? %>
         <%= params[:q] %>
         <span class="govuk-caption-m">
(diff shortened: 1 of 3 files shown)


KyleMacPherson (approved):

nasirkhanbeis (commented):
Hi @slorek  let me have quick look in review app

nasirkhanbeis (approved):
looks good! (there's some indexing issue on review apps due to which count is not 100% accurate) however should be good on stage and prod!

PR #1078, opened 2020-12-15
Edit risk validation

KyleMacPherson:
https://trello.com/c/6jbpjhvr/843-remove-validation-status

## Description
- Allow editing of the risk validation level.
- Allow adding of a reason for when a validation is removed which is added to the activity log.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/controllers/investigations/risk_validations_controller.rb
+++ b/app/controllers/investigations/risk_validations_controller.rb
@@ -27,8 +27,7 @@ module Investigations
 
     def edit
       @investigation = Investigation.find_by(pretty_id: params["investigation_pretty_id"])
-      is_risk_validated = @investigation.risk_validated_by ? true : false
-      @risk_validation_form = RiskValidationForm.new(is_risk_validated: is_risk_validated)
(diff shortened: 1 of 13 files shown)


nasirkhanbeis:
looks good to me, just minor pre-select issue and a major permissions stuff pending still.

nasirkhanbeis (approved):
Looks good as expected. 
Note: as agreed, validate case role is added to user level(not team) so user with this permission can  add/edit validate on any case

tobyhughes:
Think we need to work on the question page a bit, we need to include a text area rather than text input, see:

image

tobyhughes:
Looks like the text input is persisting even if the validation status is already `No` ?

We should only show the text input if the validation status is being changed from validated

tobyhughes:
can we update the back link to reflect what we have elsewhere?

(Think it is `Back to {case type}: {case id}`)

PR #1081, opened 2020-12-16
Question to determine if the product has a UKCA mark, a UKNI mark, or a CE mark

slorek:
https://trello.com/c/AUGFmp9W/852-question-to-determine-if-the-product-has-a-ukca-mark-a-ukni-mark-or-a-ce-mark

## Description
I've gone with an Array column type which is proprietary to PostgreSQL. If this app was ever to be ported to another RDMS we would need to implement Rails' serialization.

I had to change the design slightly to use a `legend` tag instead of hint text, for accessibility.

Unfortunately we need to continue supporting `nil` values for existing data, so we can't implement validation or not allowing null at the model or database level.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


app/helpers/products_helper.rb
```diff
@@ -6,8 +6,8 @@ module ProductsHelper
   # Never trust parameters from the scary internet, only allow the white list through.
   def product_params
     params.require(:product).permit(
-      :brand, :name, :subcategory, :category, :product_code, :webpage, :description, :batch_number, :country_of_origin, :gtin13, :authenticity, :affected_units_status, :number_of_affected_units, :exact_units, :approx_units
-    )
+      :brand, :name, :subcategory, :category, :product_code, :webpage, :description, :batch_number, :country_of_origin, :gtin13, :authenticity, :affected_units_status, :number_of_affected_units, :exact_units, :approx_units, :has_markings, markings: []
+    ).with_defaults(markings: [])
```
StupidCodeFactory, line 10: Nice! Didn't know about this nifty trick!

StupidCodeFactory (approved):
Looks good!

slorek:
Changes to the design required prior to merging.

KyleMacPherson (approved):

nasirkhanbeis:
Outstanding issues to resolve 1. regression bug around unit counts value not shown- https://trello.com/c/XrIJAjtm/863-regression-issue-in-pr-1081-units-count-is-not-saved

nasirkhanbeis:
re-tested regression issue, looking good!
Unknown is still showing as "Not provided" either we need to show Unknown or blank

app/decorators/product_decorator.rb
```diff
@@ -54,4 +55,12 @@ def units_affected
       I18n.t(".product.not_provided")
     end
   end
+
+  def markings
+    return I18n.t(".product.not_provided") unless object.has_markings
+    return I18n.t(".product.unknown") if object.has_markings == "markings_unknown"
```
StupidCodeFactory, line 61:
```suggestion
    return I18n.t(".product.unknown") if object.markings_unknown?
```

app/decorators/product_decorator.rb
```diff
@@ -54,4 +55,12 @@ def units_affected
...
+    return I18n.t(".product.none") if object.has_markings == "markings_no"
```
StupidCodeFactory, line 62:
```suggestion
    return I18n.t(".product.none") if object.markings_no?
```

app/helpers/products_helper.rb
StupidCodeFactory: maybe make this a `Set.new` ?

app/models/product.rb
```diff
@@ -28,4 +36,8 @@ class Product < ApplicationRecord
   has_many :tests, dependent: :destroy
 
   has_one :source, as: :sourceable, dependent: :destroy
+
+  def has_markings?
```
StupidCodeFactory, line 40: I can not see where this is used? Is it redundant?

db/schema.rb
```diff
@@ -19,6 +19,7 @@
   # These are custom enum types that must be created before they can be used in the schema definition
   create_enum "affected_units_statuses", ["exact", "approx", "unknown", "not_relevant"]
   create_enum "authenticities", ["counterfeit", "genuine", "unsure"]
+  create_enum "has_markings_values", ["markings_yes", "markings_no", "markings_unknown"]
```
StupidCodeFactory, line 22: I think you forgot to commit migration timespamp

spec/forms/product_form_spec.rb
```diff
@@ -110,4 +110,84 @@
       end
     end
   end
+
+  describe "#markings=" do
+    context "when the value supplied is nil" do
+      it "sets the value to an empty array" do
+        form.markings = nil
+        expect(form.attributes["markings"]).to eq([])
```
StupidCodeFactory, line 118:
```suggestion
        expect(form).to have_attributes(markings: [])
```

spec/models/product_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+RSpec.describe Product do
+  subject(:product) { build(:product, has_markings: has_markings) }
+
+  describe "#has_markings?" do
+    context "when has_markings == markings_yes" do
+      let(:has_markings) { "markings_yes" }
+
+      it "returns true" do
+        expect(product.has_markings?).to be true
```
StupidCodeFactory, line 11:
```suggestion
        expect(product).to have_markings
```

StupidCodeFactory (changes requested):
Looks good I've suggested a few improvements but I think you are missing the timestamps of the migration at the top of the `schema.rb` file

nasirkhanbeis (approved):
looks good as expected!

slorek:
> Looks good I've suggested a few improvements but I think you are missing the timestamps of the migration at the top of the `schema.rb` file

@StupidCodeFactory there was a newer hotfix migration added and merged to master by you so the schema version is correct. The migration will still run.

app/models/product.rb
slorek: Yes looks like it is now redundant.

app/helpers/products_helper.rb
slorek: I considered using a `Set` to avoid the de-duplication code, but the database data type is `Array` and that would be the type returned by ActiveRecord. We would depend on the casting when creating a new record. This could be confusing and have unexpected consequences.

db/schema.rb
slorek: There was a newer hotfix migration added and merged to master by you so the schema version is correct. The migration will still run.

StupidCodeFactory (approved):

spec/forms/product_form_spec.rb
slorek: `have_attributes` uses the getter method `markings` which has different behaviour we do not want to test in this instance.

PR #1090, opened 2020-12-21
add online recall information

StupidCodeFactory:
## Description

1. Merge both Corrrective action controller and create a form to harmonise the
application code. 
2. Removed the confirmation page in the flow and got rid of  `Wicked::Wizard`. 
3. Added the missing form object 
4. remove old audit log created by active record callback (removing on more `User.current`)

Also fixed the ordering of errors on the edit corrective action's page. 

## testing
@nasirkhanbeis you should be able to use  [this corrective action](https://psd-pr-1090.london.cloudapps.digital/cases/2101-0001/corrective-actions/1/edit) to test legacy corrective action without online recall information previously given.
With your account in `OPSS Operational support unit`

I've refactored both the `create` and `update` so it would be prudent to test these parts  thoroughly.

Another point to have a close look at is whether the old corrective action's document still show up correctly. I've applied the same fix to corrective action then last time with test results. This time I did the migration that should prevent this problem from happening again.

## Checklist:
- [X] Has the CHANGELOG been updated? (If change is worth telling users about.)


nasirkhanbeis:
spotted few minor issues
1. for existing  corrective actions, we should show new field with value "Not provided"
2. when I just update recall information only, activity log doesn't seem to record new changes.

nasirkhanbeis:
One more issue, corrective action in TS flow is not recored. steps to replicate
1. As TS user, open a new case, fill in all the details, follow the journey.
2. When prompted for corrective action, select yes and fill-in corrective action form, are there other actions to report? don't select any option, submit the form. Expected to throw validation error however 
  it creates case successfully with no corrective action 
image

nasirkhanbeis:
re-call information field in view page to be placed after business as given in the miro board
image

app/controllers/investigations/corrective_actions_controller.rb
```diff
@@ -11,49 +38,40 @@ def show
...
+      @corrective_action_form.assign_attributes(corrective_action_params)
+
+      return render :edit if @corrective_action_form.invalid?
 
       result = UpdateCorrectiveAction.call(
-        corrective_action: corrective_action,
-        corrective_action_params: corrective_action_params,
-        user: current_user
+        @corrective_action_form
+          .serializable_hash(except: :further_corrctive_action)
```
StupidCodeFactory, line 64:
```suggestion
          .serializable_hash
```

app/models/audit_activity/corrective_action/base.rb
```diff
@@ -5,33 +5,23 @@ class AuditActivity::CorrectiveAction::Base < AuditActivity::Base
   belongs_to :business, optional: true, class_name: "::Business"
   belongs_to :product, class_name: "::Product"
 
-  private_class_method def self.from(corrective_action)
```
StupidCodeFactory, line 8: through exception

app/models/audit_activity/corrective_action/base.rb
```diff
@@ -5,33 +5,23 @@ class AuditActivity::CorrectiveAction::Base < AuditActivity::Base
...
-    body += "Duration of action: **#{CorrectiveAction.human_attribute_name("duration.#{corrective_action.duration}")}**<br>"
-    body += "Geographic scope: **#{corrective_action.geographic_scope}**<br>"
-    body += "Attached: **#{sanitize_text corrective_action.documents.first.filename}**<br>" if corrective_action.documents.attached?
-    body += "<br>#{sanitize_text corrective_action.details}" if corrective_action.details.present?
-    body
+  def corrective_action
+    @corrective_action ||= begin
+                             if metadata&.dig("corrective_action_id")
+                               CorrectiveAction.find_by!(id: metadata["corrective_action_id"])
+                             elsif attachment.attached?
```
StupidCodeFactory, line 12: wrap this in a method `load_legacy_audit_corrective_record`

nasirkhanbeis:
Hi @StupidCodeFactory , there's a bug in activity log, view product and view corrective action links are missing 
Add new corrective action and go to activity log 
image

nasirkhanbeis:
Another issue - Edit existing corrective action, returns error as below when I try save with recall information (any option yes, no or Not relevant returns error). Expected to save with recall information
image

tobyhughes:
No capitalisation required on 'relevant' :

image

nasirkhanbeis:
Few more issues to sort on latest build 
1. recall information is missing  in activity log when new corrective action is added
2. View product details link shows with leading character '4' in activity log
3. when recall information is chosen 'No' then show value "Not published" in view page and activity log.

nasirkhanbeis:
hey @StupidCodeFactory 
Found a potential issue around activity log, when corrective action is updated, along with changes log, it updates original activity log too.

app/models/audit_activity/corrective_action/add.rb
```diff
@@ -1,26 +1,14 @@
 class AuditActivity::CorrectiveAction::Add < AuditActivity::CorrectiveAction::Base
-  def self.from(corrective_action)
-    super(corrective_action)
+  def self.build_metadata(corrective_action, changes)
+    { corrective_action_id: corrective_action.id, updates: changes }
```
slorek, line 3: Suggestion: I'm also looking at a pattern for achieving this separately. It seems a bit strange to pass all the new attributes from the controller to the service alongside a `changes` object which is relayed to the audit activity. Especially as it is a new object so there can be no updates. Since we receive the `corrective_action` we could just serialize the attributes here and remove all the `changes`.

nasirkhanbeis:
Re-tested latest changes on review app 1090 and spotted couple issues
1. Attachment file when replaced with new file, it replaces in original add corrective action log too.
2. When attachment description is removed, it doesn't records in activity log but shows empty log 
image

nasirkhanbeis:
Another issue, when corrective action added in TS flow, view business link in activity log is missing(when business was added in the TS flow)

nasirkhanbeis (approved):
looking good as expected!

app/controllers/investigations/corrective_actions_controller.rb
```diff
@@ -1,8 +1,35 @@
...
+
+      @corrective_action_form = CorrectiveActionForm.new(corrective_action_params)
+
+      @investigation = investigation.decorate
+      return render :new if @corrective_action_form.invalid?
+
+      result = AddCorrectiveActionToCase.call(
+        @corrective_action_form
+          .serializable_hash(except: :further_corrctive_action)
+          .merge(user: current_user, investigation: investigation, changes: @corrective_action_form.changes)
```
slorek, line 24: We've discussed this passing of the `changes` hash through the service to the audit activity as something we want to avoid. We have a proposal for an alternative pattern in a separate PR (https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/1144). It is not necessary to change this now, but it will be a little more difficult to do so afterwards if the feature is used in production, as we'll have to write a data migration. I'll leave it to you to decide whether to refactor this now or afterwards.

app/helpers/investigations/corrective_actions_helper.rb
```diff
@@ -1,20 +1,51 @@
 module Investigations
   module CorrectiveActionsHelper
+    TRUNCATED_ACTION_MAP = {
```
slorek, line 3: Please change: This appears to be a duplicate of `CorrectiveAction::TRUNCATED_ACTION_MAP`. Also I think it would ideally be achieved with I18n.

slorek (changes requested):
Minor changes requested in comments.

KyleMacPherson (approved):

slorek (approved):

nasirkhanbeis:
Re-tested latest fix, looking good as expected.

PR #1100, opened 2020-12-30
Team roles

slorek:
https://trello.com/c/mdnO2nMb/860-5-team-roles

## Description
Allows users to inherit the roles of their team.

This renders the `opss_user` role and the juggling of it for new users redundant, as we can set the role on the OPSS teams instead, and users will inherit this automatically. I have added a migration for this.

I have also migrated the `risk_level_validator` role to the relevant team. Currently we only have one team with this role, and all of their users have this role.

### Deployment

To be deployed out of hours due to changes to the user roles data and schema which might temporarily affect active users for a short time during deployment.

## Checklist:

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -240,24 +240,23 @@ RSpec.describe User do
   end
 
   describe "#has_role?" do
-    subject(:result) { user.has_role?(:test) }
+    subject(:user) { create(:user, team: team, organisation: team.organisation, roles: user_roles) }
 
     let(:team) { create(:team, roles: team_roles) }
-    let(:user) { create(:user, team: team, organisation: team.organisation, roles: user_roles) }
     let(:team_roles) { [] }
 
     context "when the user has no roles" do
       let(:user_roles) { [] }
 
       it "returns false" do
-        expect(result).to be false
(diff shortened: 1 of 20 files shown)


spec/models/user_spec.rb
```diff
@@ -238,4 +238,70 @@
       expect(user).not_to be_mobile_number_change_allowed
     end
   end
+
+  describe "#has_role?" do
+    subject(:result) { user.has_role?(:test) }
```
StupidCodeFactory, line 243: Suggestion: 
I'd remove this, it confusingly changes the subject of the file which is supposed to be an instance of a `User` and not the result of the method `#has_role?` of the subject that is currently under test. Then below in the test body:
```ruby
it { is_expected.to have_role :test }
it { is_expected.not_to have_role :test }
```
This in my opinion reads better and does not change the value of `subject` throughout the file.

StupidCodeFactory (approved):

spec/models/user_spec.rb
slorek: 👍 there seems to be a bug in `super_diff` with the `has` matcher using that syntax. I’ve implemented a compromise solution.

nasirkhanbeis (approved):
looking good!

PR #1144, opened 2021-01-26
Store test result metadata on audit log on creation

slorek:
https://trello.com/c/CGFTJSbk/892-regression-bug-supporting-information-original-log-activity-being-over-written-when-any-changes-made

## Description
Provides a fix for test result data not being preserved on the audit log.

Does not address phone call correspondence, but there are no instances of these being updated on production. This can be fixed relatively easily in a follow-up PR once the conventions introduced here are agreed.

I have taken the opportunity to make the field labelling consistent across added and updated audit activity ("Date of test"/"Test date" for example).

I have tested the migration against some local test data with several edits. Note the migration only works because the `ActiveStorage::PurgeJob` queue is not being worked. This is a separate bug we will need to fix later, although the queue size is fairly small for now. I have checked the staging and production databases that all files needed are present, so confidence is good that the migration will be successful.

Introduces a new pattern for storing metadata about an object on creation. Rather than passing a `changes` object from 
 the form through the service to the model, which ties both the service and model to the front-end UI, this proposal just stores the attributes of the mode lat the point they were saved. This removes the need to use a form object to communicate with the service object, simplifying the interface and allowing it to be used in other contexts (such as the seeds file, where it is used currently).

Should this proposal be accepted, we would expect to adopt this pattern going forward and to refactor existing usage to be consistent.

Removed `RemoveFilesWithoutAttachmentsJob` because this is not desired behaviour (we need orphaned attachments when referenced in audit activity), and doesn't seem to work.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


KyleMacPherson:
@slorek could this be a sidekiq issue? I don't know much about this but active storage seems to enqueue purge jobs and we seem to have the queue for this set to nil in `config/initializers/new_framework_defaults_6_1.rb`. I am guessing that we stub sidekiq in our tests and maybe this is where the discrepancy comes in?

slorek:
> @slorek could this be a sidekiq issue? I don't know too much about this but I believe active storage enqueues purge jobs and we seem to have the queue for this set to nil in `config/initializers/new_framework_defaults_6_1.rb`. I assume that we stub sidekiq in our tests and perhaps this is where the discrepancy comes in? Would be interesting to check sidekiq and see if the purge jobs are ever enqueued/successfully run.

@KyleMacPherson you're right, it's due to the different queue adapters on the service spec vs the feature spec. Thanks for the pointer! 🙌

app/decorators/audit_activity/test/result_decorator.rb
```diff
@@ -0,0 +1,12 @@
+class AuditActivity::Test::ResultDecorator < ApplicationDecorator
+  delegate_all
+
+  def title(_viewing_user = nil)
+    return super if metadata.nil?
+
+    # The decorator provides logic to create the title but it may have changed
+    # in subsequent edits, so we need to create a new instance using the
+    # original attributes in order to decorate it
+    Test::Result.new(metadata["test_result"].except("document")).decorate.title
```
slorek, line 10: This might break if we start removing attributes in the future, but for now it works.

app/models/audit_activity/test/result.rb
```diff
@@ -1,28 +1,48 @@
...
+    metadata["test_result"]["standards_product_was_tested_against"]
+  end
+
+  def result
+    metadata["test_result"]["result"]
+  end
+
+  def details
+    metadata["test_result"]["details"]
+  end
```
StupidCodeFactory, line 40: Suggestion: this methods have typically been added to the decorators since they're only used in views.

app/views/investigations/activities/test/_result.html.erb
```diff
@@ -1,18 +1,18 @@
...
+  Legislation: <strong><%= activity.legislation %></strong>
   <br />
-  Standards: <strong><%= activity.test_result.standards_product_was_tested_against.join(", ") %></strong>
+  Standards: <strong><%= activity.standards_product_was_tested_against.join(", ") %></strong>
   <br />
-  Test date: <strong><%= activity.test_result.date.to_s(:govuk) %></strong>
+  Date of test: <strong><%= activity.date.to_s(:govuk) %></strong>
   <br />
-  Test result: <strong><%= activity.test_result.result.titleize %></strong>
+  Result: <strong><%= activity.result.titleize %></strong>
```
StupidCodeFactory, line 10: Suggestion: move `.to_s(:govuk), .join(", "), .titleize` to decorator

app/views/investigations/activities/test/_test_result_updated.html.erb
```diff
@@ -3,35 +3,35 @@
...
   <% end %>
 
   <% if activity.new_legislation %>
-    Legislation: <b><%= activity.new_legislation %></b><br>
+    Legislation: <strong><%= activity.new_legislation %></strong><br>
   <% end %>
 
   <% if activity.new_standards_product_was_tested_against %>
-    Standards: <b><%= activity.new_standards_product_was_tested_against.join(", ") %></b><br>
+    Standards: <strong><%= activity.new_standards_product_was_tested_against.join(", ") %></strong><br>
```
StupidCodeFactory, line 22: Suggestion: move `.join(", ")` to decorator

StupidCodeFactory (approved):

app/models/audit_activity/test/result.rb
slorek: I moved the `title` method to a new decorator but I think these others are fine in the model as they are just convenience accessors to existing data. Makes sense to move transformations to the decorator though.

nasirkhanbeis:
spotted few minor issues otherwise looks good
Edit test results - remove further details doesn't show in audit log correctly
image
When attachment description removed then audit log to read as Attachment description: removed 
image
When same file is attached then I see empty log created 
image

slorek:
@nasirkhanbeis these are existing bugs unrelated to this change. We will manage these fixes independently.

PR #1157, opened 2021-02-02
Fix corrective action add audit

StupidCodeFactory:
Migrate old activities to use the metadata hash. 

Fixes this issue: https://sentry.io/organizations/beis/issues/2193557658/?referrer=slack 

## notes:

Will need to follow up with a PR to remove revert this change once on prod


app/models/audit_activity/corrective_action/add.rb
```diff
@@ -3,6 +3,34 @@ def self.build_metadata(corrective_action)
     { corrective_action: corrective_action.attributes, document: corrective_action.document_blob&.attributes }
   end
 
+  def self.metadata_from_legacy_audit_activity(audit_activity)
+    attributes = audit_activity.body.split("<br>").each_with_object({}) do |fragment, corrective_action_attributes|
+      next if fragment.empty?
+
+      case fragment
+      when /Legislation: \*\*(.*)\*\*/
```
StupidCodeFactory, line 11: my first ascii art!

slorek:
```suggestion
      when /\ALegislation: \*\*(.*)\*\*\z/
```
Ensures the match is for the full fragment rather than within a larger string.

app/models/audit_activity/corrective_action/add.rb
```diff
@@ -3,6 +3,34 @@ def self.build_metadata(corrective_action)
...
+      case fragment
+      when /\ALegislation: \*\*(.*)\*\*\z/
+        corrective_action_attributes[:legislation] = Regexp.last_match(1)
+      when /Date came into effect: \*\*(.*)\*\*/
+        corrective_action_attributes[:decided_date] = Regexp.last_match(1)
+      when /Type of measure: \*\*(.*)\*\*/
+        corrective_action_attributes[:measure] = Regexp.last_match(1)
+      when /Duration of action: \*\*(.*)\*\*/
+        corrective_action_attributes[:duration] = Regexp.last_match(1)
+      when /Geographic scopes: \*\*(.*)\*\*/
```
slorek, line 19: There is a lot of data on prod where the title is `Geographic scope` rather than `Geographic scopes`. Is this a typo or do we need to catch both?

app/models/audit_activity/corrective_action/add.rb
```diff
@@ -3,6 +3,34 @@ def self.build_metadata(corrective_action)
...
+        corrective_action_attributes[:geographic_scopes] = Regexp.last_match(1)
+      when /Attached|Product/
+        next
+      else
```
slorek, line 23: Missing matches for `Business responsible: .*` and `Risk: .*`

app/models/audit_activity/corrective_action/add.rb
```diff
@@ -3,6 +3,34 @@ def self.build_metadata(corrective_action)
...
+        corrective_action_attributes[:details] = fragment
```
slorek, line 24: Suggestion: `trim` the fragment as there are a lot of empty lines.

app/models/audit_activity/corrective_action/add.rb
StupidCodeFactory: nope it's a typo :)

app/models/audit_activity/corrective_action/add.rb
StupidCodeFactory: good spot

StupidCodeFactory: @slorek after investigation `Risk:` is included in the `details` fields and is anyway not related to the corrective action model, we have the `risk_assessments` for this so non need to take action on either Business nor risk

app/models/audit_activity/corrective_action/add.rb
```diff
@@ -3,6 +3,34 @@ def self.build_metadata(corrective_action)
...
+      when /\AType of measure: \*\*(.*)\*\*\z/
+        corrective_action_attributes[:measure] = Regexp.last_match(1)
+      when /\ADuration of action: \*\*(.*)\*\*\z/
+        corrective_action_attributes[:duration] = Regexp.last_match(1)
+      when /\AGeographic scope: \*\*(.*)\*\*\z/
+        corrective_action_attributes[:geographic_scopes] = Regexp.last_match(1)
+      when /\AAttached|\AProduct|\ABusiness responsible:/
+        next
+      else
+        corrective_action_attributes[:details] ||= fragment.strip.presence
```
slorek, line 24: This code would replace `corrective_action_attributes[:details]` on each matching line. I think we need to append, as there can be multiple lines.

```
  corrective_action_attributes[:details] = []
  
  ...
  
  corrective_action_attributes[:details] << fragment.strip.presence
  
  ...
  
  corrective_action_attributes[:details].join("\n")
```

slorek (approved):

PR #1182, opened 2021-02-12
Allow export of products as CSV

slorek:
https://trello.com/c/hq74CuGh/901-csv-export-of-product-data
https://trello.com/c/5eW7fDI2/872-add-product-details-export
https://trello.com/c/2x6xL2gG/228-allow-users-to-export-product-information

## Description
I decided to make this a re-usable feature rather than a one-off, since this has been on the backlog for a while and didn't take long, and there is a need to do this regularly.

I decided to export products as a CSV rather than in Excel format as elsewhere, since CSV is more interoperable, simpler, and we are not using any specific features. The implementation used here will allow an easy switchover to CSV for the cases export in a future piece of work.

As with export of cases, if a user has the `psd_admin` role they will see a link at the bottom of the products listing which will allow them to download the CSV file.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/spec/policies/product_policy_spec.rb
+++ b/spec/policies/product_policy_spec.rb
@@ -10,13 +10,13 @@ RSpec.describe ProductPolicy do
       before { user.roles.create!(name: "psd_admin") }
 
       it "returns true" do
-        expect(policy.export?).to be true
+        expect(policy).to be_export
       end
     end
 
     context "when the user does not have the psd_admin role" do
       it "returns false" do
-        expect(policy.export?).to be false
+        expect(policy).not_to be_export
       end
     end
   end
--- a/spec/policies/product_policy_spec.rb
+++ b/spec/policies/product_policy_spec.rb
@@ -1,0 +1,23 @@
+require "rails_helper"
+
+RSpec.describe ProductPolicy do
+  subject(:policy) { described_class.new(user, Product) }
+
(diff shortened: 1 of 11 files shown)


spec/policies/product_policy_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  subject(:policy) { described_class.new(user, Product) }
+
+  let(:user) { create(:user) }
+
+  describe "#export?" do
+    context "when the user has the psd_admin role" do
+      before { user.roles.create!(name: "psd_admin") }
+
+      it "returns true" do
+        expect(policy.export?).to be true
```
StupidCodeFactory, line 13:
```suggestion
        expect(policy).to be_export
```

spec/policies/product_policy_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      end
+    end
+
+    context "when the user does not have the psd_admin role" do
+      it "returns false" do
+        expect(policy.export?).to be false
```
StupidCodeFactory, line 19:
```suggestion
        expect(policy).not_to be_export
```

StupidCodeFactory (approved):

PR #1184, opened 2021-02-15
Fix audit activity not storing phone call metadata at time of creation

slorek:
https://trello.com/c/FoYTaDpb/913-phone-call-correspondence-audit-activity-not-storing-metadata-at-time-of-creation

## Description
Fix for phone call metadata following conventions introduced in #1144 

There are no phone call updated activities in production so it is not necessary to rewind changes - the correspondence object is still the same state as when created.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


app/decorators/audit_activity/correspondence/add_phone_call_decorator.rb
```diff
@@ -1,27 +1,21 @@
 module AuditActivity
   module Correspondence
     class AddPhoneCallDecorator < AuditActivity::Correspondence::BaseDecorator
-      def phone_call_by(viewing_user)
```
slorek, line 4: This was a redundant method.

app/models/audit_activity/correspondence/add_phone_call.rb
```diff
@@ -3,12 +3,40 @@ class AuditActivity::Correspondence::AddPhoneCall < AuditActivity::Correspondenc
...
+    metadata["correspondence"]["phone_number"]
+  end
+
+  def filename
+    metadata["correspondence"]["transcript"]&.fetch("filename")
+  end
+
+  def details
+    metadata["correspondence"]["details"]
   end
```
StupidCodeFactory, line 41: Suggestion: move this presentational methods to presentational layer

StupidCodeFactory (approved):

app/models/audit_activity/correspondence/add_phone_call.rb
slorek: Thanks for the comment. As per https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/1144#discussion_r567772446 I don't think these methods are presentational, they're convenience accessors to data. Any presentational transformation can be added to decorators.

PR #1185, opened 2021-02-16
Resend link invite bug

KyleMacPherson:
https://trello.com/c/8cqGiWhN/521-resend-link-invite-should-not-displayed-once-account-set-up-is-complete

## Description
Before this change, if a user received an email invite and entered their details, but signed out before accepting the declaration then the inviting user had the option to resend an email invitation. This is not what we want as it could confuse the user, and possibly cause them to sign up multiple accounts.

This change only shows the 'resend invitation' button if the user has not yet visited the site and entered in their necessary details. 

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -149,7 +149,7 @@ class User < ApplicationRecord
   end
 
   def has_filled_out_account_setup_form?
-    !name.blank? && !email.blank?
+    name.present? && email.present?
   end
 
 private
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -148,6 +148,10 @@ class User < ApplicationRecord
     !mobile_number_verified?
   end
 
+  def has_filled_out_account_setup_form?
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -148,6 +148,10 @@ def mobile_number_change_allowed?
     !mobile_number_verified?
   end
 
+  def has_filled_out_account_setup_form?
+    name.present? && mobile_number.present?
```
slorek, line 152: 🤔 I think we would want to use `mobile_number_verified?` instead because this will be set when the 2FA has been completed. Otherwise the mobile number may be entered but not verified. What do you think?

slorek (approved):

nasirkhanbeis (approved):
looks good as expected

PR #1194, opened 2021-02-22
Allow export of businesses

slorek:
https://trello.com/c/7YYl8CaJ/871-add-business-details-export

## Description
Same behaviour as product and case export - if you have the `psd_admin` role you will see an additional link to the CSV file download on the list page.

Business type is derived from associated investigations. A business may have more than one type if they are added to a case in different contexts.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


KyleMacPherson (approved):

app/decorators/businesses_decorator.rb
```diff
@@ -1,3 +1,20 @@
...
+  end
+
+private
+
+  def attributes_for_export
+    Business.attribute_names.dup.concat(%w[types case_ids primary_location_address_line_1 primary_location_address_line_2 primary_location_city primary_location_country primary_location_county primary_location_postal_code primary_location_phone_number primary_contact_email primary_contact_name primary_contact_phone_number primary_contact_job_title]).sort.freeze
+  end
+
+  def export_attribute_values_for_business(business)
+    attributes_for_export.map { |key| business.send(key) }
```
StupidCodeFactory, line 18:
```suggestion
    attributes_for_export.map { |key| business.public_send(key) }
```

app/decorators/businesses_decorator.rb
```diff
@@ -1,3 +1,20 @@
 class BusinessesDecorator < Draper::CollectionDecorator
   delegate :current_page, :total_entries, :total_pages, :per_page, :offset
+
+  def to_csv
+    CSV.generate do |csv|
+      csv << attributes_for_export
+      each { |business| csv << export_attribute_values_for_business(business) }
```
StupidCodeFactory, line 7: I'm confuse to where this `each(&:block)` work? 
seem the decorator itself is the message receiver?

KyleMacPherson (approved):

StupidCodeFactory (approved):

app/decorators/businesses_decorator.rb
slorek: @StupidCodeFactory yes that's correct. It will be delegated to the underlying `object` which is an `Array`.

PR #1197, opened 2021-02-23
migrate non array geographic scopes to array

StupidCodeFactory:
Fix  this sentry error https://sentry.io/organizations/beis/issues/2226867247/?project=1329381&referrer=slack

### DIFF
--- a/db/migrate/20210223125626_fix_corrective_actions_added_geographic_scopes.rb
+++ b/db/migrate/20210223125626_fix_corrective_actions_added_geographic_scopes.rb
@@ -1,5 +1,4 @@
 class FixCorrectiveActionsAddedGeographicScopes < ActiveRecord::Migration[6.1]
-
   def change
     safety_assured do
       reversible do |dir|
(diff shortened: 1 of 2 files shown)


slorek (approved):
This is a partial fix but I also think some of the values don't match those in `CorrectiveAction::GEOGRAPHIC_SCOPES` exactly. This could be migrated separately however.

StupidCodeFactory:
> This is a partial fix but I also think some of the values don't match those in `CorrectiveAction::GEOGRAPHIC_SCOPES` exactly. This could be migrated separately however.

```ruby
  GEOGRAPHIC_SCOPES_MIGRATION_MAP = {
    "Local" => %w[local],
    "Regional" => %w[great_britain northern_ireland],
    "National" => %w[great_britain northern_ireland],
    "EEA wide" => %w[eea_wide],
    "EU wide" => %w[eu_wide],
    "Unknown" => %w[unknown],
    nil => %w[unknown]
  }.freeze
```
What values are you thinking of? Here the keys are the values that where in  `confing/constants/corrective_action_constants.yml`

PR #1201, opened 2021-02-24
Add case_ids to product export

KyleMacPherson:
https://trello.com/c/iRe6TLRB/935-add-caseid-to-product-export

## Description

Add case ids to product export

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


slorek:
Let's take time to align the approach with https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/1194. I stuck the list of extra attributes in the model, but I think I prefer your approach of putting it in the decorator. Also the concatenation if IDs differs from the `markings` field so we should be consistent with this and the other PR. I'm not sure if a comma separator is better or something else.

app/decorators/products_decorator.rb
```diff
@@ -3,8 +3,14 @@ class ProductsDecorator < Draper::CollectionDecorator
 
   def to_csv
     CSV.generate do |csv|
-      csv << Product.attribute_names
+      csv << Product.attribute_names.dup.append(additional_csv_fields).flatten
```
slorek, line 6:
```suggestion
      csv << Product.attribute_names.dup.concat(additional_csv_fields)
```

app/decorators/products_decorator.rb
```diff
@@ -3,8 +3,14 @@ class ProductsDecorator < Draper::CollectionDecorator
...
       each { |product| csv << product.to_csv }
     end
   end
+
+private
+
+  def additional_csv_fields
+    %w[case_ids]
+  end
```
slorek, line 15: I think this should go in the `ProductDecorator` as this defines the fields a single product should return.

slorek: Agreed to remove `to_csv` in the individual decorators since they are not useful.

app/decorators/product_decorator.rb
```diff
@@ -79,7 +79,13 @@ def markings
     object.markings.join(", ")
   end
 
-  def to_csv
-    attributes.keys.map { |key| send(key) }
+  def case_ids
+    object.investigations.map(&:pretty_id).to_s
```
slorek, line 83:
```suggestion
    object.investigations.map(&:pretty_id)
```
CSV should do this automatically.

app/decorators/product_decorator.rb
```diff
@@ -79,7 +79,13 @@ def markings
...
+  end
+
+private
+
+  def additional_csv_fields
+    %w[case_ids]
   end
```
slorek, line 90:
```suggestion
```
Looks to be redundant.

app/decorators/products_decorator.rb
```diff
@@ -3,8 +3,18 @@ class ProductsDecorator < Draper::CollectionDecorator
...
   end
+
+private
+
+  def product_to_csv(product)
+    attributes_for_export.map { |key| product.send(key) }
+  end
+
+  def attributes_for_export
+    Product.attribute_names.dup.concat(%w[case_ids])
```
slorek, line 18: `attribute_names` is sorted alphabetically so we can ensure this isn't shown out of order.

```suggestion
    Product.attribute_names.dup.concat(%w[case_ids]).sort
```

slorek (changes requested):
We should also ensure the `investigations` is included in the database query to avoid n+1 issues. See https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/1194/files#diff-1fffcf1a4effb60cbfb062145b6ececaffc723517e646b154034d46a0f447e46R24 for an example.

slorek (approved):

PR #1211, opened 2021-02-26
fix lost title for legacy audit that never had an action

StupidCodeFactory:
## Description

Call `super` for legacy audit which never had an action set

## Notes for testing 
@nasirkhanbeis I've artificially created a legacy audit https://psd-pr-1211.london.cloudapps.digital/cases/2103-0001/activity


StupidCodeFactory:
[Corrective action activity log entry has lost it's H3](https://trello.com/c/wDuVDDqA/939-corrective-action-activity-log-entry-has-lost-its-h3)

slorek (approved):

PR #1223, opened 2021-03-04
filter cases by team with access

StupidCodeFactory:
## Description

Filter case by team with access

1. I've had to clean up a bit the the helper to understand better default value being overridden in the params post request.     That meant changing "unchecked_value" for values understood by `ActiveModel::Type:Boolean`. Which made it easier to provide default value in the `SearchParams`  for object. I could have moved it under `app/forms` but would have made harder to evaluate the changes.
2. I've created the teams_with_access nested mapping type. I'm only using the `id` but also stored the `name` attributes it made it easier to debug and provide also an example  on how to filter nested data mapping. We could greatly simplify the index on "owner_id" and "creator_id" and "teams_with_access" by storing the `collaborations#collaborator_id`  and `collaborations#collaborator_type` in a nested type.
3. Investigations are re-index when the collaborators are added to and removed from a case. thanks to service objects this has been fairly trivial change.


spec/support/case_list_matcher.rb
```diff
@@ -4,10 +4,10 @@
   end
 end
 
-RSpec::Matchers.define :list_cases_in_order do |expected_cases|
+RSpec::Matchers.define :list_cases_in_order do |expected_cases_ids|
   match do |element|
     pattern = /(Allegation|Project|Enquiry): (\d{4}-\d{4})/
     page_case_ids = element.all(".psd-case-card span", text: pattern).map(&:text).map { |t| t.match(pattern).captures.last }
-    page_case_ids == expected_cases.map(&:pretty_id)
+    page_case_ids == expected_cases_ids
```
StupidCodeFactory, line 11: this made it easier to read the rspec output diff.

StupidCodeFactory:
[(8) Add filters for 'teams added to case' to case list.](https://trello.com/c/FcX8U2Gh/626-8-add-filters-for-teams-added-to-case-to-case-list)

spec/features/filter_investigations_spec.rb
```diff
@@ -150,16 +213,17 @@
   end
 
   scenario "combining filters" do
-    pending
```
StupidCodeFactory, line 153: finally fixed that pending spec!

nasirkhanbeis:
Hey @StupidCodeFactory seems "Team added to case" filter broke other case owner filter. 
Step to replicate
1. Filter cases by " Team added " select Other team - enter "Southampton council team" returns only 1 case
2. Now un check selected filter and apply case owner filter "My team" , it suppose to return few cases but returns only 1 case

nasirkhanbeis:
Also pagination seems broken on cases list page.

StupidCodeFactory:
[(8) Add filters for 'teams added to case' to case list.](https://trello.com/c/FcX8U2Gh/626-8-add-filters-for-teams-added-to-case-to-case-list)

nasirkhanbeis:
Re-tested this on review app 1223, looks like Teams added to case filter works fine on its own however when applied along with other filter such as case owner or created by "My team" , it suppose to return cases where My team is added and also created by "My team" but it is returning only Teams added to case.

nasirkhanbeis:
Yann and me discussed and we agreed to raise a separate ticket for above issue. Tested all other scenarios looking good as expected. I will raise a separate ticket to fix filtering issue.

nasirkhanbeis (approved):

PR #1224, opened 2021-03-04
Add 'Notifying country' to case overview tab

KyleMacPherson:
https://trello.com/c/0ianlRdz/898-5-add-notifying-country-to-case-overview-tab

## Description
- Migration to add `country` field to teams table
- Migration to add `notifying_country` field to investigations table
- Rake task to backfill all teams new `country` field
- Logic to add populate `notifying_country` when a new investigation is created based on the country the current users team is based in.
- Tests.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


lib/tasks/backfill_team_countries.rb
```diff
@@ -0,0 +1,239 @@
+namespace :team do
+  desc "Backfill country field"
+  task backfill_country: :environment do
+    Team.all.each do |team|
+      team.update(country: team_to_country[team.name])
```
slorek, line 5: Suggestion: throw an exception if the value is `nil`.

Perhaps there should be a presence validation on the `Team` model.

spec/features/create_allegation_as_opss_user_spec.rb
```diff
@@ -34,7 +34,8 @@
     }
   end
 
-  let(:user) { create(:user, :activated, :opss_user) }
+  let(:team) { create(:team, country: "Iceland") }
```
slorek, line 37: Suggestion: Set a default `country` on the team factory.

And also a default `notifying_country` on the investigation factories.

So that the generated data better reflects real world usage.

app/controllers/investigations/project_controller.rb
```diff
@@ -64,6 +64,11 @@ def investigation_request_params
 
   def set_investigation
     @investigation = Investigation::Project.new(investigation_params).build_owner_collaborations_from(current_user)
+    set_notifying_country
+  end
+
+  def set_notifying_country
+    @investigation.notifying_country = current_user.team.country
```
slorek, line 71: Suggestion: presence validation on the `Investigation` model for `notifying_country`.

lib/tasks/backfill_team_countries.rb
slorek, line 5: Please change: We also need to backfill `notifying_country` on all cases where the team or one of its users was the case creator.

slorek (changes requested):
Please change: add the `notifying_country` to the case spreadsheet export template.

app/controllers/investigations/project_controller.rb
KyleMacPherson: Good call but would prefer to do this in a later PR if you don't mind. I think if we do it now then until the migration is successfully run users won't be able to create cases, so if the migration doesn't work properly first time we could have an issue!

nasirkhanbeis (approved):
looking good as expected.

lib/tasks/backfill_team_countries.rb
```diff
@@ -0,0 +1,244 @@
...
```
slorek, line 5: We probably want to use `update!` here to ensure exceptions are raised if validation fails.

lib/tasks/backfill_team_countries.rb
```diff
@@ -0,0 +1,244 @@
...
+    end
+
+    Investigation.includes(:creator_team).all.each do |investigation|
+      investigation.update(notifying_country: investigation.creator_team.country)
```
slorek, line 9: We probably want to use `update!` here to ensure exceptions are raised if validation fails.

lib/tasks/backfill_team_countries.rb
```diff
@@ -0,0 +1,244 @@
...
+        "West Lothian Council" =>	"Scotland",
+        "West Sussex County Council" =>	"England",
+        "West Yorkshire Joint Services" =>	"England",
+        "Wigan Metropolitan Borough Council" =>	"England",
+        "Wiltshire County Council" =>	"England",
+        "Wirral Borough Council" =>	"England",
+        "Wolverhampton City Council" =>	"England",
+        "Worcestershire County Council" =>	"England",
+        "Wrexham County Borough Council" =>	"Wales"
+      }
```
slorek, line 241: These values may not match with those in the `govuk-country-and-territory-autocomplete` node module. For example, United Kingdom is expressed as `country:GB` and other countries are expressed in the same format using their ISO country code (see https://github.com/alphagov/govuk-country-and-territory-autocomplete/blob/master/dist/location-autocomplete-canonical-list.json). We might consider using codes as described in https://en.wikipedia.org/wiki/ISO_3166-2:GB

lib/tasks/create_organisation.rake
```diff
@@ -1,7 +1,7 @@
 namespace :organisation do
   desc "Creates an organisation, team, and team admin user, and sends an invitation email to the user"
   task create: :environment do
-    result = CreateOrganisationWithTeamAndAdminUser.call(org_name: ENV.fetch("ORG_NAME"), admin_email: ENV.fetch("ADMIN_EMAIL"))
+    result = CreateOrganisationWithTeamAndAdminUser.call(org_name: ENV.fetch("ORG_NAME"), admin_email: ENV.fetch("ADMIN_EMAIL"), country: ENV.fetch("COUNTRY"))
```
slorek, line 4: Please update the docs in doc/tasks.md.

lib/tasks/backfill_team_countries.rb
KyleMacPherson: Ah great call, thanks @slorek 👌

app/controllers/investigations/allegation_controller.rb
```diff
@@ -16,6 +16,7 @@ def model_params
 
   def set_investigation
     @investigation = Investigation::Allegation.new(investigation_params).build_owner_collaborations_from(current_user)
+    set_notifying_country
```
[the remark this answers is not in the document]
KyleMacPherson: Nice! Was looking for where I should do this but couldn't seem to find the right spot, this is perfect, thanks @slorek

app/models/country.rb
```diff
@@ -2,9 +2,20 @@ class Country
...
+    ["Northern Ireland", "country:GB-NIR"]
+  ].freeze
+
   class << self
     def all
       @all ||= JSON.parse(File.read(PATH_TO_COUNTRIES_LIST))
     end
+
+    def notifying_countries
+      all + ADDITIONAL_COUNTRIES
```
slorek, line 18: Do we want to sort this list after adding the additional countries?

spec/factories/investigations.rb
```diff
@@ -10,6 +10,7 @@
     hazard_type           {}
     hazard_description    {}
     non_compliant_reason  {}
+    notifying_country     { "country:GB" }
```
slorek, line 13: We should be able to remove this as `CreateCase` is called in the `before(:create)` callback.

slorek (approved):
Just a couple of suggestions. Thanks for the good work!

nasirkhanbeis:
Re-tested on review app, looking good as expected. Give me shout when this is merged, I will do sanity check on prod.

PR #1236, opened 2021-03-11
Add opens in new tab hint text to external links

slorek:
https://trello.com/c/yb1dpVXU/894-add-opens-in-new-tab-to-links-with-this-behaviour

## Description
As per https://design-system.service.gov.uk/styles/typography/#links I have amended existing similar text to always use the correct language. I have not hidden the hint text as has been done elsewhere as according to the docs it is not appropriate in the other contexts.

There is a case to be made for removing the new tab functionality from many of the links according to the design system guidance, but I will leave this as a point for discussion.

Also fixes a bug with the online recall information link for corrective actions on the activity timeline not being rendered correctly.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


app/views/investigations/activities/corrective_action/_add.html.erb
```diff
@@ -4,7 +4,7 @@
   Legislation: <strong><%= activity.legislation %></strong>
   <br />
   <% if activity.online_recall_information.match?("http") %>
-    Recall information: <%= link_to activity.online_recall_information, rel: "noopener", target: "_blank" %>
```
slorek, line 7: Decorator already provides the HTML anchor so this caused the view to render incorrectly.

StupidCodeFactory (approved):

nasirkhanbeis (approved):
looks good as expected!
I think we need text "opens in new tab" in few more places too which we can discuss as team

PR #1249, opened 2021-03-15
Edit notifying country

KyleMacPherson:
https://trello.com/c/ilAkOSEO/942-5-ability-to-change-notifying-country

## Description
Allow users with `notifying_country_editor` role to edit the notifying country for an investigaiton

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis:
tested this on review app, looking good as expected, just the notification to be send to teams added rather to case owner

app/controllers/investigations/notifying_country_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+        redirect_to investigation_path(@investigation), flash: { success: "Notifying country changed." }
+      else
+        render :edit
+      end
+    end
+
+  private
+
+    def set_investigation
+      @investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id]).decorate
```
slorek, line 30: We have run into problems passing decorated objects around before where the decorator changes behaviour. This instance is passed to `NotifyingCountryForm` and `UpdateNotifyingCountry`. Suggest removing the `decorate` call here and delaying it until just before views are rendered.

app/controllers/investigations/notifying_country_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  class NotifyingCountryController < ApplicationController
+    before_action :set_investigation
+    before_action :authorize_user
+
+    def edit
+      @notifying_country_form = NotifyingCountryForm.from(@investigation)
+    end
+
+    def update
+      @notifying_country_form = NotifyingCountryForm.new(country: params["investigation"]["country"])
```
slorek, line 11: Suggest using Rails strong parameters instead of referring to the `params` directly.

app/controllers/investigations/ts_investigations_controller.rb
```diff
@@ -137,6 +137,7 @@ def product
 
   def set_investigation
     @investigation = Investigation::Allegation.new(investigation_step_params).build_owner_collaborations_from(current_user)
+    @investigation.notifying_country = current_user.team.country
```
slorek, line 140: I don't think this is needed since it is done in `CreateCase`.

app/decorators/audit_activity/investigation/update_notifying_country_decorator.rb
```diff
@@ -0,0 +1,11 @@
+class AuditActivity::Investigation::UpdateNotifyingCountryDecorator < ApplicationDecorator
+  delegate_all
+
+  def new_country
+    metadata["updates"]["notifying_country"].last
+  end
+
+  def previous_country
+    metadata["updates"]["notifying_country"].first
+  end
```
slorek, line 10: Not sure we need this decorator since these methods are present in the model, which I think is the appropriate place as these methods are not transforming the data.

app/models/audit_activity/investigation/update_notifying_country.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def previous_country
+    metadata["updates"]["notifying_country"].first
+  end
+
+  def new_country
+    metadata["updates"]["notifying_country"].second
+  end
+
+  # Do not send investigation_updated mail. This is handled by the ChangeRiskValidation service
```
slorek, line 32:
```suggestion
  # Do not send investigation_updated mail. This is handled by the UpdateNotifyingCountry service
```

app/models/user.rb
```diff
@@ -67,6 +67,10 @@ def is_psd_admin?
     has_role? :psd_admin
   end
 
+  def is_notifying_country_editor?
```
slorek, line 70: I think we want to move away from using the `is_` prefix. One benefit is when using Rspec matchers such as `be_X`. Prefer `notifying_country_editor?` or `can_edit_notifying_country?` as used on `can_validate_risk_level?`.

app/services/update_notifying_country.rb
```diff
@@ -0,0 +1,62 @@
+class UpdateNotifyingCountry
```
slorek, line 1: For consistency with other services which change an individual case property, suggest `ChangeCaseNotifyingCountry`.

app/services/update_notifying_country.rb
```diff
@@ -0,0 +1,62 @@
...
+  include Interactor
+  include EntitiesToNotify
+
+  delegate :investigation, :country, :user, to: :context
+
+  def call
+    context.fail!(error: "No investigation supplied") unless investigation.is_a?(Investigation)
+    context.fail!(error: "No user supplied") unless user.is_a?(User)
+
+    context.changes_made = false
```
slorek, line 11: Suggest removing this as it's not used anywhere.

app/forms/notifying_country_form.rb
```diff
@@ -0,0 +1,13 @@
+class NotifyingCountryForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveModel::Serialization
+
+  attribute :country, :string
+
+  validates :country, presence: true
```
slorek, line 8: Suggestion: we could also validate that the supplied value is in the `Country.notifying_countries` list.

slorek (changes requested):
Good work. Just some minor changes.

spec/features/change_notifying_country_spec.rb
```diff
@@ -0,0 +1,45 @@
+require "rails_helper"
+
+RSpec.feature "Adding and removing business to a case", :with_stubbed_mailer, :with_stubbed_elasticsearch do
```
slorek, line 3:
```suggestion
RSpec.feature "Changing the notifying country of a case", :with_stubbed_mailer, :with_stubbed_elasticsearch do
```

spec/features/change_notifying_country_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  context "when user is a notifying_country_editor" do
+    before do
+      user.roles.create!(name: "notifying_country_editor")
+    end
+
+    it "can succesfully change pre-populated notifying_country" do
+      investigation.update!(notifying_country: "country:GB-ENG")
+
+      sign_in_and_visit_change_notifying_country_page("England")
```
slorek, line 15: Suggestion: it would be beneficial to test that the notifying country is shown as `England` on the case overview page and clicking the `Change` link there, and also checking that `England` is the pre-populated value on the form.

spec/features/change_notifying_country_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      select "Scotland", from: "Notifying country"
+      click_button "Change"
+      expect(page.find("dt", text: "Notifying country")).to have_sibling("dd", text: "Scotland")
```
slorek, line 19: Suggestion: Check we are on the case overview page before making content assertions, as this would help debug a failing test for example if the form showed a validation error.

spec/models/audit_activity/investigation/change_notifying_country_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      it "produces a Hash of the change" do
+        expect(metadata).to eq({
+          updates: {
+            "notifying_country" => [previous_country, new_country]
+          }
+        })
+      end
+    end
+
+    context "when the case'scountry has not been updated" do
```
slorek, line 28:
```suggestion
    context "when the case's country has not been updated" do
```

app/models/audit_activity/investigation/change_notifying_country.rb
```diff
@@ -0,0 +1,34 @@
...
+      updates: updated_values
+    }
+  end
+
+  def title(*)
+    "Notifying country changed"
+  end
+
+  def body
+    "Notifying country changed from #{previous_country} to #{new_country}"
```
slorek, line 19: Suggestion: remove the `#body` method and move this text to the view template.

spec/models/audit_activity/investigation/change_notifying_country_spec.rb
slorek, line 28: This isn't a use case we should ever encounter so I'm not sure we need to test for this.

app/models/audit_activity/investigation/change_notifying_country.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def previous_country
```
slorek, line 22: Add a test for this method.

app/models/audit_activity/investigation/change_notifying_country.rb
```diff
@@ -0,0 +1,34 @@
...
+    metadata["updates"]["notifying_country"].first
+  end
+
+  def new_country
```
slorek, line 26: Add a test for this method.

slorek (approved):
Just some suggestions to make but otherwise looks good. Thanks!

spec/features/change_notifying_country_spec.rb
KyleMacPherson: If I understand what you are saying correctly I believe that is tested in the `sign_in_and_visit_change_notifying_country_page`.

PR #1258, opened 2021-03-22
Fix corrective action form validation when changing action from 'Other'

slorek:
https://trello.com/c/CBms5RQr/900-edit-corrective-action-when-action-is-updated-from-other-type-to-a-different-it-returns-validation-error

## Description
Fixes the form validation when changing the action from "Other' to another value. The form will now clear the 'other' description before validation, preventing an error being thrown which is not particularly helpful to the user. On the downside, the description will no longer be retained if the user made a mistake.


KyleMacPherson (approved):

app/forms/corrective_action_form.rb
```diff
@@ -54,6 +54,7 @@ class CorrectiveActionForm
   before_validation { trim_whitespace(:other_action, :details, :file_description, :online_recall_information) }
   before_validation { nilify_blanks(:other_action, :details, :file_description, :online_recall_information) }
   before_validation :clear_related_file, unless: :related_file
+  before_validation :clear_other_action, unless: -> { other? }
```
KyleMacPherson, line 57: This is cool!

nasirkhanbeis (approved):
looks good as expected

PR #1261, opened 2021-03-22
Fix corrective action form not retaining file attachment on validation errors

slorek:
https://trello.com/c/bUOP9Y2Q/883-regression-bug-on-corrective-action-page-uploaded-file-gets-cleared-on-other-field-errors

## Description
Fixes the add corrective action form not retaining file uploads when there are validation errors.


KyleMacPherson (approved):

nasirkhanbeis (approved):
looking good as expected!

PR #1279, opened 2021-03-28
Fix test result bug

KyleMacPherson:
https://trello.com/c/64VjkgMm/881-regression-test-result-in-ts-flow-is-not-being-recored

## Description
Raise an error when further_test_results field is not populated by a ts user.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


StupidCodeFactory (approved):
Could have added an extra test in `spec/features/report_product_spec.rb` but this is ok I think.

PR #1288, opened 2021-03-30
Fix test that fails due to difference in ActiveRecord and Rails Timez…

KyleMacPherson:
…ones

Our application has the timezone set to "Europe/London" within application.rb, however,
as noted in the same file, ActiveRecord is set to stay in UTC. This seems like it is a decision
that has been made deliberately so I have used a date outside of British Summer Time in order
to respect this and not changed the timezone in the test in order to not artificially manipulate
the interaction between Rails and AR in this case.

https://trello.com/c/iRbYnTzh/989-fix-failing-test-due-to-timezone-issue

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/spec/requests/export_investigations_spec.rb
+++ b/spec/requests/export_investigations_spec.rb
@@ -240,7 +240,7 @@ RSpec.describe "Export investigations as XLSX file", :with_elasticsearch, :with_
 
       context "when investigation is closed" do
         it "date_closed column is empty" do
-          closed_at_date = Date.new(2021, 01, 01)
+          closed_at_date = Date.new(2021, 0o1, 0o1)
           create(:allegation, is_closed: true, date_closed: closed_at_date)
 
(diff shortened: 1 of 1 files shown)


spec/requests/export_investigations_spec.rb
```diff
@@ -240,15 +240,16 @@
 
       context "when investigation is closed" do
         it "date_closed column is empty" do
-          create(:allegation, is_closed: true, date_closed: Date.yesterday)
+          closed_at_date = Date.new(2021, 0o1, 0o1)
```
StupidCodeFactory, line 243: @KyleMacPherson I'm confused by this `0o1` notation?

StupidCodeFactory (approved):

PR #1291, opened 2021-04-01
Spike paralellised test

StupidCodeFactory:
## description

1. split the test by folder and further split them by whether they run Elasticsearch test or not
2. run with parallel test, each job runner has two cores so this will further parallelise the test run


slorek (approved):

slorek:
One thing we will need to bear in mind is that if a new directory is introduced under `spec/` it will need to be added to the test run matrix.

StupidCodeFactory:
> One thing we will need to bear in mind is that if a new directory is introduced under `spec/` it will need to be added to the test run matrix. I wonder if one of them can include a wildcard to catch these?

@slorek  yes this is definitely a potential issue, I'll try to see if I can add and excluding regexp to catch those not in the listed directory

PR #1292, opened 2021-04-06
Change close/re-open case behaviour

slorek:
https://trello.com/c/YhFmvGBb/660-update-case-status-behaviour-in-non-js-case-bar-action-flow

## Description
Changes the behaviour of the close case call to action. Now the user will be directed to a page tailored to the intended action (close or re-open the case) rather than a generic case status page.

I have taken the opportunity to refactor the legacy code according to our established conventions, being a service object to handle the business logic, removal of callbacks and transient attributes on the `Investigation` model to generate activity and emails, and storing metadata on the activity.

This has slightly changed the behaviour of who receives the notification email and the text of the email.

I have slightly changed some of the copy from the design to use "re-open" consistently rather than a mix of open/reopen/re-open.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


app/views/investigations/close.html.erb
```diff
@@ -0,0 +1,26 @@
...
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds-from-desktop">
+      <%= render "form_components/govuk_error_summary", form: form %>
+
+      <%= render "form_components/govuk_textarea",
+                 key: :rationale,
+                 form: form,
+                 attributes: { maxlength: 10_000 },
+                 label: { text: "Why are you closing the case?", classes: "govuk-label--l", isPageHeading: true },
+                 hint: { text: "For example, the product is low risk, so no further activity is expected." }
```
KyleMacPherson, line 21: Suggestion: Could put these in en.yml

app/forms/change_case_status_form.rb
```diff
@@ -0,0 +1,10 @@
+class ChangeCaseStatusForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveModel::Serialization
+
+  attribute :new_status
+  attribute :rationale
+
+  validates_inclusion_of :new_status, in: %w[open closed]
```
KyleMacPherson, line 9: Trello card seems to indicate that we want to validate presence of rationale (for both closing or re-opening).

KyleMacPherson (approved):

app/forms/change_case_status_form.rb
slorek: Need to check this with Toby as it is a change in behaviour I'm not sure is expected.

nasirkhanbeis:
Tested this on review app, functionally working as expected. JS and non-js working fine as expected. Also I see confirmation banner message is clearer than what we have on prod.

observed few things to address
1. Cancel link isn't shown on re-open/close form
2. Notifications - don't think it is sending notification to case creator when a case is closed or re-opened.

slorek:
Confirmed with @tobyhughes that the reason field should not be mandatory, and the word "successfully" should not be used in notification copy since it is redundant when used with the new-look notification component which will be implemented shortly.

nasirkhanbeis:
Re-tested this on review app 1292, cancel link looks good. Agreed to keep the reason field optional. 
re: notification, don't know if we still need but I think when a case is closed or re-open we send notification to case creator but don't think it is sending. @tobyhughes FYI

config/routes.rb
```diff
@@ -96,8 +96,10 @@
             param: :pretty_id,
             concerns: %i[document_attachable] do
     member do
-      get :status
-      patch :status
+      get :close
+      get :reopen
+      patch :close
+      patch :reopen
```
StupidCodeFactory, line 102: Suggestion for future controller/routes: I think it would have been nicer to have an `Investigation::StatusController` other than further clutter the existing investigation controller which already does too much,

spec/features/change_case_status_spec.rb
```diff
@@ -6,34 +6,37 @@
...
+
+    fill_in "Why are you closing the case?", with: "Case has been resolved."
 
-    click_button "Save"
+    click_button "Close case"
 
     expect_to_be_on_case_page(case_id: investigation.pretty_id)
 
-    expect(page).to have_text("Allegation was successfully updated")
+    expect(page).to have_text("Allegation was closed")
```
StupidCodeFactory, line 39: Suggestion: use `have_css` for better test readability, robustness and consistency with the rest of the test suite.

```ruby
expect(page).to have_css ".hmcts-banner__message", text: "Allegation was closed"
```

StupidCodeFactory (approved):

app/controllers/investigations/status_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  private
+
+    def change_case_status(new_status:, template:, flash:)
+      investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
+      authorize investigation, :change_owner_or_status?
+
+      @change_case_status_form = ChangeCaseStatusForm.new(change_case_status_form_params.merge(new_status: new_status))
+
+      if !@change_case_status_form.valid? || !request.patch?
```
StupidCodeFactory, line 19: Suggestion: I'm not sure why we need to check the HTTP verb? is this method used for several HTTP verbs? if so would it not be preferable to have further split the controller? `Investigations::Status::ReopenController` and `Investigations::Status::CloseController`? it gets tricky to debug and test when the same action is used for multiple HTTP verbs.

I think it would at least be useful to have a comment state why we need to check the HTTP verb.

StupidCodeFactory: not a blocker and I should have been clearer in my previous review! 🤷‍♂️

StupidCodeFactory (approved):

app/controllers/investigations/status_controller.rb
slorek: If it’s a GET request then we should just show the form without performing validation. My initial preference was to just duplicate the code to avoid the ambiguity and extra complexity at the expense of duplicate code, but on balance I felt this was slightly better. I did consider splitting it further as you suggested, but that would create more code and/or need a new helper, and while status makes sense as a sub resource, status/close does not. I’ll add a comment.

PR #1298, opened 2021-04-08
new remove business page

StupidCodeFactory:
## Description
new remove  business behaviour. It's more or less a copy of Kyle's work on remove a product.

## NOTES
I had to run `rails  app:update:bin` as I fiddled on another branch trying to update `webpack`. Not sure exactly why but this caused my dev environment to stop working. Ultimately this bring this binaries more inline with the current rails version.

Screenshot 2021-04-15 at 15 58 26
Screenshot 2021-04-15 at 15 58 36
Screenshot 2021-04-15 at 15 58 41
Screenshot 2021-04-15 at 16 08 10
Screenshot 2021-04-15 at 16 08 44
Screenshot 2021-04-15 at 16 09 03

## Checklist:


StupidCodeFactory:
[(8) Update remove business behaviour (new version)](https://trello.com/c/EHX478iJ/977-8-update-remove-business-behaviour-new-version)

nasirkhanbeis:
I think Remove business should be h1?

nasirkhanbeis:
@StupidCodeFactory  also h1 should read as Remove {business name}

nasirkhanbeis (approved):
Re-tested all the fixes, looking good as expected.

tobyhughes:
Couple of things from me:

- White spacing looks a little weird and we need to remove the  capitalisation on `Business Details`
- Error summary should be above the H1
- `Reason for removing the business from the case` should be a label rather than hint text

image


image


image

bin/rails
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env ruby
-APP_PATH = File.expand_path("../config/application", __dir__)
+load File.expand_path("spring", __dir__)
+APP_PATH = File.expand_path('../config/application', __dir__)
```
slorek, line 3: Please put these changes in a separate PR

bin/rake
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env ruby
+load File.expand_path("spring", __dir__)
```
slorek, line 2: Please put these changes in a separate PR

bin/spring
```diff
@@ -1,10 +1,14 @@
...
   require "bundler"
-  Bundler.locked_gems.specs.find { |spec| spec.name == "spring" }&.tap do |spring|
+
+  # Load Spring without loading other gems in the Gemfile, for speed.
+  Bundler.locked_gems&.specs&.find { |spec| spec.name == "spring" }&.tap do |spring|
     Gem.use_paths Gem.dir, Bundler.bundle_path.to_s, *Gem.path
     gem "spring", spring.version
     require "spring/binstub"
+  rescue Gem::LoadError
+    # Ignore when Spring is not installed.
```
slorek, line 12: Please put these changes in a separate PR

bin/yarn
```diff
@@ -1,16 +1,14 @@
...
+  yarn = ENV["PATH"].split(File::PATH_SEPARATOR).
+    select { |dir| File.expand_path(dir) != __dir__ }.
+    product(["yarn", "yarn.cmd", "yarn.ps1"]).
+    map { |dir, file| File.expand_path(file, dir) }.
+    find { |file| File.executable?(file) }
 
-  if executable_path
-    exec File.expand_path(Pathname.new(executable_path).join('yarn')), *ARGV
+  if yarn
+    exec yarn, *ARGV
```
slorek, line 11: Please put these changes in a separate PR

slorek (changes requested):

spec/services/remove_business_from_case_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      before { corrective_action }
+
+      it "is a failure with the relevant error", :aggregate_failures do
+        expect(result).to be_a_failure
+        expect(result.error).to eq(:business_is_attached_to_supporting_information)
+      end
+    end
+
+    context "when a risk assessemnt is attached to the business" do
```
slorek, line 57:
```suggestion
    context "when a risk assessment is attached to the business" do
```

app/decorators/audit_activity/business/destroy_decorator.rb
```diff
@@ -0,0 +1,9 @@
+module AuditActivity
+  module Business
+    class DestroyDecorator < ActivityDecorator
+      def title(_viewing_user)
```
slorek, line 4: We'll need a migration to generate metadata for existing activities if we override the `title` method.

slorek (changes requested):

app/controllers/investigations/businesses_controller.rb
```diff
@@ -23,19 +14,40 @@ def index
...
+    @investigation = investigation.decorate
   end
 
   def create
-    authorize @investigation, :update?
-    create!
+    investigation = Investigation.find_by(pretty_id: params[:investigation_pretty_id])
+    authorize investigation, :update?
+
+    @business = investigation.businesses.new(business_params)
```
StupidCodeFactory, line 28: I'm aware this is slightly different than the rest of the code. Here a form object would only be used to to run validation so  I figured maybe it's ok to do it like this.
I'm open to suggestions

nasirkhanbeis (approved):
Re-tested remove business behaviour, looking good as expected. 
Also tested add business in TS and add business to case, looks good.

app/models/investigation.rb
```diff
@@ -40,10 +40,7 @@ class Investigation < ApplicationRecord
   has_many :products, through: :investigation_products
 
   has_many :investigation_businesses, dependent: :destroy
-  has_many :businesses,
-           through: :investigation_businesses,
-           after_add: :create_audit_activity_for_business,
-           after_remove: :create_audit_activity_for_removing_business
+  has_many :businesses, through: :investigation_businesses
```
slorek, line 43: 👏

lib/tasks/one_off.rake
```diff
@@ -0,0 +1,13 @@
+namespace :db do
+  namespace :migration do
+    desc "Migrates AuditActivity::Business::Destroy to new metadata format"
+    task business_removed_audit: :environment do
+      AuditActivity::Business::Destroy.where(metadata: nil).find_each(&:migrate_to_metadata)
```
slorek, line 5: I think it would be worth considering using the metadata migration convention I've introduced in https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/1354 - this will avoid potential errors and can be run more safely at any time after deployment.

app/models/audit_activity/business/destroy.rb
```diff
@@ -1,16 +1,25 @@
...
+  end
+
+  def self.from(_business, _investigation)
+    raise "Deprecated - use RemoveBusinessFromCase.call instead"
   end
 
-  def email_update_text(viewer = nil)
-    "Business was removed from the #{investigation.case_type} by #{source&.show(viewer)}."
+  def migrate_to_metadata
+    match = title.match(/Removed: (?<tradding_name>.*)/)
```
slorek, line 13: `tradding_name` is a typo I think?

app/models/audit_activity/business/destroy.rb
```diff
@@ -1,16 +1,25 @@
...
+    self.metadata = { business: { trading_name: match["tradding_name"] } }
```
slorek, line 14: Can we get the other business attributes so it matches new records?

slorek (changes requested):
Thanks for tackling this and the valuable refactoring. I've added some comments for suggested changes as I think there are a few things we should address before going ahead for consistency and safety.

app/controllers/investigations/business_investigations_controller.rb
```diff
@@ -0,0 +1,24 @@
+class Investigations::BusinessInvestigationsController < ApplicationController
```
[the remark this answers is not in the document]
StupidCodeFactory: The first page of the flow uniquely deals with building the `InvestigationBusiness` - just realising now that i've inverted the join model name 🤦  , so yes that could have been indeed confusing. However it seemed adequate to have an `InvestigationBusinessController` for dealing with the building of `InvestigationBusiness` model, seemed pretty much rails text book to me like a scaffold. At least that was the idea behind. It's also generally good practice to have single purpose classes.

So I'm not going to restore `Wicked::Wizard` I find it far more confusing to understand the flow of the requests, hard to debug and generally unhelpful when one can just use the regular HTTP forms on a given URL.
Generally, I don't understand what problem this library is trying to solve? 

I'm not a massive fan of fat controller but I'm going to add those two actions to the `Investigation::BusinessController`.

app/forms/add_business_to_case_form.rb
```diff
@@ -0,0 +1,37 @@
+# This class helps building the InvestigationBusiness model
+class AddBusinessToCaseForm
```
StupidCodeFactory, line 2: I'm not convinced `InvestigationBusinessForm` was not a better name.

slorek (approved):

PR #1300, opened 2021-04-08
Test result attachment errors

KyleMacPherson:
https://trello.com/c/6NkU8dEZ/857-test-result-audit-log-missing-some-info

## Description
Various fixes around the audit log for test results

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/models/audit_activity/test/test_result_updated.rb
+++ b/app/models/audit_activity/test/test_result_updated.rb
@@ -32,7 +32,9 @@ class AuditActivity::Test::TestResultUpdated < AuditActivity::Test::Base
   end
 
   def new_details
-    updates["details"]&.second
+    if updates["details"]&.second
+      updates["details"]&.second.blank? ? "Removed" : updates["details"]&.second
+    end
   end
 
   def new_legislation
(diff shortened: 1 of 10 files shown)


nasirkhanbeis (approved):

app/models/audit_activity/test/test_result_updated.rb
```diff
@@ -32,7 +32,9 @@ def new_result
   end
 
   def new_details
-    updates["details"]&.second
+    if updates["details"]&.second
+      updates["details"]&.second.presence || "Removed"
```
slorek, line 36: I think we will need a migration for existing activity records to nilify the values where they are empty for this to work.

slorek (changes requested):
Looks good, just need to think about migrating existing data I think.

app/models/audit_activity/test/test_result_updated.rb
KyleMacPherson: Ok after some research we don't actually need this after all:
``` 
"".presence => nil
"x".presence => "x"
```

slorek (approved):
Migration not needed after all.

PR #1308, opened 2021-04-14
Fix flaky spec

slorek:
This one keeps failing regularly because the ElasticSearch index has not finished re-indexing before the test has run. This forces the spec to wait for the response from ElasticSearch before continuing.


KyleMacPherson (approved):
Niiiiice

PR #1312, opened 2021-04-15
Replace deprecated sentry-raven gem

slorek:
The [sentry-raven gem](https://github.com/getsentry/sentry-ruby/tree/master/sentry-raven) has been deprecated and replaced with [sentry-ruby](https://github.com/getsentry/sentry-ruby/tree/master/sentry-ruby).

This brings our Sentry integration up to date and removes the deprecation warnings.


Gemfile.lock
```diff
@@ -657,7 +663,8 @@ DEPENDENCIES
   sass-rails (~> 6.0)
   scout_apm (~> 4.0)
   scss_lint-govuk (~> 0.2)
-  sentry-raven (~> 3.1)
+  sentry-rails
+  sentry-sidekiq
```
StupidCodeFactory, line 667: is there a configuration option where we can specify a different queue for sidekick, in case loads of errors comes in a back up the queues, preventing 2FA codes to be sent or reset email password and such?

StupidCodeFactory (approved):

Gemfile.lock
slorek: I’m not sure I understand. The Sentry gem should not alter the failure behaviour of jobs.

PR #1317, opened 2021-04-15
Contact details bug

KyleMacPherson:
https://trello.com/c/MmTX4MW0/639-dont-display-contact-details-row-in-case-overview-when-its-not-provided

## Description
If we do not have any contact details for the complainant on an allegation there is no point showing the contact details row, this change removes it in such a situation.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis (approved):
Suggestion: permission notice, it'd be nice if we can bring that down below the contact information. We can always improve the design how we want to show in separate ticket.

PR #1329, opened 2021-04-21
Update rails bins

StupidCodeFactory:
## Description
When rails was upgrade to 6.1 bins weren't updated which caused my lots of
headaches try to update my local development environment


slorek (approved):

StupidCodeFactory:
[Upgrade webpacker](https://trello.com/c/EqD0cb27/1004-upgrade-webpacker)

PR #1331, opened 2021-04-22
Fix error bugs on actions form

KyleMacPherson:
https://trello.com/c/jRw9wt3E/1019-non-js-actions-page-no-field-level-error-when-nothing-selected

## Description
Errors not correctly displaying on actions form bug. This PR fixes the issue.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/controllers/investigations/actions_controller.rb
+++ b/app/controllers/investigations/actions_controller.rb
@@ -43,8 +43,6 @@ module Investigations
     end
 
     def actions_params
-      return {} if params[:investigation_actions_form].blank?
-
       params.require(:investigation_actions_form).permit(:investigation_action)
     end
   end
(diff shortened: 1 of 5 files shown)


nasirkhanbeis (approved):
Nice work. looks good as expected!

StupidCodeFactory (dismissed):

app/controllers/investigations/actions_controller.rb
```diff
@@ -41,5 +41,11 @@ def path_for_action(action)
         investigation_risk_level_path(@investigation)
       end
     end
+
+    def actions_params
+      return {} if params[:investigation_actions_form].blank?
```
StupidCodeFactory, line 46: ✂️  this needs to be removed no? this method is only used in the `#create` action. Which should always have a form POSTed to it. If it does not then `params.require`will return the appropriate status. It is its main purpose.

PR #1333, opened 2021-04-22
Fix docker setup

StupidCodeFactory:
## Description

 Fix docker setup

re-added bootsnap gem for faster load time as it was quite slow. It comes anyway by default with rails. 

Commented the Kibana section in the docker-compose file as this is for debugging elasticsearch purpose.


.gitignore
```diff
@@ -64,3 +64,10 @@ config/master.key
 
 # Redis snapshots. See https://redis.io/topics/persistence
  dump.rdb
+
+/public/packs
+/public/packs-test
+/node_modules
+/yarn-error.log
+yarn-debug.log*
+.yarn-integrity
```
slorek, line 73: Please remove: These are already declared in the file earlier on.

slorek (changes requested):

slorek (approved):

PR #1334, opened 2021-04-22
Add product and business to case through 'add new' menu

KyleMacPherson:
https://trello.com/c/OJcbB7cs/658-expand-add-new-case-bar-functionality-to-include-businesses-and-products-js-and-non-js

## Description
Add product and business to case through 'add new' menu

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


slorek:
@KyleMacPherson do we still need `investigations/supporting_information_controller.rb`?

tobyhughes (approved):
Checked both `Product` and `Business` options in the case bar (with JS enabled and disabled) - all working as expected.

Also checked that only teams added to the case with edit permissions, and the case owner, are able to add products and businesses via the case bar.

KyleMacPherson:
> @KyleMacPherson do we still need `investigations/supporting_information_controller.rb`?

Yeah I think so @slorek because we still want the regular supporting information workflows, but we also additionally want this actions workflow which includes both supporting information and products and businesses.

slorek:
> > @KyleMacPherson do we still need `investigations/supporting_information_controller.rb`?
> 
> Yeah I think so @slorek because we still want the regular supporting information workflows, but we also additionally want this actions workflow which includes both supporting information and products and businesses.

It seems we are duplicating a lot of code which could present issues with maintaining consistency later on. I wonder if we can use the same code for both journeys, since the only difference is the addition of two extra options.

app/views/investigations/supporting_information/_form.html.erb
```diff
@@ -0,0 +1,27 @@
+<% information_types = @add_to_case_action ? SupportingInformationTypeForm::MAIN_TYPES.merge({ product: "Product", business: "Business" }) : SupportingInformationTypeForm::MAIN_TYPES %>
```
slorek, line 1: Suggestion: set this in the form object in the controller. This will reduce logic implemented in the view, making it more testable, and we can also use that list of valid options for validation.

slorek (approved):

PR #1337, opened 2021-04-26
Fix sentry error

StupidCodeFactory:
## Description

This should fix the following error https://sentry.io/organizations/beis/issues/2328211151/?project=1329381&referrer=slack

I was not able to reproduce the error, this could have happen during a deploying the new search form


slorek:
This error seems to have been introduced by https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/1223.

slorek (approved):

PR #1338, opened 2021-04-26
fix investigation change status legacy audit activities

StupidCodeFactory:
## Description
migrate legacy audit activities - fixes for https://trello.com/c/y4742107/1021-fix-sentry-https-sentryio-organizations-beis-issues-2358055839-project1329381referrerslack


app/models/audit_activity/investigation/update_status.rb
```diff
@@ -1,4 +1,9 @@
 class AuditActivity::Investigation::UpdateStatus < AuditActivity::Investigation::Base
+  def migrate_to_metadata
+    self.metadata = { updates: { is_closed: [false, true], date_closed: [nil, created_at] } }
```
slorek, line 3: Please change: Need to cover re-opened case activities as well.

slorek (changes requested):

slorek (approved):

PR #1340, opened 2021-04-27
Visibility flow

KyleMacPherson:
https://trello.com/c/Edg1OCS9/507-new-case-restriction-flow

## Description
Change the case restriction flow to be more explicit for the user.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


slorek (changes requested):
This is a good opportunity to refactor the code, to remove the `create_audit_activity_for_visibility` callback on the `Investigation` model and use service object to handle the creation of activity and notifications. This work is very similar to that done on #1292 for reference. In particular, for consistency we should consider having separate endpoints for making a case unrestricted or restricted, as we have done with close and re-open.

app/decorators/audit_activity/investigation/update_visibility_decorator.rb
```diff
@@ -0,0 +1,33 @@
...
+  def new_visibility
+    new_is_private? ? "restricted" : "unrestricted"
+  end
+
+  def rationale
+    metadata.dig("rationale")
+  end
+
+  def user_name
+    User.find(source.user_id).name
```
slorek, line 21:
```suggestion
    source.user.name
```

app/models/investigation.rb
```diff
@@ -34,8 +34,6 @@ class Investigation < ApplicationRecord
   validates :custom_risk_level, absence: true, if: -> { risk_level != "other" }
   validates :custom_risk_level, presence: true, if: -> { risk_level == "other" }
 
-  after_update :create_audit_activity_for_visibility
```
slorek, line 37: 🙌

app/services/change_case_visibility.rb
```diff
@@ -0,0 +1,83 @@
...
+        recipient.name,
+        recipient.email,
+        email_body(recipient),
+        email_subject
+      ).deliver_later
+    end
+  end
+
+  def email_recipients
+    (email_recipients_for_case_owner + email_recipients_for_case_creator).uniq
```
slorek, line 55: I'm not sure we need to notify the case creator for this change. I think the current behaviour for this activity type is the default as defined in `Activity` (implemented as `email_recipients_for_case_owner`).

app/forms/change_case_visibility_form.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def self.from(investigation)
+    old_visibility = investigation.is_private? ? "restricted" : "unrestricted"
+    new(case_type: investigation.case_type, old_visibility: old_visibility)
+  end
+
+private
+
+  def new_visibility_is_different
+    errors.add(:new_visibility, :same_as_existing, case_type: case_type, visibility: new_visibility) if new_visibility == old_visibility
```
slorek, line 22: I think this is a change in behaviour, let's check with @DH02 that we want to throw an error instead of accepting the form when the restriction hasn't changed.

slorek (approved):
Thanks for this, we are nearly there in removing all the callbacks and being able to refactor the `Activity` classes. Just a few minor comments to check before merging.

app/forms/change_case_visibility_form.rb
slorek: Ignore this, I didn't look at the designs!

KyleMacPherson: Cool, thanks @slorek, I think I will let Nas have a last look over this on Monday then merge then if he is happy!

nasirkhanbeis (approved):
re-tested this on review app, looks all good as expected.

PR #1349, opened 2021-04-29
use run task to run migration

StupidCodeFactory:
This has still a major flaw which is does not fail the build when the migration fails


StupidCodeFactory:
closing as this was just an example of how to run a long running tasks

PR #1354, opened 2021-05-04
Fix risk assessment file not displayed on added audit activity

slorek:
https://trello.com/c/2DTwhgRB/731-add-risk-assessment-file-attached-doesnt-get-record-in-activity-log

## Description
Fixes the attached file not being displayed on the activity log. Also tidies up the view template and moves logic which should not be in the view to a decorator, and backfills missing tests.

We now store the complete file metadata in the activity so can trace the history of a risk assessment more completely.

**Importantly, this introduces a proposed design for how to handle migrations of activity metadata without the use of ActiveRecord migrations.**

Migrations have previously caused issues during deployments due to deployment timeout thresholds being reached, and the zero-downtime deploys resulting in two versions of the code being active during the deployment - sometimes resulting in errors shown to users for a short time.

This design is summarised below:

* Override the `metadata` getter method to convert the Hash structure on the fly. This is useful during deployments when there are still instances with the old code active, and provides a neat way to implement the conversion logic without introducing extra methods
* Provide a generic rake task which can be used to migrate the metadata for a given class name.

Using the `metadata` override means the migration task is non-destructive if run repeatedly, and we don't need to keep creating single use rake tasks every time we want to change the structure.

Example usage:

```
CLASS_NAME=AuditActivity::RiskAssessment::RiskAssessmentAdded bin/rake activities:update_metadata
```

This task can be run at any time post-deployment, and in the meantime users should not suffer any errors.

### Limitations
On investigation of production data, I concluded it is not possible to reconcile the original risk assessment file if it has been subsequently updated. As of this writing, this affects 105 records out of a total of 560.

This is due to some of the files having been deleted (our codebase previously automatically deleted orphaned files), and the updated activity metadata only stores the filename which in most cases provides an ambiguous match. It may be possible to reconcile a subset of these 105 records via more intricate detective work, but that can be tackled as a separate task.

### DIFF
--- a/app/models/audit_activity/risk_assessment/risk_assessment_added.rb
+++ b/app/models/audit_activity/risk_assessment/risk_assessment_added.rb
@@ -86,7 +86,9 @@ private
     return metadata if metadata["risk_assessment"].present?
 
     risk_assessment = RiskAssessment.find(metadata["risk_assessment_id"])
-    risk_assessment_updated = AuditActivity::RiskAssessment::RiskAssessmentUpdated.where("metadata->'risk_assessment_id' = ?", risk_assessment.id.to_s).count.positive?
+
+    # ID must be cast to String to avoid SQL error querying JSON column
+    risk_assessment_updated = AuditActivity::RiskAssessment::RiskAssessmentUpdated.where("metadata->'risk_assessment_id' = ?", risk_assessment.id.to_s).any?
 
     if risk_assessment_updated
       # We cannot reliably find the old file since orphaned files were
--- a/app/models/audit_activity/risk_assessment/risk_assessment_added.rb
+++ b/app/models/audit_activity/risk_assessment/risk_assessment_added.rb
@@ -4,46 +4,121 @@ class AuditActivity::RiskAssessment::RiskAssessmentAdded < AuditActivity::Base
   end
 
   def self.build_metadata(risk_assessment)
-    {
-      risk_assessment_id: risk_assessment.id,
-      assessed_on: risk_assessment.assessed_on,
-      risk_level: risk_assessment.risk_level,
-      custom_risk_level: risk_assessment.custom_risk_level,
-      assessed_by_team_id: risk_assessment.assessed_by_team_id,
-      assessed_by_business_id: risk_assessment.assessed_by_business_id,
-      assessed_by_other: risk_assessment.assessed_by_other,
-      details: risk_assessment.details,
-      product_ids: risk_assessment.product_ids
-    }
+    { "risk_assessment" => risk_assessment.attributes.merge(
+      "product_ids" => risk_assessment.product_ids,
+      "risk_assessment_file" => risk_assessment.risk_assessment_file.blob&.attributes
+    ) }
   end
 
-  def title(*)
-    "Risk assessment"
+  # Temporarily override getter until data is migrated
+  # TODO: Remove method once data migrated
+  def metadata
+    migrate_metadata_structure
   end
 
   def subtitle_slug
     "Added"
   end
 
+  def risk_assessment
+    RiskAssessment.find(metadata["risk_assessment"]["id"])
+  end
+
+  # Some older activities cannot refer to the file as it was not recorded on the activity and may have subsequently been deleted, or the filename recorded on the updated activity was ambiguous
+  def risk_assessment_file
+    ActiveStorage::Blob.find(metadata["risk_assessment"]["risk_assessment_file"]["id"]) if metadata["risk_assessment"]["risk_assessment_file"]
+  end
+
   def products_assessed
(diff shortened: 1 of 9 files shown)


app/models/audit_activity/risk_assessment/risk_assessment_added.rb
```diff
@@ -4,46 +4,121 @@ def self.from(*)
...
+  #   product_ids: risk_assessment.product_ids
+  # }
+  def migrate_metadata_structure
+    metadata = self[:metadata]
+
+    # Already in new format
+    return metadata if metadata["risk_assessment"].present?
+
+    risk_assessment = RiskAssessment.find(metadata["risk_assessment_id"])
+    risk_assessment_updated = AuditActivity::RiskAssessment::RiskAssessmentUpdated.where("metadata->'risk_assessment_id' = ?", risk_assessment.id.to_s).count.positive?
```
StupidCodeFactory, line 89:
```suggestion
    risk_assessment_updated = AuditActivity::RiskAssessment::RiskAssessmentUpdated.where("metadata->'risk_assessment_id' = ?", risk_assessment.id.to_s).any?
```
I think this should work? not certain though as this queries the JSON field. But feel free to ignore.

spec/models/audit_activity/risk_assessment/risk_assessment_added_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+            "risk_level" => risk_assessment.risk_level,
+            "product_ids" => risk_assessment.product_ids
+          })
+        })
+      end
+      # rubocop:enable RSpec/ExampleLength
+
+      context "when the risk assessment has not been subsequently updated" do
+        it "returns the file blob metadata" do
+          expect(activity.metadata["risk_assessment"]["risk_assessment_file"]).to match(hash_including(risk_assessment.risk_assessment_file.blob.attributes.except("created_at", "updated_at")))
```
StupidCodeFactory, line 66: A trick I use to get around the string timestamps.
```suggestion
          expect(activity.metadata["risk_assessment"]["risk_assessment_file"]).to match(hash_including(JSON.parse(risk_assessment.risk_assessment_file.blob.attributes.to_json)))
```

lib/tasks/update_activity_metadata.rake
```diff
@@ -0,0 +1,25 @@
...
+    raise "Error: #{class_to_update} is not a subclass of Activity" unless class_to_update.new.is_a?(Activity)
+
+    succeeded = 0
+    failed = 0
+
+    puts "Migrating #{class_to_update.count} records of #{class_to_update}"
+
+    activities = class_to_update.all
+
+    activities.each do |activity|
```
StupidCodeFactory, line 15: Suggestion: Maybe a find_each instead of an `class_to_update.all` and `.each` would be a little less memory intensive?

app/models/audit_activity/risk_assessment/risk_assessment_added.rb
StupidCodeFactory, line 89: `risk_assessment.id.to_s` is to ensure this is queried as a string and not an integer if I understand correctly. Might be worth added a comment to document this intentional behaviour.

StupidCodeFactory (approved):

app/models/audit_activity/risk_assessment/risk_assessment_added.rb
slorek: Yes, it throws an SQL error otherwise.

lib/tasks/update_activity_metadata.rake
slorek: Yes you’re right.

nasirkhanbeis:
Hi @slorek can we make the attachment file name as hyperlink as we do for test results? to make is consistent? 
image

see the test result attachment 
image

app/models/audit_activity/risk_assessment/risk_assessment_added.rb
```diff
@@ -4,46 +4,121 @@ def self.from(*)
...
-      details: risk_assessment.details,
-      product_ids: risk_assessment.product_ids
-    }
+    { "risk_assessment" => risk_assessment.attributes.merge(
+      "product_ids" => risk_assessment.product_ids,
+      "risk_assessment_file" => risk_assessment.risk_assessment_file.blob&.attributes
+    ) }
   end
 
-  def title(*)
```
KyleMacPherson, line 20: Looks like we removed the title method, Is this intended?

slorek: It's moved to the decorator. Activity objects are already decorated by the time they are rendered in the view.

StupidCodeFactory (approved):

PR #1358, opened 2021-05-05
fix product not appearing in the search after removal

StupidCodeFactory:
## Description

Address bug where instead of updating the investigation index after removing the product, the product search index was removed.

### DIFF
--- a/app/services/remove_product_from_case.rb
+++ b/app/services/remove_product_from_case.rb
@@ -11,7 +11,7 @@ class RemoveProductFromCase
 
     investigation.products.delete(product)
 
-    product.__elasticsearch__.delete_document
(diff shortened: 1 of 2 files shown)


StupidCodeFactory:
[Remove product from a case not suppose to delete product from the db](https://trello.com/c/OmlS83Qs/1028-remove-product-from-a-case-not-suppose-to-delete-product-from-the-db)

KyleMacPherson (approved):

slorek (approved):

nasirkhanbeis (approved):
Tested this on review app, looking good as expected. will be nice to merge this independently and bring in changes in 'remove business behaviour' pr

PR #1380, opened 2021-05-12
Search improvement

StupidCodeFactory:
## Description

Improve multi term search by use the "and" operator on fuzzy matches


StupidCodeFactory:
[(3) Search by AND not OR](https://trello.com/c/7XpAgQVe/270-3-search-by-and-not-or)

PR #1381, opened 2021-05-12
Case owner page changes

KyleMacPherson:
https://trello.com/c/GRECFDAi/518-3-improve-case-owner-change-page

## Description
The main aim for this change is to split the radios up into sub groups (using dividers) so that it is easier for the user to select the person or group that they want to assign case ownership to, however there is also some extra logic to which users we should show in which groups:

You team section should show options (if available) in the following order

- case assignee
- current user
- someone in your team
- current team

- nothing should be repeated (eg. if case assignee is current user, the user should not be shown twice)

For OPSS users the `other` section should show:
 - OPSS Enforcement
 - OPSS Incident Management
 - OPSS Trading Standards Co-ordination
 - Other team [autocomplete of teams]
 - Other user[autocomplete of users]

For non OPSS users the `other` section should show:
- OPSS Incident Management (with hint text: "For reporting serious risks to OPSS")
 - Other team [autocomplete of teams]
 - Other user[autocomplete of users]

General rules:
- Do not show `Other teams added to the case` divider if there are no other teams added to the case
- Do not show the case_owner_team in Other teams added to the case
- Do not show the  case_owner_team in Other
- Do not show teams added to the case in other

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis (approved):
Tested this on review app, looks all good to me.

DH02:
Different HTML, including:
1.	Attribute missing from Form
2.	Different class on parent grid column
3.	The main title is different
4.	Missing title legend elements
5.	The fieldset is missing
6.	The main hint text HTML is different 
7.	The hint text is different and missing a full stop
8.	Form group div in wrong place
9.	Continue button has been wrapping in a form group
10.	Continue button is currently an input type=button (use a hidden input for value if required)
11.	The inline hint text for OPSS incident management doesn’t’ is missing the described by attributes and logic
12.	The inline hint text for OPSS incident management is using the wrong html element and is missing classes (wrong text size)
13.	The inline hint text for OPSS incident management is missing the full stop.

* Its become apparent that the list has hidden/expanding form select options – these non-gds styles can be picked up (and HTML improved) on the sitewide audit. (Until then we can keep the parent class govuk-grid-column-two-thirds-from-desktop to manage the ajax select forms)
* The sitewide habit of having a lone text hyperlink below the action button (inside paragraph tags?) can be reviewed/corrected during a sitewide audit.
* Toby: not sure about the sub-title “Enquiry: 2105-005” here, but that can be reviewed as part of a sitewide audit. Also, a more elegant accessibility compliant design solution can be found to avoid the JavaScript selects? I’m not even sure why we’re using them as the standard fall back (JavaScript disabled version) is much better?

slorek:
Bug spotted: OPSS Incident Management is shown twice in the list.

image

app/assets/stylesheets/_opss.scss
```diff
@@ -0,0 +1,18 @@
+.opss-radios__divider--title .govuk-radios__divider,
```
slorek, line 1: Unless we create a new bona fide component, I think we're supposed to use the `.app-` prefix as used elsewhere  to override design system components (some code is inconsistent, we also use the `.psd-` prefix but I think this is legacy code) - see https://design-system.service.gov.uk/get-started/extending-and-modifying-components/.

app/controllers/investigations/ownership_controller.rb
```diff
@@ -79,10 +79,18 @@ def form
...
-    @selectable_teams = [current_user.team, Team.get_visible_teams(current_user), @investigation.owner_team].flatten.uniq.compact
     @other_teams = Team.not_deleted
     @other_users = User.active.includes(:team)
+    @other_teams_added_to_case = Team.find(@investigation.collaboration_accesses.where("type IN (?)", ["Collaboration::Access::Edit", "Collaboration::Access::ReadOnly"]).map(&:collaborator_id))
+    @default_opss_teams = (default_opss_teams - [@investigation.owner, current_user.team] - @other_teams_added_to_case)
+  end
+
+  def default_opss_teams
+    if current_user.is_opss?
+      Team.where("name IN (?)", ["OPSS Enforcement", "OPSS Incident Management", "OPSS Trading Standards Co-ordination", "OPSS Operational support unit"])
```
slorek, line 91: I think we should use the existing `Team.get_visible_teams(current_user)` method (as used in the current code) which is configured in `config/important_team_names.yml` unless there is a reason not to? I think it provides this functionality. If we're not using we should remove this method and config file as I don't think it's used elsewhere, but I am uncomfortable with having names hardcoded in code.

app/controllers/investigations/ownership_controller.rb
```diff
@@ -79,10 +79,18 @@ def form
   end
 
   def get_potential_assignees
-    @selectable_users = [@investigation.owner_user, current_user].uniq.compact
     @team_members = current_user.team.users.active.includes(:team)
-    @selectable_teams = [current_user.team, Team.get_visible_teams(current_user), @investigation.owner_team].flatten.uniq.compact
     @other_teams = Team.not_deleted
     @other_users = User.active.includes(:team)
+    @other_teams_added_to_case = Team.find(@investigation.collaboration_accesses.where("type IN (?)", ["Collaboration::Access::Edit", "Collaboration::Access::ReadOnly"]).map(&:collaborator_id))
```
slorek, line 85: You could consider using the `changeable` scope here rather than hardcoding class names.
```suggestion
    @other_teams_added_to_case = Team.find(@investigation.collaboration_accesses.changeable).map(&:collaborator_id))
```
Or perhaps look at the `Investigation#teams_with_edit_access` and `Investigation#teams_with_read_only_access` associations and make another combined one.

slorek (changes requested):

app/assets/stylesheets/_opss.scss
DH02: Thanks - the provided link does recommend using a prefix (which opss- is). It uses "app-" as an example. This is taken from the page: 

"When writing code for your application, use a different prefix, **_like_** app-. If your department has its own design resources, they should use a new separate prefix. It’s a good idea to use departmental initials, like hmcts- or dvla-."

.opss- gives me a clean control for uninfected styles. Eventually we'll be able to remove legacy styles (app or psd).

app/controllers/investigations/ownership_controller.rb
DH02: @KyleMacPherson I've now merged into the prototypes master changes to the CSS as highlighted by @slorek . 

This has removed a new class that was overriding a govuk class. The HTML/sass has been updated on page: 
case-owner-change-page/improve-case-owner-change-page

<div ="govuk-radios__divider **opss-width-100**">
 
Please add class .opss-width-100 to the .govuk-radios__divider DIVs and get the latest sass from the opss-psd sass file.

app/controllers/investigations/ownership_controller.rb
```diff
@@ -79,10 +79,18 @@ def form
...
+    @other_teams_added_to_case = @investigation.non_owner_teams_with_access
+    @default_opss_teams = (default_opss_teams - [@investigation.owner, current_user.team] - @other_teams_added_to_case)
+  end
+
+  def default_opss_teams
+    if current_user.is_opss?
+      Team.get_visible_teams(current_user)
+    else
+      Team.where(name: "OPSS Incident Management")
+    end
```
slorek, line 94: This logic is already implemented in `Team.get_visible_teams`
```suggestion
    Team.get_visible_teams(current_user)
```

slorek (changes requested):
Bugs look to be resolved, thanks.

Just a small change needed I believe.

app/controllers/investigations/ownership_controller.rb
KyleMacPherson: @slorek done!

slorek (approved):

PR #1383, opened 2021-05-13
Improve search results

StupidCodeFactory:
## Description

This improves the search efficiency by applying a logic whereby if two words are searched for the search will perform and exact match but also search for exact match of the two terms individually in the across fields (fields are attributes in the Elasticsearch jargon comparing to ActiveRecord).


app/services/elasticsearch_query.rb
```diff
@@ -86,8 +86,9 @@ def match_exact_fields
       multi_match: {
         query: query,
         fields: @exact_fields,
-        type: "phrase",
-        fuzzy_transpositions: "false"
+        type: "cross_fields",
```
StupidCodeFactory, line 89: this is allows to "fallback" to cross field match

nasirkhanbeis:
Tested this on review app and spotted few issues
1. two and word and more search is returning accurate along with other single word match too, however when I search by single keyword it should return exact match on the top in search result - see attached screen shot.
image

2. Business search - when I search by empty keyword field it is returning on ABC business , expected to return all the businesses in search result page.

nasirkhanbeis:
Also I see, two word search isn't returning the accurate search results. 
e.g I have products name "Smile Spot Sorria Sempre Teeth whitening" and "Smile spot" when I search by keyword " Smile spot" I don't see exact match search result see screen below 
image

app/services/elasticsearch_query.rb
```diff
@@ -86,8 +86,9 @@ def match_exact_fields
       multi_match: {
```
KyleMacPherson, line 86: Do we need this to be a multi_match? Seems like we are only exact matching by one field (pretty_id)

app/services/elasticsearch_query.rb
```diff
@@ -97,7 +98,8 @@ def match_fuzzy_fields
...
         query: query,
         fields: @fuzzy_fields,
-        fuzziness: "AUTO"
+        fuzziness: "AUTO",
+        operator: "or"
```
KyleMacPherson, line 102: This operator should be "and" shouldn't it? As the trello card states the issue is: ```Currently, when a user searches for something if they add more than one word in the search box elastic search uses OR not AND. So as a use adds more terms the search becomes less precise. Contrary to user expectations.```

spec/services/elasticsearch_query_spec.rb
```diff
@@ -148,5 +148,31 @@ def perform_search
...
+        investigation_four.reload.__elasticsearch__.index_document(refresh: :wait_for)
+      end
+
+      it "finds the relevant investigations", :aggregate_failures do
+        expect(perform_search.records.to_a)
+          .to include(
+            investigation_one,
+            investigation_two,
+            investigation_three,
+            investigation_four
```
KyleMacPherson, line 173: If I am reading the trello card correctly `investigation_three` and `investigation_four` should be filtered out in this case

KyleMacPherson (changes requested):

StupidCodeFactory:
@nasirkhanbeis  this was PR index not being correctly indexed.

image

PR #1386, opened 2021-05-14
Highlight cases in nav tab for add new comments, images or other docs

KyleMacPherson:
https://trello.com/c/8H4oeahI/609-add-comment-flow-does-not-highlight-cases-in-the-nav

## Description
Cases was not highlighted in the nav bar when a user was adding a comment to the case. I also noticed the same was happening for add new image and add other documents. Have taken a look through the app and cannot see it happening anywhere else.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis (approved):
looks good as expected!

StupidCodeFactory (approved):

PR #1390, opened 2021-05-17
Improve search results

StupidCodeFactory:
## Description
This pull request aims at drastically improve the search results by using and "and" logic when entering a search query containing multiple keywords. 
We still want to retain to retain an "or" logic but those should have lower precedence. **However** this is not quite working at the moment, as some of the fuzzy "or" matches are on top of the results when there is an exact match on let's say a product name. If you search "pretty plastic doll" in the review with the both fuzzy logic (*and* and *or*) you get the following result: 

Screen Shot 2021-05-21 at 11 45 01

However with removing the fuzzy "or" 

image

Here the trade off I think is how likely are use search searching for things that they know exactly what the name is and my hunch on this is probably not as they would not use the search. 

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


app/services/elasticsearch_query.rb
```diff
@@ -86,8 +86,9 @@ def match_exact_fields
       multi_match: {
         query: query,
         fields: @exact_fields,
-        type: "phrase",
-        fuzzy_transpositions: "false"
+        type: "cross_fields",
+        fuzzy_transpositions: "false",
+        operator: "and"
```
KyleMacPherson, line 91: I don't think we need any changes here.

spec/services/elasticsearch_query_spec.rb
```diff
@@ -148,5 +148,31 @@ def perform_search
         it_behaves_like "finds the relevant investigation"
       end
     end
+
+    context "with a multi term search" do
+      let!(:investigation_one)   { create(:allegation, description: "multi term search") }
+      let!(:investigation_two)   { create(:allegation, description: "multi term") }
+      let!(:investigation_three) { create(:allegation,  description: "term search") }
+      let!(:investigation_four)  { create(:allegation,  description: "multi search") }
```
KyleMacPherson, line 156: Could we add in an investigation that we expect to be filtered out here? I think that would really help the test

StupidCodeFactory: @KyleMacPherson fixing the specs made `investigation_three` and `investigation_four` to be filtered out

app/services/elasticsearch_query.rb
StupidCodeFactory: will give it a try

nasirkhanbeis (approved):
Tested this on review app, looking good as expected. Nice work!

StupidCodeFactory:
[(3) Search by AND not OR](https://trello.com/c/7XpAgQVe/270-3-search-by-and-not-or)

app/services/elasticsearch_query.rb
```diff
@@ -86,7 +86,7 @@ def match_exact_fields
       multi_match: {
         query: query,
         fields: @exact_fields,
-        type: "phrase",
+        type: "cross_fields",
```
KyleMacPherson, line 89: I don't think we need "cross_fields" here because we are only using a single exact field

spec/services/elasticsearch_query_spec.rb
KyleMacPherson: Nice nice

nasirkhanbeis:
re-tested this after new changes, looks good. Let's merge this in when devs are happy.

KyleMacPherson (approved):

PR #1396, opened 2021-05-20
Refactor Activity model to remove inherited notification callbacks

slorek:
This change removes the notification callback functionality in the `Activity` model, which is inherited by all subclasses in the `AuditActivity` namespace.

Since all but 6 of these classes have now been refactored to use service class interfaces for their creation and the notification logic moved to those, it is no longer appropriate for all of the subclasses to inherit this functionality, which can cause undesirable and unexpected behaviour.

Moving the logic to a concern which is included only in those classes which have not yet been refactored makes it easier to identify the outstanding work, removing some overhead in the refactoring and is a long-awaited improvement.

For reference, the remaining classes that depend on this are:

`CommentActivity`
`AuditActivity::Alert::Add`
`AuditActivity::Correspondence::AddEmail`
`AuditActivity::Document::Add`
`AuditActivity::Document::Destroy`
`AuditActivity::Document::Update`

I'm working on removing this dependency from the last three in a separate branch.

Also improves one curious design choice where by case comments inherit from `Activity` and are in fact the default type in the database schema. The comments controller now instantiates the correct `CommentActivity` class explicitly which improves understanding and mitigates potential unexpected behaviour.

### DIFF
--- a/app/models/audit_activity/test/request.rb
+++ b/app/models/audit_activity/test/request.rb
@@ -1,5 +1,9 @@
 # Recording of test requests is deprecated - existing data is still supported
 class AuditActivity::Test::Request < AuditActivity::Test::Base
+  def readonly?
+    true
+  end
+
 private
 
   def subtitle_slug
--- a/app/models/audit_activity/test/request.rb
+++ b/app/models/audit_activity/test/request.rb
@@ -1,17 +1,5 @@
+# Recording of test requests is deprecated - existing data is still supported
 class AuditActivity::Test::Request < AuditActivity::Test::Base
-  def self.from(test)
-    title = "Test requested: #{test.product.name}"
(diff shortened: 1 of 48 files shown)


app/models/audit_activity/test/request.rb
```diff
@@ -1,17 +1,5 @@
+# Recording of test requests is deprecated - existing data is still supported
```
StupidCodeFactory, line 1: Suggestion: might be worth adding

```ruby
def readonly?
  true
end
```

StupidCodeFactory (approved):

KyleMacPherson (approved):
nice

nasirkhanbeis (approved):
looking good, don't see any regressions. nice work!

PR #1400, opened 2021-05-24
Fix bug when user has been deleted their session is not destroyed

KyleMacPherson:
https://trello.com/c/gt8s9UU3/530-2-invalidate-user-active-sessions-when-same-user-is-deleted-in-backend

## Description
Prior to this change when a user is deleted manually (using the delete user rake task) while said user is logged in, we do not stop this user from being able to the make changes for the remainder of their session.

This change logs the user out if the user has a deleted_at value (which is populated when the delete user rake task is run).

Note: The feature test I included is pretty weird and can be removed if you prefer.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -23,6 +23,11 @@ class ApplicationController < ActionController::Base
   def set_current_user
     return unless user_signed_in?
 
+    if current_user.deleted_at
+      sign_out current_user
+      redirect_to '/'
+    end
+
     User.current = current_user
   end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -23,7 +23,7 @@ class ApplicationController < ActionController::Base
   def set_current_user
     return unless user_signed_in?
 
-    if current_user.deleted_at
+    if current_user.deleted?
       sign_out current_user
       redirect_to "/"
     end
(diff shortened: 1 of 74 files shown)


nasirkhanbeis (approved):
Tested this on review app, works fine as expected.

app/controllers/application_controller.rb
```diff
@@ -23,6 +23,11 @@ class ApplicationController < ActionController::Base
   def set_current_user
     return unless user_signed_in?
 
+    if current_user.deleted_at
```
slorek, line 26:
```suggestion
    if current_user.deleted?
```

spec/features/delete_user_mid_session_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  let(:user)          { create(:user, :activated, has_viewed_introduction: true) }
+
+  scenario "Deleting a signed in user" do
+    sign_in user
+    visit "/cases/"
+
+    expect(page).to have_link "Your account"
+    expect(page).to have_link "Sign out"
+
+    DeleteUser.call(user: user)
```
slorek, line 13:
```suggestion
    DeleteUser.call!(user: user)
```

slorek (approved):

PR #1403, opened 2021-05-25
Improve searchability re index when product is updated

StupidCodeFactory:
## Description

The intent of this PR is to update the relevant elastic search indexes when updating a product.
the product needs to be re-index as well as the investigations linked to the product.

I've written the tests but they are not passing yet.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis (approved):

slorek (approved):
Suggest adding a comment to explain why the `sleep` was added. This might be brittle, but I can't think of an alternative in a feature spec context right now.

PR #1413, opened 2021-06-01
Change radio button on edit corrective action

KyleMacPherson:
https://trello.com/c/qk5MQogL/688-update-edit-corrective-action-behaviour-and-activity-log-file-upload-removal

## Description
It seems the only remaining work that needed to be done on the ticket above was to dynamically render the text for the no option on the question "Are there any files related to the action?" to say "Remove attached file" if there is an existing file attached to the corrective action being edited.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/services/update_corrective_action.rb
+++ b/app/services/update_corrective_action.rb
@@ -12,6 +12,7 @@ class UpdateCorrectiveAction
     corrective_action.transaction do
       corrective_action.document.detach unless related_file
       replace_attached_file             if file_changed?
+      remove_blanks_from_changes
 
       if any_changes?
         corrective_action.save!
@@ -46,6 +47,10 @@ private
 
   def any_changes?
     file_changed? || changes.except(:related_file, :existing_document_file_id, :document).any?
+  end
+
+  def remove_blanks_from_changes
+    changes.delete_if { |k,v| v.first.blank? && v.last.blank? }
   end
 
   def file_changed?
--- a/app/services/update_corrective_action.rb
+++ b/app/services/update_corrective_action.rb
@@ -52,7 +52,7 @@ private
     return false if document.nil? && related_file
     return false if document == @previous_attachment
 
-    [document, @previous_attachment].compact.any?
+    document.checksum != @previous_attachment.checksum
   end
 
   def new_file_description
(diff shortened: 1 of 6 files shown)


nasirkhanbeis:
spotted minor issue, when same file is attached then audit log shows empty log record. Expected not to show any changes when replaced with same file.

nasirkhanbeis (approved):
Tested this on review app, finally appears all good. Nice work @KyleMacPherson 
there are some further improvements needed which I will raise as separate ticket.

app/services/update_corrective_action.rb
```diff
@@ -12,6 +12,7 @@ def call
     corrective_action.transaction do
       corrective_action.document.detach unless related_file
       replace_attached_file             if file_changed?
+      remove_blanks_from_changes
```
slorek, line 15: I think this should go in the form class as it is sanitising user input. We have a helper called `SanitizationHelper` which can do this. An example of usage is here https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/app/forms/corrective_action_form.rb#L54

app/views/investigations/activities/corrective_action/_update.erb
```diff
@@ -29,7 +29,11 @@
   <% end %>
 
   <% if activity.business_updated? %>
```
slorek, line 31: Not a big one, but as a rule we should be reading the `business_id` in the `metadata` property. The extra relationships on activity models were added a long time ago to patch over an inflexible model, and will ideally be removed in the near future.

slorek (changes requested):

app/services/update_corrective_action.rb
KyleMacPherson: This example usage is actually in place here, the issue is that we were setting this value to nil via the sanitisation (and also via a method in the update_corrective_action class, but the record on creation is being saved with the file description value as an empty string. I have made some changes to make this uniform across both create and update and removed any sanitisation being done in the update_corrective_action class.

Let me know what you think!

slorek (approved):

PR #1420, opened 2021-06-07
Comments refactor

KyleMacPherson:
https://trello.com/c/bfggCJdX/1055-refactor-comments-audit-log

## Description
Refactor comments to no longer rely on 'ActivityNotification' 

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


app/controllers/comments_controller.rb
```diff
@@ -1,41 +1,33 @@
...
-    @comment = CommentActivity.new
-    @investigation = @investigation.decorate
+    redirect_to investigation_activity_path(@investigation), flash: { success: "Comment was successfully added." }
   end
 
 private
 
   def set_investigation
-    @investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
+    @investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id]).decorate
```
slorek, line 25: We've learned it's safer to delay the `decorate` call until just before any view is rendered because the decorator can alter behaviour of the object. In this case we're passing a decorated object to the `AddCommentToCase` service class which in the past has caused us unexpected issues.

app/controllers/comments_controller.rb
```diff
@@ -1,41 +1,33 @@
...
     authorize @investigation, :show?
   end
 
   def comment_activity_params
-    params.require(:comment_activity).permit(:body).tap do |p|
+    params.require(:comment_form).permit(:body).tap do |p|
       p[:body] = ActionController::Base.helpers.sanitize(p[:body])
```
slorek, line 31: Suggestion: Might be worth moving the sanitisation to the form object in a `before_validation` block, since we do this elsewhere to strip whitespace for example.

app/models/audit_activity/comment/add_comment.rb
```diff
@@ -0,0 +1,15 @@
...
+    }
+  end
+
+  def title(*)
+    "Add comment"
+  end
+
+  def subtitle_slug
+    "Added"
+  end
```
slorek, line 14: Suggestion: Move these presentation methods to a decorator and use I18n.

app/models/audit_activity/comment/add_comment.rb
```diff
@@ -0,0 +1,15 @@
+class AuditActivity::Comment::AddComment < AuditActivity::Base
```
slorek, line 1: Suggestion: Since the comment is on a case it would make sense for this to live in the `AuditActivity::Investigation` namespace.

app/models/comment_activity.rb
```diff
@@ -1,25 +1,2 @@
 class CommentActivity < Activity
```
slorek, line 1: I think we should create a one-off Rake task to migrate the existing `CommentActivity` data to the new class with the body move to the metadata property, and add a `TODO` comment to this file to remove it once the data is migrated.

app/services/add_comment_to_case.rb
```diff
@@ -0,0 +1,49 @@
...
+        recipient.name,
+        recipient.email,
+        email_update_text(recipient),
+        email_subject
+      ).deliver_later
+    end
+  end
+
+  def email_update_text(recipient)
+    "#{source.show(recipient)} commented on the allegation."
```
slorek, line 43:
```suggestion
    "#{source.show(recipient)} commented on the #{investigation.case_type}."
```

slorek (changes requested):

nasirkhanbeis:
comment h1 in activity log is wrong, it should  read as "Comment added"
image

app/models/audit_activity/comment/add_comment.rb
```diff
@@ -0,0 +1,15 @@
...
+  def self.build_metadata(body)
+    {
+      comment_text: body
+    }
+  end
+
+  def title(*)
+    "Add comment"
```
slorek, line 9:
```suggestion
    "Comment added"
```
To match existing behaviour.

app/models/audit_activity/comment/add_comment.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def subtitle_slug
+    "Added"
```
slorek, line 13:
```suggestion
    "Comment added"
```
To match existing behaviour.

app/controllers/comments_controller.rb
```diff
@@ -16,13 +17,14 @@ def create
       })
     )
 
+    @investigation = @investigation.decorate
```
slorek, line 24: No need to decorate before a redirect.

slorek (approved):

PR #1432, opened 2021-06-10
Add email refactor

KyleMacPherson:
https://trello.com/c/WXs23KiP/1056-refactor-addemail-audit-log

## Description
Refactor the add email audit log in order to remove the need to include `ActivityNotification` and update the code to follow our pattern of using forms and service objects rather than calling the logic on the model itself.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis:
spotted few issues: 
1. When replaced with same file, it shows changes with same file- expected not to show any changes when replaced with sam file
2. When To or From is updated , it records empty log
3. when email subject and body is entered and remove the email file then audit log shows Body in new line, layout issue. should show in same line I think. see attached screen shot. 
image

app/services/add_email_to_case.rb
```diff
@@ -23,14 +24,56 @@ def call
...
+  def audit_activity_metadata
+    AuditActivity::Correspondence::AddEmail.build_metadata(email)
+  end
+
+  def create_audit_activity(correspondence, investigation)
+    activity = AuditActivity::Correspondence::AddEmail.create!(
+      metadata: audit_activity_metadata,
+      source: UserSource.new(user: User.current),
+      investigation: investigation,
+      title: correspondence.overview,
```
slorek, line 43: The `title` attribute in the database is deprecated. Let the decorator generate the title. This should be stored as `nil`.

app/services/add_email_to_case.rb
```diff
@@ -23,14 +24,56 @@ def call
...
+      correspondence: correspondence
+    )
+
+    activity.attach_blob(correspondence.email_file.blob, :email_file) if correspondence.email_file.attached?
+    activity.attach_blob(correspondence.email_attachment.blob, :email_attachment) if correspondence.email_attachment.attached?
+  end
+
+  def sanitize_text(text)
+    return text.to_s.strip.gsub(/[*_~]/) { |match| "\\#{match}" } if text
+  end
```
slorek, line 53:
```suggestion
```

Doesn't appear to be used.

slorek (changes requested):
Let's leave migration of old data for another day as it will require quite complex pattern matching.

app/decorators/audit_activity/correspondence/add_email_decorator.rb
```diff
@@ -0,0 +1,13 @@
+module AuditActivity
+  module Correspondence
+    class AddEmailDecorator < AuditActivity::Correspondence::BaseDecorator
+      def correspondence_date
+        Date.parse(metadata["correspondence_date"]).to_s(:govuk)
+      end
+
+      def title(_user)
+        object.metadata["overview"]
```
slorek, line 9: Please change: `metadata` will be `nil` for legacy activity records. This will throw an error in that case.

app/decorators/audit_activity/correspondence/add_email_decorator.rb
```diff
@@ -0,0 +1,13 @@
+module AuditActivity
+  module Correspondence
+    class AddEmailDecorator < AuditActivity::Correspondence::BaseDecorator
+      def correspondence_date
+        Date.parse(metadata["correspondence_date"]).to_s(:govuk)
```
slorek, line 5: Please change: `metadata` will be `nil` for legacy activity records. This will throw an error in that case.

slorek (changes requested):

app/decorators/audit_activity/correspondence/add_email_decorator.rb
```diff
@@ -0,0 +1,15 @@
...
+  module Correspondence
+    class AddEmailDecorator < AuditActivity::Correspondence::BaseDecorator
+      def correspondence_date
+        Date.parse(metadata["correspondence_date"]).to_s(:govuk) if object.metadata
+        object.correspondence_date
+      end
+
+      def title(_user)
+        return object.metadata["overview"] if object.metadata
+        object.title
```
KyleMacPherson, line 11: Is this what you were looking for @slorek?

app/decorators/audit_activity/correspondence/add_email_decorator.rb
```diff
@@ -0,0 +1,17 @@
+module AuditActivity
+  module Correspondence
+    class AddEmailDecorator < AuditActivity::Correspondence::BaseDecorator
+      def correspondence_date
+        return Date.parse(metadata["correspondence_date"]).to_s(:govuk) if object.metadata
+
+        object.correspondence_date
```
slorek, line 7: I can't see where this is defined in the model. For legacy activities it is buried in a text string in the `body`.

KyleMacPherson: I was confused by why this works so have been looking into it and it turns out we don't need this to handle old AuditActivity entries for adding email. This is because in `app/views/investigations/activities/_activity_card.html.erb` on lines 38-44 we render the body as a whole if we do not have any metadata for an activity:

```  <% else %>
    <% if activity.metadata %>
      <%= render "investigations/activities/#{activity.template_name}", activity: activity %>
    <% else  %>
      <% if activity.body %>
        <%= markdown simple_format(activity.body) %>
      <% end %>```

slorek (approved):

PR #1438, opened 2021-06-14
Fix error when user navigates back to the change ownership form after submission

slorek:
https://trello.com/c/IQznMlyf/1060-fix-error-when-navigating-back-to-change-owner-form-after-form-submission

## Description
According to examples of this exception on Sentry, an error occurs when a GET request is made to the `confirm` step. This probably happens when the user clicks the back button after submitting the form, since this should have corresponding form params.

This PR introduces a check and redirect to the first step page to avoid the error.

### DIFF
--- a/app/views/investigations/ownership/confirm.html.erb
+++ b/app/views/investigations/ownership/confirm.html.erb
@@ -23,7 +23,7 @@
       <%= render "minimal_investigation_heading", investigation: @investigation, title: page_heading %>
       <p>
         You are changing the case owner to:
-        <strong><%= @potential_owner.decorate.display_name(viewer: current_user) %></strong>
+        <strong><%= @potential_owner.display_name(viewer: current_user) %></strong>
       </p>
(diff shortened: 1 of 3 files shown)


app/views/investigations/ownership/confirm.html.erb
```diff
@@ -23,7 +23,7 @@
       <%= render "minimal_investigation_heading", investigation: @investigation, title: page_heading %>
       <p>
         You are changing the case owner to:
-        <strong><%= @potential_owner.decorate.display_name(viewer: current_user) %></strong>
+        <strong><%= @potential_owner.display_name(viewer: current_user) %></strong>
```
slorek, line 26: The object is already decorated as you can see above.

KyleMacPherson (approved):

nasirkhanbeis (approved):
Had a quick look on review app, looks good as expected. I think there was a bug ticket in backlog on this so if we see can move it to done.

PR #1443, opened 2021-06-17
Refactor alert

KyleMacPherson:
https://trello.com/c/4SR4Nbxi/1061-refactor-add-alert

## Description
This PR refactors the add alert workflow to follow our existing pattern of using forms and service classes to make changes to resources rather than callbacks and legacy audit log logic.

This change also removes the wizard flow from the alerts controller.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis (approved):
Looks good as expected!

app/controllers/investigations/alerts_controller.rb
```diff
@@ -1,55 +1,63 @@
...
+    authorize @investigation, :investigation_restricted?
 
-      redirect_to next_wizard_path
-    else
-      render_wizard
-    end
+    @alert_form = AlertForm.new(alert_request_params.merge(investigation_url: investigation_url(@investigation)))
+    render :new unless @alert_form.valid?
+
+    @user_count = number_with_delimiter(User.active.count, delimiter: ",")
```
slorek, line 27:
```suggestion
    set_user_count
```

app/controllers/investigations/alerts_controller.rb
```diff
@@ -1,55 +1,63 @@
...
 
   def create
-    @alert.save!
-    redirect_to investigation_path(@investigation), flash: { success: "Email alert sent to #{@user_count} users" }
+    set_investigation
+    authorize @investigation, :investigation_restricted?
+
+    @alert_form = AlertForm.new(alert_request_params)
+    set_user_count
+    if @alert_form.valid?
```
slorek, line 38: Please change: need to handle invalid input.

app/controllers/investigations/alerts_controller.rb
```diff
@@ -1,55 +1,63 @@
...
-    session.delete(:alert)
-  end
-
   def set_investigation
     @investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
     authorize_investigation
   end
 
   def authorize_investigation
     authorize @investigation, :send_email_alert?
```
slorek, line 60:
```suggestion
  def set_and_authorize_investigation
    @investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
    authorize @investigation, :send_email_alert?
```

app/controllers/investigations/alerts_controller.rb
```diff
@@ -1,55 +1,63 @@
 class Investigations::AlertsController < ApplicationController
-  include Wicked::Wizard
   include Pundit
   include ActionView::Helpers::NumberHelper
 
-  steps :about_alerts, :compose, :preview
-
   before_action :set_investigation
```
slorek, line 5:
```suggestion
```
This is duplicated in the individual action methods.

app/controllers/investigations/alerts_controller.rb
```diff
@@ -1,55 +1,63 @@
...
+  def about
+    set_investigation
+    @investigation = @investigation.decorate
   end
 
-  def show
-    render_wizard
+  def new
+    set_investigation
+    authorize @investigation, :investigation_restricted?
```
slorek, line 14: Suggestion: We could possibly avoid having to do this by using a named finder scope on `Investigation` which limits to non-restricted cases.

app/forms/alert_form.rb
```diff
@@ -0,0 +1,34 @@
+class AlertForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveModel::Serialization
+  include ActiveModel::Dirty
+
+  attribute :summary
+  attribute :description
+  attribute :user_count
```
slorek, line 9:
```suggestion
```
Doesn't appear to be used.

app/controllers/investigations/alerts_controller.rb
```diff
@@ -1,55 +1,63 @@
...
+
+    @alert_form = AlertForm.new(alert_request_params)
+    set_user_count
+    if @alert_form.valid?
+      AddAlert.call!(
+        @alert_form.attributes.merge({
+          investigation: @investigation,
+          user: current_user,
+          user_count: @user_count
```
slorek, line 43: Suggestion: Let the service object calculate this. Simplifies the interface so that it can be re-used more easily. The query will be cached anyway so there's very little performance implication.

app/forms/alert_form.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def default_summary
+    "Product safety alert: "
+  end
+
+  def default_description
+    "\r\n\r\n\r\nMore details can be found on the case page: #{investigation_url}"
+  end
+
+  def summary_validation
```
slorek, line 23:
```suggestion
private

  def summary_validation
```

I'd move the `default_` methods to private as well if you implement the getter override suggestion.

app/views/investigations/alerts/new.html.erb
```diff
@@ -22,14 +22,14 @@
...
-                 value: (@alert.summary.presence || @alert.default_summary) %>
+                 value: (@alert_form.summary.presence || @alert_form.default_summary) %>
 
       <%= render "form_components/govuk_textarea",
                  form: form,
                  key: :description,
                  rows: 12,
                  maxlength: 1000,
-                 value: (@alert.description.presence || @alert.default_description),
+                 value: (@alert_form.description.presence || @alert_form.default_description),
```
slorek, line 32: Suggestion: Override the `description` getter in the form to return `default_description` if no `description` set. Removes logic from the view.

app/views/investigations/alerts/new.html.erb
```diff
@@ -22,14 +22,14 @@
                  key: :summary,
                  label: { text: "Alert subject", classes: "govuk-label--m" },
                  hint: { text: "This will be used in the email subject line." },
-                 value: (@alert.summary.presence || @alert.default_summary) %>
+                 value: (@alert_form.summary.presence || @alert_form.default_summary) %>
```
slorek, line 25: Suggestion: Override the `summary` getter in the form to return `default_summary` if no summary set. Removes logic from the view.

app/models/audit_activity/alert/add.rb
```diff
@@ -1,25 +1,17 @@
...
-      source: UserSource.new(user: User.current),
-      investigation: alert.investigation
-    )
+  def self.build_metadata(alert, user_count)
+    {
+      user_count: user_count,
+      subject: alert.summary,
+      date_sent: alert.created_at,
+      description: alert.description
+    }
```
slorek, line 11: I think we would want to store an `Integer` here - I think at the moment it would store a string passed through from the controller. I don't think we need to pass through the count at all.

```suggestion
  def self.build_metadata(alert)
    {
      user_count: User.active.count,
      subject: alert.summary,
      date_sent: alert.created_at,
      description: alert.description
    }
```

app/models/audit_activity/alert/add.rb
```diff
@@ -1,25 +1,17 @@
 class AuditActivity::Alert::Add < AuditActivity::Base
-  include ActivityNotification
   extend ActionView::Helpers::NumberHelper
```
slorek, line 2:
```suggestion
```
I don't think we need this any more.

app/services/add_alert.rb
```diff
@@ -0,0 +1,43 @@
...
+    )
+
+    send_alert_emails
+    create_audit_activity
+  end
+
+private
+
+  def send_alert_emails
+    emails = User.active.map(&:email)
```
slorek, line 24: Suggestion: Put this as a new method in the `EntitiesToNotify` concern so that it is in one place.

app/services/add_alert.rb
```diff
@@ -0,0 +1,43 @@
...
+    SendAlertJob.perform_later(emails, subject_text: summary, body_text: description)
+  end
+
+  def audit_activity_metadata
+    AuditActivity::Alert::Add.build_metadata(alert, user_count)
```
slorek, line 29:
```suggestion
    AuditActivity::Alert::Add.build_metadata(alert)
```
Don't need to pass this through from the controller.

app/views/investigations/activities/alert/_add.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="govuk-body">
+  From: <b>Office for Product Safety and Standards</b><br>
+  To: <b><%= "All users (#{activity.metadata["user_count"]} people)" %></b><br>
+  Subject: <b><%= activity.metadata["subject"] %></b><br>
+  Date sent: <b><%= activity.metadata["date_sent"] %></b><br>
+
+  <br><%= activity.metadata["description"] %>
+</div>
```
slorek, line 8: Use `strong` and `p` instead of `b` and double`br` tags.

slorek (changes requested):
Good work, just a bit of tidying up needed.

slorek:
Bug spotted: If submitting the form with default values, the following is displayed:

Screenshot 2021-06-23 at 11 42 06

app/controllers/investigations/alerts_controller.rb
KyleMacPherson: 👍 One side affect of this is that we won't render the number of users that the alert was sent to in the flash success message, I personally don't think it is an issue but just FYI

app/views/investigations/alerts/new.html.erb
KyleMacPherson: I'm struggling trying to get this to work, is there an example of where we do this is the codebase already?

app/controllers/investigations/alerts_controller.rb
```diff
@@ -1,55 +1,63 @@
...
+  def preview
+    set_investigation
+    authorize @investigation, :investigation_restricted?
 
-      redirect_to next_wizard_path
-    else
-      render_wizard
-    end
+    @alert_form = AlertForm.new(alert_request_params.merge(investigation_url: investigation_url(@investigation)))
+    render :new unless @alert_form.valid?
```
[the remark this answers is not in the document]
KyleMacPherson: Good spot, thanks Steve. Added assertions to check for this in the feature spec.

app/controllers/investigations/alerts_controller.rb
slorek: We could potentially return this count in the service object context for use in the flash message.

app/views/investigations/alerts/new.html.erb
slorek: There's some similar override usage in https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/app/forms/risk_assessment_form.rb and https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/app/forms/product_form.rb#L60-L85.

I think the method in the form class would look something like:

```ruby
def description
  return super if self[:description].present?
  default_description
end
```

app/controllers/investigations/alerts_controller.rb
```diff
@@ -1,55 +1,59 @@
...
+        })
+      )
 
-  def set_investigation
-    @investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
-    authorize_investigation
+      @investigation = @investigation.decorate
+      redirect_to investigation_path(@investigation), flash: { success: "Email alert sent to users" }
+    else
+      render :new
```
slorek, line 48:
```suggestion
      @investigation = @investigation.decorate
      render :new
```
I think we need to decorate here.

app/controllers/investigations/alerts_controller.rb
```diff
@@ -1,55 +1,59 @@
...
+    end
   end
 
-  def authorize_investigation
+private
+
+  def set_and_authorize_investigation
+    @investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
```
slorek, line 55:
```suggestion
    @investigation = Investigation.not_private.find_by!(pretty_id: params[:investigation_pretty_id])
```
Adding a `not_private` scope to `Investigation` would negate the need to use `authorize @investigation, :investigation_restricted?` everywhere.

app/forms/alert_form.rb
```diff
@@ -0,0 +1,35 @@
...
+  include ActiveModel::Serialization
+  include ActiveModel::Dirty
+
+  attribute :summary
+  attribute :description
+  attribute :investigation_url
+
+  validate :summary_validation
+  validate :description_validation
+
```
slorek, line 13: Suggestion: Something like this would negate the need to check and revert to default in the view.

```ruby
def summary
  self[:summary] || default_summary
end

def description
  self[:description] || default_description
end

private
```

app/views/investigations/alerts/new.html.erb
slorek, line 25:
```suggestion
                 value: @alert_form.summary %>
```

app/views/investigations/alerts/new.html.erb
slorek: I've added an example as a suggestion.

slorek, line 32:
```suggestion
                 value: @alert_form.description,
```

slorek (approved):

PR #1457, opened 2021-06-30
Product export

KyleMacPherson:
https://trello.com/c/6sSYxB6t/1069-expand-product-export

## Description
This PR allows users to export products and associations into an xlsx file with tabs related to each associated model.

The user will request an export via the "Export as spreadsheet button", they will then be redirected to the products page and shown a flash message letting them know that the export is being generated and that they will soon receive and email. This email will be sent once the report is generated and will link them to a page within our app that allows them to download the report.

This approach has been taken as the export would take an extremely long time, and we have already seen the less complex investigation export timing out for users.

I have continued to use XLSX as it is consistent with our investigations export and it can be opened in google sheets or with open source apps like Libra Office as well as Excel so I think it fits our requirements.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis:
Had another quick look,  new changes as recommended are looking good. just few minor issues(not show stoppers) I've noticed otherwise all good. Nice to see product name on supporting information tabs.
1. Cases id column not show square brackets[] for empty values, also would be nice to get rid of square brackets and make it consistent with other column values
2. Also marking column show values in square brackets which I think not needed, we can show comma separate for more than one value.

slorek (changes requested):
We should investigate exporting the data in ODF format in the first instance. Government strategy is to prefer using open formats which are interoperable.

app/decorators/risk_assessment_decorator.rb
```diff
@@ -39,6 +39,13 @@ def activity_cell_partial(_viewing_user)
     "activity_table_cell_with_link"
   end
 
+  def assessed_by
+    return Business.find(assessed_by_business_id).legal_name if assessed_by_business_id
+    return Team.find(assessed_by_team_id).name if assessed_by_team_id
```
slorek, line 44: This would potentially be more performant if the associated records are included in the search query.
```suggestion
    return assessed_by_business.legal_name if assessed_by_business_id
    return assessed_by_team.name if assessed_by_team_id
```

app/controllers/products_controller.rb
```diff
@@ -17,13 +17,10 @@ def index
...
       end
-      format.csv do
+      format.xlsx do
         authorize Product, :export?
 
-        results = search_for_products.includes(:investigations)
-        @products = ProductDecorator.decorate_collection(results)
-
-        render csv: @products, filename: "products"
+        @products = search_for_products(20, %i[investigations corrective_actions risk_assessments test_results]).sort
```
slorek, line 23: I think we would want to do something like this to potentially speed up the rendering of the risk assessment `assessed_by`.

```suggestion
        @products = search_for_products(20, [:investigations, :corrective_actions, risk_assessments: [:assessed_by_business, :assessed_by_team], :test_results]).sort
```
It would be worth going through each column and their rendering methods to identify other speed improvements. We have had issues with the database query timing out on the investigations export Excel before. Given there are about as many products as cases and the fact more data is rendered on this endpoint, the performance is likely to be a lot worse so we should try to optimise the query as much as possible.

slorek (approved):
I think we just need to carefully review the performance of this before deployment, but otherwise it looks good. It might be worth benchmarking with production data to validate this.

slorek:
I think we want @DH02 to review this page 
Screenshot 2021-07-20 at 10 49 03

app/models/product_export.rb
```diff
@@ -0,0 +1,135 @@
+class ProductExport < ApplicationRecord
+  has_one_attached :export_file
+
+  def export(products)
+    Axlsx::Package.new do |p|
+      book = p.workbook
+      book.add_worksheet name: "product_info" do |sheet_investigations|
```
slorek, line 7: Suggestion: Each of these `add_worksheet` calls could be delegated to a separate method rather than inline code blocks.

app/models/product_export.rb
```diff
@@ -0,0 +1,135 @@
...
+            product.investigations.first.try(:non_compliant_reason),
+            product.investigations.first.try(:risk_level)
+          ], types: :text
+        end
+      end
+
+      book.add_worksheet name: "test_results" do |sheet_investigations|
+        sheet_investigations.add_row %w[product_id legislation standards date_of_test result how_product_failed further_details product_name]
+        products.each do |product|
+          product.tests.where(type: "Test::Result").sort.each do |test_result|
```
slorek, line 68: Suggestion: Create a new `test_results` relationship to clean this up.

app/models/product_export.rb
```diff
@@ -0,0 +1,135 @@
...
+              corrective_action.duration,
+              corrective_action.geographic_scopes.join(","),
+              corrective_action.details,
+              product.name
+            ], types: :text
+          end
+        end
+      end
+
+      p.serialize(Rails.root.join("product_export.xlsx"))
```
slorek, line 130: Please change: I think we would want to give this file a unique name as there could be multiple jobs running in parallel which may cause a race condition/exception. It probably should be stored as a `Tempfile` so that it is cleaned up automatically.

app/models/product_export_worker.rb
```diff
@@ -0,0 +1,14 @@
+class ProductExportWorker < ApplicationJob
+  def perform(products, product_export_id, user)
```
slorek, line 2: Suggestion: It may be a good idea to catch exceptions and report them to Sentry in this job so we become aware of any problems. Otherwise it will fail silently and the user will never receive the email they're expecting.

app/views/product_exports/show.html.erb
```diff
@@ -0,0 +1,3 @@
+<% if @product_export.export_file.attached? %>
+  Click <%= link_to "here", rails_blob_path(@product_export.export_file, disposition: 'attachment') %> to download your product export.
```
slorek, line 2: Suggestion: it may be useful to users to know the date and time the export was generated since emails are kept indefinitely and a user could click on an old export link.

slorek (changes requested):
A further suggestion: It may or may not be useful to clean up old exports after a period of time since the data will no longer be representative. In this case we could show an 'expired' error on the show page.

Good job, it's working in my testing.

There are a bunch of suggestions here and only one necessary change. Please review with @tobyhughes and @DH02.

app/views/product_exports/show.html.erb
slorek: Another minor one: the Products tab should be pre-selected on this page.

DH02:
@KyleMacPherson @slorek 

Ignoring the header and footer (which will form a separate review another time) and only looking at the attached screenshot, I have some simple amendments - sorry, if some are repeated (as I can't view the HTML): below.

1. To ensure we are using the correct font family I have added a class to the P tag (govuk-body)
2. To ensure the link is styled correctly (and supports various GDS states, I have added classes to the A tag
3. To ensure the download behaves as expected I've added the "download" boolean attribute. (You can also write x="x" in lowercase) 
4. For accessibility we should avoid "Click here" text - I have changed the link text
5. As the link sits in the page by itself I've increased the font size
6. Without seeing it in the page I've assumed a little margin (classes) won't go a miss too. (You can remove one or more "margin" classes if required, or change 8 to 6 if required?)

<p class="govuk-body govuk-!-font-size-19 govuk-!-margin-top-6 govuk-!-margin-bottom-8" download>
        <a href="..." class="govuk-link govuk-link--no-visited-state">Download</a> your product export.
</p>

Thanks

app/controllers/product_exports_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  before_action :set_search_params, only: %i[generate]
+
+  def generate
+    authorize Product, :export?
+
+    @products = search_for_products(Product.count, [:investigations, :test_results, { corrective_actions: [:business], risk_assessments: %i[assessed_by_business assessed_by_team] }]).sort
+    product_export = ProductExport.create!
+    ProductExportWorker.perform_later(@products, product_export.id, current_user)
+    redirect_to products_path, flash: { success: "Your product export is being prepared. You will receive an email when your export is ready to download." }
```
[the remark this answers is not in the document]
KyleMacPherson: Good call. Done!

app/controllers/product_exports_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def show
+    authorize Product, :export?
+
+    @product_export = ProductExport.find(params[:id])
```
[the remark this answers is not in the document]
KyleMacPherson: I did think about this, but I believe every user who has authorisation to export data can already get every record so it doesn't expose anything they can't already see. IMO this means that it isn't a problem.

app/models/product_export.rb
```diff
@@ -0,0 +1,135 @@
...
+                                        product_code
+                                        subcategory
+                                        updated_at
+                                        webpage
+                                        when_placed_on_market
+                                        reported_reason
+                                        hazard_type
+                                        non_compliant_reason
+                                        risk_level
+                                        name]
```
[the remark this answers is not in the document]
KyleMacPherson: These are actually just the header names Toby suggested

app/models/product_export.rb
KyleMacPherson: Done!

app/models/product_export.rb
KyleMacPherson: Good idea, done!

app/models/product_export_worker.rb
KyleMacPherson: Done!

app/views/product_exports/show.html.erb
KyleMacPherson: Both done, good call(s)

KyleMacPherson:
> @KyleMacPherson @slorek
> 
> Ignoring the header and footer (which will form a separate review another time) and only looking at the attached screenshot, I have some simple amendments - sorry, if some are repeated (as I can't view the HTML): below.
> 
> 1. To ensure we are using the correct font family I have added a class to the P tag (govuk-body)
> 2. To ensure the link is styled correctly (and supports various GDS states, I have added classes to the A tag)
> 3. To ensure the download behaves as expected I've added the "download" boolean attribute. (You can also write x="x" in lowercase)
> 4. For accessibility we should avoid "Click here" text - I have changed the link text
> 5. As the link sits in the page by itself I've increased the font size
> 6. Without seeing it in the page I've assumed a little margin (classes) won't go a miss too. (You can remove one or more "margin" classes if required, or change 8 to 6 if required?)
> 
> ```
> <p class="govuk-body govuk-!-font-size-19 govuk-!-margin-top-6 govuk-!-margin-bottom-8" download>
>         <a href="..." class="govuk-link govuk-link--no-visited-state">Download</a> your product export.
> </p>
> ```
> 
> Thanks

Thanks Darren, this is a screenshot after implementing your suggestions, and some of Steve's from the comments.

Screen Shot 2021-07-20 at 15 51 23

DH02:
> > @KyleMacPherson @slorek
> > Ignoring the header and footer (which will form a separate review another time) and only looking at the attached screenshot, I have some simple amendments - sorry, if some are repeated (as I can't view the HTML): below.
> > 
> > 1. To ensure we are using the correct font family I have added a class to the P tag (govuk-body)
> > 2. To ensure the link is styled correctly (and supports various GDS states, I have added classes to the A tag)
> > 3. To ensure the download behaves as expected I've added the "download" boolean attribute. (You can also write x="x" in lowercase)
> > 4. For accessibility we should avoid "Click here" text - I have changed the link text
> > 5. As the link sits in the page by itself I've increased the font size
> > 6. Without seeing it in the page I've assumed a little margin (classes) won't go a miss too. (You can remove one or more "margin" classes if required, or change 8 to 6 if required?)
> > 
> > ```
> > <p class="govuk-body govuk-!-font-size-19 govuk-!-margin-top-6 govuk-!-margin-bottom-8" download>
> >         <a href="..." class="govuk-link govuk-link--no-visited-state">Download</a> your product export.
> > </p>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Thanks
> 
> Thanks @DH02, this is a screenshot after implementing your suggestions, and some of @slorek's from the comments.
> 
> Screen Shot 2021-07-20 at 15 51 23

That looks fine, but a full-stop at the end would be nice?! Thanks.

app/models/product_export.rb
KyleMacPherson: Done

app/models/product_export.rb
KyleMacPherson: I have given the files unique names but I found testing to be very difficult using temp files so I have reverted back to not using them.

app/models/product_export.rb
```diff
@@ -0,0 +1,157 @@
...
+      add_risk_assessments_worksheet(book, products)
+
+      add_corrective_actions_worksheet(book, products)
+
+      p.serialize(Rails.root.join("product_export_#{id}.xlsx"))
+
+      export_file.attach(io: File.open(Rails.root.join("product_export_#{id}.xlsx")), filename: "products_export.xlsx")
+
+    rescue StandardError => e
+      Sentry.capture_exception(e)
```
slorek, line 21: I think this should be in the parent job class as we might want to run this in the console for debugging for example.

Query: How do we want to handle failure? This will send us an alert but the user would not receive the email they're expecting. Should we re-raise the exception  after reporting to Sentry so that it enters the Sidekiq retry queue? Or are we going to requeue them manually once any issue is fixed?

app/controllers/product_exports_controller.rb
```diff
@@ -0,0 +1,21 @@
+class ProductExportsController < ApplicationController
+  include ProductsHelper
+  helper_method :sort_column, :sort_direction
+
+  before_action :set_search_params, only: %i[generate]
+
+  def generate
+    authorize Product, :export?
+
+    @products = search_for_products(Product.count, [:investigations, :test_results, { corrective_actions: [:business], risk_assessments: %i[assessed_by_business assessed_by_team] }]).sort
```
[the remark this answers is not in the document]
slorek: On reflection I think we can leave this here for now as it will make testing easier and failure will happen immediately for the user if there is an issue with the query. However I think we should add a `TODO` comment here to move it to the job once we're satisfied it's working well.

app/models/product_export_worker.rb
slorek: I think the exception capture should be in the job class to ensure it's only done in this context.

slorek (changes requested):

app/models/product_export.rb
KyleMacPherson: Hmmm re-raising them sounds like the way to go, good idea.

slorek (approved):

PR #1471, opened 2021-07-20
Update ID generation to avoid conflicts after deleting an Investigation

stdstaff:
https://trello.com/c/hRe9rGwp/1084-change-case-id-generation-to-allow-cases-to-be-deleted-without-breaking-the-service

## Description
Currently, if you delete a case which was created during the current month, users will be prevented from creating any new cases for the rest of the month.

This is because the case ID (`pretty_id`) generation method takes the total count of cases from the current month in the database and appends 1, which will cause a duplicate ID to be generated.

Instead, this change takes the latest case ID (`pretty_id`) for the month and adds one to it. This is independent of the actual record `id` or the number of cases, and ensures case numbers will always increment.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


slorek (approved):

spec/services/create_case_spec.rb
```diff
@@ -65,6 +65,18 @@
...
+      context "with previous investigations" do
+        let(:investigations) { build_list(:enquiry, 3, creator: user) }
+
+        it "generates a successive pretty_id", :aggregate_failures do
+          investigations.each { |i| described_class.call(investigation: i, user: user) }
+          expect(investigations.pluck(:pretty_id).map { |id| id.split("-").last })
+            .to eq(%w[0001 0002 0003])
+          result
+          expect(investigation.reload.pretty_id).to end_with("0004")
+        end
```
KyleMacPherson, line 77: Looks good overall, but I think this test would pass before the change.

KyleMacPherson (approved):

spec/services/create_case_spec.rb
```diff
@@ -65,6 +65,27 @@
...
+
+        it "does not generate a conflicting pretty_id following a case deletion", :aggregate_failures do
+          investigations.second.destroy!
+          expect(Investigation.pluck(:pretty_id).map { |id| id.split("-").last })
+            .to eq(%w[0001 0003])
+          result
+          expect(investigation.reload.pretty_id).to end_with("0004")
+        end
+      end
```
KyleMacPherson, line 87: LGTM!

nasirkhanbeis (approved):
looks good. Tested on review and compared on stage. works fine

PR #1493, opened 2021-07-30
Refactor document attachments flow to fix bugs and address tech debt

slorek:
https://trello.com/c/Zh44dkYD/430-8-missing-field-validation-on-add-attachment-page

## Description
The ticket above ostensibly required a small bug fix, but fixing it was not easily possible due to the burden of technical debt and inconsistencies with the rest of the application. This pull request also fixes a number of long standing issues in line with our overall tech debt remediation strategy.

### Before
The current file upload process is a two-page form. On the first page, the user selects the file:

Screenshot 2021-07-30 at 12 25 25

On the second page, the user enters the file metadata:

Screenshot 2021-07-30 at 12 25 38

The file preview always the shows "processing" placeholder and will never update. There is really no value in splitting this flow into multiple pages, and doing so introduced some technical issues with the `DocumentsFlowController`:

* Use of a `Wicked::Wizard`. We are actively working to remove `Wicked::Wizard` usage from the application
* Weird nondescript route names like `new_investigation_new_path`
* Use of the `FileConcern` which has a very clunky interface

As well as this the pages were not implemented very well with no error highlighting of the fields due to not using the GOV.UK form helpers.

### After
Screenshot 2021-07-30 at 12 24 59

The attachment upload flow has been reworked to a single page form which allows the user to replace the file and properly highlights field errors.

### Other UI improvements
* When uploading/editing/removing an attachment from a product, the "product' nav item is now selected
* The user can replace the file once uploaded
* The add/edit forms now have a consistent design
* The remove page properly shows the attachment's metadata
* User will no longer lose their file uploaded when pressing 'Back' on the second page
* The forms use the proper GOV.UK form components and should be accessible
* Improved the validation error messages
* Added a flash message on successful update of a file

### Code improvements
* Removed the final use case for `User.current`, which has caused a number of problems for the team over nearly two years. A follow-up PR will be opened immediately to remove the last incorrect usage and the gem.
* Removed the final use case for the `ActivityNotification` concern - now, finally, all activity models are no longer sending update emails via callbacks and all business workflows are encapsulated in service objects
* Store document add/edit/remove activity data in the `metadata` column - the end of a very long-running refactoring exercise across all activity models.
* Removed usage of `ActivityAttachable` which was not providing any benefit and creating problems with implementation. It's still used elsewhere but we could consider removing it later on
* Refactored the activity card view template so that the individual view templates for the relevant activity type determine how  an attached image is shown, and the activity model doesn't need to implement `ActivityAttachable` or inherited methods.
* The route names for the documents controller now make sense

### Post-deployment steps
The activity metadata should be migrated by running the following commands:
```
CLASS_NAME=AuditActivity::Document::Add rake activities:update_metadata
CLASS_NAME=AuditActivity::Document::Destroy rake activities:update_metadata
CLASS_NAME=AuditActivity::Document::Update rake activities:update_metadata
``` 
This will allow us to remove the data migration methods from the classes and tidy up the database.

A PR will also be opened to remove the RequestStore gem.

### Possibilities
* We can now think about how audit logging should ideally work without risking breaking the business logic
* We should tidy up any remaining legacy data and migrate it to `metadata` in the `Activity` hierarchy. We can then remove the `title` and `body` columns from the activities table.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


app/controllers/application_controller.rb
```diff
@@ -129,11 +129,16 @@ def render_404_page
 
   def highlight_businesses?
     return true if params[:controller].start_with?("businesses")
-    return true if params[:controller] == "documents_flow" && params[:business_id]
+    return true if params[:controller] == "documents" && params[:business_id]
+  end
+
+  def highlight_products?
```
slorek, line 135: It's probably better if the controller determines this but that's left for a later refactor.

app/decorators/audit_activity/corrective_action/update_decorator.rb
```diff
@@ -81,10 +81,6 @@ def new_business
     Business.find(metadata.dig("updates", "business_id", 1))
   end
 
-  def attachment
-    @attachment ||= (signed_id = metadata.dig("updates", "existing_document_file_id", 1)) && ActiveStorage::Blob.find_signed!(signed_id)
-  end
```
slorek, line 86: This was moved to the model as it is not a presentational concern and resolves an issue with inherited methods in `Activity`.

app/controllers/documents_controller.rb
```diff
@@ -1,72 +1,125 @@
...
+    @file = @parent.documents.find(params[:id])
+
+    DeleteDocument.call!(
+      document: @file,
+      parent: @parent,
+      user: current_user
+    )
 
-    return redirect_to @parent, flash: { success: "File was successfully removed" } unless @parent.is_a? Investigation
+    flash[:success] = "File was successfully removed"
```
stdstaff, line 99: We've used a translated string for the flash message when a file is added. Do we need one for when it's removed?

app/models/audit_activity/document/update.rb
```diff
@@ -1,38 +1,87 @@
...
 
-  def self.no_change?(document, previous_data)
-    document.metadata[:title] == previous_data[:title] && document.metadata[:description] == previous_data[:description]
+  def metadata
+    migrate_metadata_structure
   end
 
-  def self.title_changed?(document, previous_data)
-    document.metadata[:title] != previous_data[:title]
+  def title(_user)
```
stdstaff, line 15: Takes an argument which isn't used?

slorek: Yes, it's a 'common' interface, some of the other activity classes do use the user argument to construct the title.

nasirkhanbeis:
UI improvements looks good. Add product image link on products tab is broken otherwise all good.
Also, we need further improvements on below scenarios

1. Add/edit/remove attachment on products page, when attachment is added/updated, It should return to attachments tab rather product overview
2. Add/edit/remove image in images tab, when I click back on add image page, then It should return back to images tab
3. Edit image/document, would be nice to add replace file so user can edit image/file too
4. Remove attachment/image page shows empty type? not sure if need to show this metadata? also row URL I think should read as File?

nasirkhanbeis (approved):
Tested add attachment new behaviour, edit and delete + validations. looking good as expected.  
There are some improvements needed which I will raise separately. 
Also attachment in OPSS case flow has diff version where it does't capture title and description but uses file name as title. I think there's a ticket in backlog.

PR #1497, opened 2021-08-02
Move case (investigation) export out into a background job

stdstaff:
https://trello.com/c/krLYkQfw/1057-502-when-trying-to-export-case-list

## Description

Users were sometimes seeing a 502 error when trying to export a [large] case list. This PR copies the approach from #1457 and moves the export generation into a background job.

The user will request an export via the "Export as spreadsheet link", they will then be redirected to the case list page and shown a flash message letting them know that the export is being generated and that they will soon receive and email. This email will be sent once the report is generated and will link them to a page within our app that allows them to download the export.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis (approved):
looks good as expected!

KyleMacPherson (approved):

PR #1502, opened 2021-08-04
Fix outstanding update email audit log bugs

KyleMacPherson:
Prior to this PR we had a few small bugs relating to our audit log when a user updated a correspondence email which can be found in this ticket: https://trello.com/c/jBDFSBlh/1063-bugs-in-updateemail-audit-log

Note: the 3rd bug is around the displaying of email changes, I think the actual issue is that we display the changing of the email file, and the changing of the user supplied content of the email under the header `email` which I think causes confusion when the user edits and changes from an email file to manually inputting the details and vice versa. In order to improve user experience I have put the email file changes under `email file:` while leaving the manually entered user email details under `email:`

Screen Shot 2021-08-04 at 15 52 39

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


stdstaff (approved):
LGTM. 👍

PR #1507, opened 2021-08-09
Product edit bug

KyleMacPherson:
https://trello.com/c/un106KXq/1030-change-product-details-link-should-only-be-available-to-collaborator-and-case-owner

## Description
Prior to this change anyone could edit any product. This change ensures that only users with the ability to edit the corresponding investigation can edit the product.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis (approved):
looks good as expected.
1. Validate non-team members should not be able to edit or remove a product
2. Validate only team members should be able to edit or remove a product
3. Validate when I add a new team with edit permissions, then the team should be edit or remove a product
4. Validate when I add a new team with view only, then added team should not be able to edit/remove product.
5. Validate when a product is not associated with any case, all users should be able to edit.

stdstaff (approved):

PR #1519, opened 2021-08-12
Add `http://` to any url’s that do not include it before rendering li…

KyleMacPherson:
…nk in corrective action show page

https://trello.com/c/rcz7I2BA/1109-links-not-opening-correctly

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/decorators/corrective_action_decorator.rb
+++ b/app/decorators/corrective_action_decorator.rb
@@ -53,4 +53,8 @@ class CorrectiveActionDecorator < ApplicationDecorator
   def display_medium_title_text_size?
     page_title.length > MEDIUM_TITLE_TEXT_SIZE_THRESHOLD
   end
+
+  def online_recall_information
+    object.online_recall_information.first(4) == "http" ? object.online_recall_information : "http://#{object.online_recall_information}"
+  end
 end
--- a/app/decorators/corrective_action_decorator.rb
+++ b/app/decorators/corrective_action_decorator.rb
@@ -55,8 +55,8 @@ class CorrectiveActionDecorator < ApplicationDecorator
   end
 
   def online_recall_information
-    return unless object.online_recall_information.present?
-    
+    return if object.online_recall_information.blank?
+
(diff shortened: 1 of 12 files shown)


nasirkhanbeis:
Test scenarios: 

1. Verify URL starting with www
2. Verify URL starting with HTTP
3. Verify URL starting with https://www.google.com
4. Verify URL works in both the activity page and corrective action view page.
5. Verify URL starting with domain gov.uk

app/decorators/corrective_action_decorator.rb
```diff
@@ -53,4 +53,8 @@ def duration
   def display_medium_title_text_size?
     page_title.length > MEDIUM_TITLE_TEXT_SIZE_THRESHOLD
   end
+
+  def online_recall_information
+    object.online_recall_information.first(4) == "http" ? object.online_recall_information : "http://#{object.online_recall_information}"
```
stdstaff, line 58: I wonder if we should check for the `//` instead, or parse it, in case [the URI](https://datatracker.ietf.org/doc/html/rfc3986#section-3) starts with http, e.g. [httpstatuses.com](https://httpstatuses.com/), or another scheme like `ftp://` — rare, but not impossible?

```suggestion
    URI.parse(object.online_recall_information).scheme.present? ? object.online_recall_information : "http://#{object.online_recall_information}"
```

N.B. May need a guard clause..?
```rb
return unless object.online_recall_information.present?
```

app/helpers/investigations/corrective_actions_helper.rb
```diff
@@ -35,5 +35,9 @@ def online_recall_information_text_for(online_recall_information, has_online_rec
 
       I18n.t(".#{has_online_recall_information}", scope: %i[investigations corrective_actions helper has_online_recall_information])
     end
+
+    def online_recall_information_url(online_recall_information)
+      online_recall_information.first(4) == "http" ? online_recall_information : "http://#{online_recall_information}"
```
stdstaff, line 40: And I guess same treatment for this one 🙂

stdstaff (approved):

PR #1520, opened 2021-08-12
Move case export query into background job and find in batches

stdstaff:
https://trello.com/c/M37lIPUp/1112-full-case-export-timing-out

## Description

This PR moves a large database query from a synchronous controller request to the background job for case export. This was an existing TODO item which needed to be cleared up since during very large exports, users were still seeing a timeout error, caused by a SQL timeout.

Rather than passing the records around, this PR passes `case_ids` as an array to the background job, and then allows the exporter to find cases in batches. This has the same effect in the backend as pagination would on the frontend.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/controllers/case_exports_controller.rb
+++ b/app/controllers/case_exports_controller.rb
@@ -6,12 +6,11 @@ class CaseExportsController < ApplicationController
   def generate
     authorize Investigation, :export?
 
-    # TODO: move this query to the job once we are satisfied that this is working well.
-    answer = search_for_investigations
-    investigations = answer.records(includes: [:complainant, :products, :owner_team, :owner_user, { creator_user: :team }]).to_a
+    investigation_ids = search_for_investigations.records.ids
 
     case_export = CaseExport.create!
-    CaseExportJob.perform_later(investigations, case_export.id, current_user)
+    CaseExportJob.perform_later(investigation_ids, case_export.id, current_user)
+
     redirect_to investigations_path(q: params[:q]), flash: { success: "Your case export is being prepared. You will receive an email when your export is ready to download." }
   end
 
--- a/app/controllers/case_exports_controller.rb
+++ b/app/controllers/case_exports_controller.rb
@@ -9,7 +9,7 @@ class CaseExportsController < ApplicationController
     investigation_ids = search_for_investigations.records.ids
 
     case_export = CaseExport.create!
-    CaseExportJob.perform_later(investigation_ids, case_export.id, current_user)
+    CaseExportJob.perform_later(investigation_ids, case_export, current_user)
 
     redirect_to investigations_path(q: params[:q]), flash: { success: "Your case export is being prepared. You will receive an email when your export is ready to download." }
   end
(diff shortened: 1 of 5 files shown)


app/controllers/case_exports_controller.rb
```diff
@@ -6,12 +6,11 @@ class CaseExportsController < ApplicationController
   def generate
     authorize Investigation, :export?
 
-    # TODO: move this query to the job once we are satisfied that this is working well.
-    answer = search_for_investigations
-    investigations = answer.records(includes: [:complainant, :products, :owner_team, :owner_user, { creator_user: :team }]).to_a
+    investigation_ids = search_for_investigations.records.ids
```
slorek, line 9: I think the usage of `records` here is redundant and will be slower. I think the `id` attribute should be available from the result set without instantiating ActiveRecord instances.

app/controllers/case_exports_controller.rb
```diff
@@ -6,12 +6,11 @@ class CaseExportsController < ApplicationController
...
 
     case_export = CaseExport.create!
-    CaseExportJob.perform_later(investigations, case_export.id, current_user)
+    CaseExportJob.perform_later(investigation_ids, case_export.id, current_user)
```
slorek, line 12:
```suggestion
    CaseExportJob.perform_later(investigation_ids, case_export, current_user)
```

app/jobs/case_export_job.rb
```diff
@@ -1,9 +1,9 @@
 class CaseExportJob < ApplicationJob
-  def perform(cases, case_export_id, user)
+  def perform(case_ids, case_export_id, user)
     case_export = CaseExport.find(case_export_id)
     return unless case_export
```
slorek, line 4:
```suggestion
  def perform(case_ids, case_export, user)
```

app/controllers/case_exports_controller.rb
slorek: The original suggestion was to move the whole query to the background job, if we can include `SearchHelper` in the job. Not sure how difficult that will be but I don't think there's a need to perform the search at all in the controller (unless this is needed for feedback to the user).

stdstaff: `search_for_investigations` returns an [`Elasticsearch::Model::Response::Response`](https://www.rubydoc.info/gems/elasticsearch-model/Elasticsearch/Model/Response/Response) — the only method for `ids` I could find was on `records`, although it doesn't load or instantiate the records at that point, instead it uses the internal hash, so should be performant:
```rb
# File 'lib/elasticsearch/model/response/records.rb', line 52
def ids
  response.response['hits']['hits'].map { |hit| hit['_id'] }
end
```

slorek: Sweet

slorek (approved):

app/models/case_export.rb
```diff
@@ -1,21 +1,22 @@
 class CaseExport < ApplicationRecord
   include CountriesHelper
 
+  FIND_IN_BATCH_SIZE = 1000
```
slorek, line 4:
```suggestion
  # Helps to manage the database query execution time within the PaaS imposed limits
  FIND_IN_BATCH_SIZE = 1000
```

PR #1534, opened 2021-08-18
Revert "Add `http://` to any url’s that do not include it before rendering li…"

KyleMacPherson:
Reverts UKGovernmentBEIS/beis-opss-psd#1519


slorek (approved):
Please add rationale to the description

PR #1536, opened 2021-08-19
Validate that online recall info is a valid url

KyleMacPherson:
https://trello.com/c/DnB0oCdm/1120-corrective-action-online-recall-info-bug

## Description
Prior to this PR, users could enter anything into the corrective action online recall information box. This change ensures that users can only enter valid URLs.

The next step after deployment of this PR will be to sanitize the 50 or so existing corrective actions with online recall information to make sure they are valid URLs

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis:
@KyleMacPherson  I think we need a URL validator that can detect the URL pattern. eg. asos.com is also a valid URL
image

nasirkhanbeis (approved):
Looks good to me however there are some loose ends where it is possible to enter an invalid URL. 
Suggestion: Would be nice to improve URL format validation
Eg. it considers a valid URL for  anything which is prefixed with HTTP://
 https://nasir@com is considered a valid URL
https://nasir,com.....
Otherside, when a user enters a valid URL www.tesco.com it returns an error message that I think is not user-friendly.

app/forms/corrective_action_form.rb
```diff
@@ -109,10 +109,20 @@ def related_file_attachment_validation
     end
   end
 
+  def valid_url?
+    uri = URI.parse(online_recall_information)
+    uri.is_a?(URI::HTTP) && !uri.host.nil?
```
stdstaff, line 114: I had to check 😂 TIL `HTTPS` is a subclass of `HTTP` 👍
```rb
URI::HTTPS < URI::HTTP
=> true
```

stdstaff (approved):

PR #1541, opened 2021-08-23
Download webp package on PaaS environments to support WEBP image processing

slorek:
https://trello.com/c/G1122diI/1115-fix-webp-image-processing

Uses the CloudFoundry multi-buildpack feature and the experimental [CloudFoundry apt buildpack](https://github.com/cloudfoundry/apt-buildpack), which is similar to the Heroku apt buildpack, to automatically install the webp library on GOV.UK PaaS-hosted environments. 

This should resolve the silent failures which occur when users upload WEBP formatted images. Currently, thumbnail previews are not generated. This should ensure that thumbnails are generated correctly.

### DIFF
--- a/apt.yml
+++ b/apt.yml
@@ -1,0 +1,3 @@
+---
+packages:
+- webp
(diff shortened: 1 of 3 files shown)


stdstaff (approved):
👍

PR #1550, opened 2021-08-24
Fix flaky spec

slorek:
Fixes a flaky spec. Examples of failed runs:

https://github.com/UKGovernmentBEIS/beis-opss-psd/runs/3410174737?check_suite_focus=true
https://github.com/UKGovernmentBEIS/beis-opss-psd/runs/3411967920?check_suite_focus=true
https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/1540/checks?check_run_id=3402305870

This looks to be because the feature spec was duplicated from `export_products_spec.rb` but the model name wasn't changed.


KyleMacPherson (approved):
Oh nice spot @slorek

PR #1552, opened 2021-08-25
Fix password reset bug

KyleMacPherson:
https://trello.com/c/82WTbqac/1077-reset-password-bug

## Description

It seems this bug only shows up in prod/review apps and not locally so it is a little tricky to fully understand but to reproduce the sentry issue found on the trello card I visited `/password/edit?reset_password_token=INCORRECTTOKENHERE`. The outcome of which can be seen in the before screenshot.

To address this, I changed the order of the guard clauses in the user passwords controller edit action in order to ensue that we would not render `signed_in_as_another_user` if the token was invalid. This looks to have fixed the issue but I am completely sure why there should be differences between the local and prod environments in this case.

## Before:
Screen Shot 2021-08-25 at 15 30 15

## After: 
Screen Shot 2021-08-25 at 15 31 32

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


slorek:
@KyleMacPherson I think the core issue is the use of invalid reset tokens. Your change avoids but doesn't completely fix the underlying issue, which is that the `wrong_user?` method assumes `user_with_reset_token` is not nil.

I'm happy to reorder the guard clauses as they do make more sense this way, but I think we should add a guard to this method to check `user_with_reset_token` is a valid `User`.

slorek (approved):

PR #1553, opened 2021-08-26
Fix 500 errors when user is not logged in

KyleMacPherson:
https://trello.com/c/66HvnJqc/1001-500-error-thrown-when-logged-out-instead-of-404

## Description
Prior to this change, when a user was not logged in and the server returned a 500 error the application would not render a custom 500 error page. This change corrects that issue.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


slorek (approved):

slorek:
Suggest adding an automated test for this scenario as I think we had a similar issue with these pages in the past.

PR #1568, opened 2021-09-02
Fix resend team invite link

KyleMacPherson:
https://trello.com/c/LsYW8k1M/738-resend-invite-link-with-2fa-returns-404-page-not-found-error

## Description
The issue described in the linked Trello card is likely going to rear it's head in a number of places and it is due to our 2FA flow.

When the user needs to re-authenticate via 2FA we save the path that they were trying to access, and then redirect to it after they successfully authenticate. The issue described in the linked Trello card will arise each time the user makes a request using an action other than GET prior to re-authenticating. This is because redirects always use GET. For example, the reason that the page shows 404 in this case is because the resend_team_invitation endpoint can only be accessed via a PUT request and so we get a 404 when we redirect to it.

Some options I considered were:

1. leave it as is.
2. always redirect to homepage
3. redirect to homepage if request is not a GET
4. use obscure [repost gem](https://github.com/vergilet/repost)
5. make request explicitly in our controller action (?)

In this PR I have chosen to go for option 3 because I think it is low risk and predictable while still eliminating the 404 issues and giving the user the expected behaviour in most cases. It is definitely an improvement as requests that would 404 now just go to the homepage, however it is probably not perfect as far as UX goes.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


app/controllers/concerns/secondary_authentication_concern.rb
```diff
@@ -12,7 +12,7 @@ def require_secondary_authentication(redirect_to: request.fullpath)
 
     if user_id_for_secondary_authentication && !secondary_authentication_present?
       user = User.find(user_id_for_secondary_authentication)
-      session[:secondary_authentication_redirect_to] = redirect_to
+      session[:secondary_authentication_redirect_to] = request.get? ? redirect_to : "/"
```
slorek, line 15:
```suggestion
      if request.get?
        session[:secondary_authentication_redirect_to] = redirect_to
        session[:secondary_authentication_redirect_to_post_blocked] = false
      else
        session[:secondary_authentication_redirect_to_post_blocked] = true
      end
```
This would mean that if a POST request is sent the redirect would remain the last GET endpoint they accessed (which would be the form).

The `secondary_authentication_redirect_to_post_blocked` flag could be used to render a flash message when the user is redirected to the form informing them that they have to try again.

slorek (commented):
This does prevent the error but I added a suggestion to improve the user experience for consideration.

app/controllers/concerns/secondary_authentication_concern.rb
```diff
@@ -12,11 +12,19 @@ def require_secondary_authentication(redirect_to: request.fullpath)
...
+      else
+        session[:secondary_authentication_redirect_to] ||= "/"
+        session[:secondary_authentication_redirect_to_post_blocked] = true
+      end
       session[:secondary_authentication_user_id] = user_id_for_secondary_authentication
       auth = SecondaryAuthentication.new(user)
       auth.generate_and_send_code(current_operation)
       redirect_to new_secondary_authentication_path
+    else
+      session[:secondary_authentication_redirect_to] = redirect_to
```
slorek, line 27: Not sure if we need to check if it's a GET request here too. Maybe this and the preceding block should be moved up outside of the conditional?

app/controllers/secondary_authentications_controller.rb
```diff
@@ -51,10 +51,12 @@ def secondary_authentication
...
-    if session[:secondary_authentication_redirect_to]
-      redirect_to session.delete(:secondary_authentication_redirect_to)
+    redirect_to_path = session[:secondary_authentication_redirect_to] || root_path_for(current_user)
+
+    if session[:secondary_authentication_redirect_to_post_blocked]
+      redirect_to redirect_to_path, flash: { warning: "Request could not be completed. Please try again." }
     else
-      redirect_to root_path_for(current_user)
+      redirect_to redirect_to_path
     end
```
slorek, line 60:
```ruby
flash[:warning] = "Request could not be completed. Please try again." if session[:secondary_authentication_redirect_to_post_blocked]

redirect_to redirect_to_path
```

app/controllers/concerns/secondary_authentication_concern.rb
```diff
@@ -10,16 +10,23 @@ module SecondaryAuthenticationConcern
   def require_secondary_authentication(redirect_to: request.fullpath)
     return unless Rails.configuration.secondary_authentication_enabled
 
+    session[:secondary_authentication_redirect_to] = redirect_to if path_can_be_redirected_to?
```
KyleMacPherson, line 13: I have decided to only set secondary_authentication_redirect_to on GET requests, because we can't redirect to the POST/PUTS anyway so this seems sensible to me and we can leave the logic to handle when secondary_authentication_redirect_to is nil to the secondary authentication controller

slorek (approved):

PR #1571, opened 2021-09-03
Use new template with capitalised case type

KyleMacPherson:
https://trello.com/c/C34ub2CM/171-case-created-notifications-have-incorrect-capitalisation-in-subject

## Description
I have added a new template here which uses a variable for the capitalized case type in the subject.
https://www.notifications.service.gov.uk/services/b1dab0f5-4651-4af8-9f15-fa8143ff338d/templates/b5457546-9633-4a9c-a844-b61f2e818c24

And made a small change to the mailer to use the new template and added in the new variable.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


stdstaff (approved):

slorek (changes requested):
Change looks good but I would expect some test to fail. it looks like the emails aren't included in the relevant case creation feature specs. Please could you add this?

slorek (approved):

PR #1582, opened 2021-09-13
Smoke tests

KyleMacPherson:
https://trello.com/c/avm0IcDg/401-smoke-test-on-deployment

## Description
This PR adds smoke testing after we deploy a review app, to staging and to prod. I have made the prod deploy dependent on all preceding steps succeeding so if the deployment to staging fails, or the smoke tests, then we should expect to see the prod deployment not go through.

I have not however found a good way to automatically roll back the deployment if the smoke test fails on prod. Perhaps that is something we can look at in a further PR.

Note: If there are bugs in the changes I have made to the deploy script then deployments will be blocked until we fix/rollback the change. I think this is worth noting before we deploy to ensure we do not possibly hold up urgent changes.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


.github/workflows/deploy.yml
```diff
@@ -133,6 +173,43 @@ jobs:
...
+        key: ${{ runner.os }}-gems-no-dev-${{ hashFiles('**/Gemfile.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-gems-
+
+    - name: Install RubyGems
+      if: success()
+      run: |
+        gem install bundler --no-doc
+        bundle config path vendor/bundle
+        bundle check || bundle install --without development --jobs=4 --retry=3
```
slorek, line 194: Do we need to repeat these steps which were already performed prior to the staging deployment?

.github/workflows/deploy.yml
```diff
@@ -79,7 +80,44 @@ jobs:
...
+      if: success()
+      env:
+        ELASTICSEARCH_URL: localhost
+        RAILS_ENV: test
+        SMOKE_USER: ${{secrets.SMOKE_USER}}
+        SMOKE_PASSWORD: ${{secrets.SMOKE_PASSWORD}}
+        SMOKE_RELAY_CODE_URL: ${{secrets.SMOKE_RELAY_CODE_URL}}
+        SMOKE_RELAY_CODE_USER: ${{secrets.SMOKE_RELAY_CODE_USER}}
+        SMOKE_RELAY_CODE_PASS: ${{secrets.SMOKE_RELAY_CODE_PASS}}
+        IS_REVIEW_APP: "false"
```
slorek, line 113: I think we can remove the `IS_REVIEW_APP` variable now since the tests won't run as part of the main Rspec suite.

.github/workflows/deploy.yml
```diff
@@ -79,7 +80,44 @@ jobs:
...
+      if: success()
+      run: |
+        gem install bundler --no-doc
+        bundle config path vendor/bundle
+        bundle check || bundle install --without development --jobs=4 --retry=3
+
+    - name: Run Smoke Tests
+      if: success()
+      env:
+        ELASTICSEARCH_URL: localhost
```
slorek, line 106: I think this is redundant.

.github/workflows/deploy.yml
```diff
@@ -79,7 +80,44 @@ jobs:
...
+        RAILS_ENV: test
```
slorek, line 107: I think this is redundant since we don't load the Rails application in the smoke test.

.github/workflows/deploy.yml
```diff
@@ -133,6 +173,43 @@ jobs:
...
+        SMOKE_USER: ${{secrets.SMOKE_USER}}
+        SMOKE_PASSWORD: ${{secrets.SMOKE_PASSWORD}}
+        SMOKE_RELAY_CODE_URL: ${{secrets.SMOKE_RELAY_CODE_URL}}
+        SMOKE_RELAY_CODE_USER: ${{secrets.SMOKE_RELAY_CODE_USER}}
+        SMOKE_RELAY_CODE_PASS: ${{secrets.SMOKE_RELAY_CODE_PASS}}
+        RUN_SMOKE: "true"
+        IS_REVIEW_APP: "false"
+        SMOKE_TEST_URL: "https://www.product-safety-database.service.gov.uk/"
```
slorek, line 208: Same comments as above regarding `ELASTICSEARCH_URL`, `RAILS_ENV`, `RUN_SMOKE`, `IS_REVIEW_APP` variables.

.github/workflows/review-apps.yml
```diff
@@ -92,3 +92,40 @@ jobs:
...
+        RAILS_ENV: test
+        SMOKE_USER: ${{secrets.SMOKE_USER}}
+        SMOKE_PASSWORD: ${{secrets.SMOKE_PASSWORD}}
+        SMOKE_RELAY_CODE_URL: ${{secrets.SMOKE_RELAY_CODE_URL}}
+        SMOKE_RELAY_CODE_USER: ${{secrets.SMOKE_RELAY_CODE_USER}}
+        SMOKE_RELAY_CODE_PASS: ${{secrets.SMOKE_RELAY_CODE_PASS}}
+        IS_REVIEW_APP: "true"
+        SMOKE_TEST_URL: https://psd-pr-${{ github.event.number }}.london.cloudapps.digital
+        SMOKE_TEST_REVIEW_APP_BASIC_AUTH_USER: ${{secrets.SMOKE_TEST_REVIEW_APP_BASIC_AUTH_USER}}
+        SMOKE_TEST_REVIEW_APP_BASIC_AUTH_PASSWORD: ${{secrets.SMOKE_TEST_REVIEW_APP_BASIC_AUTH_PASSWORD}}
```
slorek, line 128: Same comments as above regarding `RAILS_ENV`, `IS_REVIEW_APP` variables.

smoke_test/cases_page_spec.rb
```diff
@@ -0,0 +1,57 @@
+require 'capybara/rspec'
+require 'capybara/mechanize'
+require "webmock"
+
+Capybara.register_driver :mechanize do |app|
+  Capybara::Mechanize::Driver.new(proc {})
+end
```
slorek, line 7: Suggestion: Add a comment to explain what this is for.

smoke_test/cases_page_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    session.visit(ENV["SMOKE_TEST_URL"])
+    expect(session).to have_css("h1", text: "Product Safety Database")
+    session.visit("#{ENV["SMOKE_TEST_URL"]}/sign-in")
+    session.fill_in "Email address", with: ENV["SMOKE_USER"]
+    session.fill_in "Password", with: ENV["SMOKE_PASSWORD"]
+    session.click_button "Continue"
+
+    expect(session).to have_current_path(/\/two-factor/)
+    expect(session).to have_content("Check your phone")
+    #
```
slorek, line 24: ✂️

smoke_test/cases_page_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      sleep attempts * 10
+    end
+
+    session.click_link "Cases"
+    expect(session).to have_css(".govuk-grid-row.psd-case-card:nth-child(1)")
+    expect(session).to have_css(".govuk-grid-row.psd-case-card:nth-child(10)")
+  end
+end
+
+def smoke_complete_secondary_authentication_with(fake_code, session)
```
slorek, line 42: Why `fake_code`?

smoke_test/cases_page_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    session.fill_in "Email address", with: ENV["SMOKE_USER"]
+    session.fill_in "Password", with: ENV["SMOKE_PASSWORD"]
+    session.click_button "Continue"
+
+    expect(session).to have_current_path(/\/two-factor/)
+    expect(session).to have_content("Check your phone")
+    #
+    attempts = 0
+    loop do
+      code = get_code.scan(/\d{5}/).first
```
slorek, line 27: I think this probably should be done within the `get_code` method. Suggest exiting early here if it doesn't return the code to make it clearer.

slorek (commented):
Looks good overall, just a few refinement suggestions to clean it up before release. I will test a failure tomorrow on the review app.

.github/workflows/deploy.yml
KyleMacPherson: I don't think so, good shout. Removed.

.github/workflows/deploy.yml
KyleMacPherson: We still need `IS_REVIEW_APP` because we need to enter basic auth credentials for review apps.

slorek:
Tested failures where the Elasticsearch index was empty (no results) and did not exist (500 error) - working as expected.

I think we need to wait for the index seeds task to complete before running the smoke test. This may produce a race condition error for someone in future. An example of polling to check the task status can be found here: https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/.github/workflows/review-apps-delete.yml#L34

slorek (approved):

PR #1589, opened 2021-09-17
2FA bugfix

KyleMacPherson:
https://trello.com/c/gtNFj1nu/1135-request-could-be-not-be-completed-error-on-home-page

## Description
Prior to this PR, if a user was idle on the 2FA page until the 2FA code expired, then requested a new 2FA code, then they would be redirected to the home page but a warning error would show which told them we could not process their request properly.

This message was set to show on any non get requests that we tried to redirect from 2FA. It is triggered in this case because the 'resend text message' button triggers a post request. However in this specific case we do not want to show the flash message even though the last user request was a POST.

This PR addresses that issue

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis (approved):
I've tested this on the review app with Kyle's help in the backend to expire the session, works fine as expected. The warning message isn't shown upon login which is definitely an improvement. It is quite tricky to test this. I will keep an eye on stage if I see any unusual behavior

slorek (approved):
Hopefully there's no other similar cases, if so we will have to rethink the solution.

PR #1620, opened 2021-10-08
Business export to run in background job

KyleMacPherson:
https://trello.com/c/zfCJ8ctt/1172-improve-business-export-functionality

## Description
This PR brings the same functionality we use for case and product exports to the business export. The export is generated in a background job, the user receives an email when it is ready to download and a link which they visit to download the export.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis (approved):
Re-tested this on the review app, looks good as expected.

slorek (approved):

PR #1621, opened 2021-10-12
Query ES for businesses in batches

KyleMacPherson:
https://trello.com/c/1HAsNX6G/1179-optimise-elasticsearch-query-for-business-export

## Description
This PR utilises elasticsearch's `search_after` functionality to retrieve all our businesses in 10k batches, which hopefully should reduce memory related issues now and in the future which could come from us retrieving >10k records in one query.

Note: I ordered by ID rather than created_at, as I found it much clearer to use ID as the search after point and the order will be the same regardless I believe. I don't think the ordering should matter much anyway as we are producing an excel file which the user can order as they see fit.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/helpers/businesses_helper.rb
+++ b/app/helpers/businesses_helper.rb
@@ -15,7 +15,7 @@ module BusinessesHelper
   end
 
   def search_for_businesses_in_batches
-    Business.search_for_businesses_in_batches(search_query)
+    Business.search_in_batches(search_query)
   end
 
   def business_export_params
--- a/app/helpers/businesses_helper.rb
+++ b/app/helpers/businesses_helper.rb
@@ -14,6 +14,32 @@ module BusinessesHelper
       .records
   end
 
+  def search_for_businesses_in_batches(size = 1000)
+    businesses= []
+    after = 0
+
+    loop do
+      query = search_query.build_query([], [], [])
+
+      query.merge!({
+          sort: [
+            {id: "asc"}
+          ],
+          size: size,
+          search_after: [after]
+      })
+
+      results =  Business.search(query)
+      results_amount = results.size
+      businesses += results
+      after += size
+
+      break if results_amount == 0
+    end
+
+    businesses
+  end
+
   def business_export_params
     params.permit(:sort, :q)
   end
--- a/app/helpers/businesses_helper.rb
+++ b/app/helpers/businesses_helper.rb
@@ -14,22 +14,8 @@ module BusinessesHelper
       .records
   end
 
-  def search_for_businesses_in_batches(size = 1000)
-    businesses = []
-    after = 0
-
-    loop do
-      query = search_after_query(search_query.build_query([], [], []), size, after)
-
-      results = Business.search(query)
-      results_amount = results.size
-      businesses += results
-      after += size
-
-      break if results_amount.zero?
-    end
-
-    businesses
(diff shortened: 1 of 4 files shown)


app/helpers/businesses_helper.rb
```diff
@@ -14,6 +14,24 @@ def search_for_businesses(page_size = Business.count)
       .records
   end
 
+  def search_for_businesses_in_batches(size = 1000)
+    businesses = []
+    after = 0
+
+    loop do
+      query = search_after_query(search_query.build_query([], [], []), size, after)
```
slorek, line 22: I think this line at least should be moved to the [Searchable concern](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/app/models/concerns/searchable.rb) and this would ensure we respect the model's overridden highlight, fuzzy and exact fields config (see https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/app/models/concerns/searchable.rb#L64 for example).

I think we can probably make this looping the default behaviour for `full_search` there unless you disagree. That would mean that all models would benefit with one code change.

app/helpers/businesses_helper.rb
```diff
@@ -80,4 +98,14 @@ def build_breadcrumb_structure
       ]
     }
   end
+
+  def search_after_query(_search_query, size, after)
+    query.merge!({
+      sort: [
+        { id: "asc" }
```
slorek, line 105: I think this page has sort params which would now not work, so again the [Searchable concern](https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/app/models/concerns/searchable.rb) would probably give us some pointers to how to implement this.

app/helpers/businesses_helper.rb
```diff
@@ -80,4 +98,14 @@ def build_breadcrumb_structure
       ]
     }
   end
+
+  def search_after_query(_search_query, size, after)
+    query.merge!({
```
slorek, line 103: I'm not sure `query` is defined here...

app/models/concerns/searchable.rb
```diff
@@ -64,6 +64,32 @@ def self.full_search(query)
...
+      records = []
+      after = 0
+
+      loop do
+        query = search_query.build_query(highlighted_fields, fuzzy_fields, exact_fields)
+
+        query.merge!({
+          sort: [
+            { id: "asc" }
+          ],
```
slorek, line 77: I think `query` already includes `sort` params (which in this case is set in `BusinessesHelper#sorting_params` and also set in the other controller helpers). I'm not sure we need to override this here and should probably remove it as it could be a change in behaviour.

```suggestion
```

app/models/concerns/searchable.rb
```diff
@@ -64,6 +64,32 @@ def self.full_search(query)
       __elasticsearch__.search(query.build_query(highlighted_fields, fuzzy_fields, exact_fields))
     end
 
+    def self.search_for_businesses_in_batches(search_query, size = 1000)
```
slorek, line 67: This should work for all models. Also suggest setting the default for `size` as a constant in the class so it could be overridden on a per-class basis (depending on the index size for each model).

```suggestion
    def self.search_in_batches(search_query, size = 1000)
```

PR #1628, opened 2021-10-18
Improve export performance and error reporting

slorek:
An iteration to improve the export functionality:

* Move the `ProductExportWorker` job class from `models` to `jobs`.
* Ensure an exception is raised if the file attachment fails for any reason (the standard `attach` method only calls `save` under the hood). This will cause the job to be retried and the error reported rather than failing silently.
* Avoid passing `Product` objects to the `ProductExportJob`. This improves performance and avoids hitting the database statement time limit, as well as improving the replay-ability and easing diagnosis of the export job.
* Retrieve products in chunks to avoid the PaaS database statement time limit, and avoid repeated iterations of products.
* Ensure exceptions encountered in the export jobs are reported to Sentry.
* Drastically improve the run time (by several minutes) of the export specs by not generating fresh export files for each attribute tested.
* Avoid writing temporary files for the export (which are sometimes not cleaned up) - note this may or may not have a worker memory impact so we'll need to monitor this
* Refactor the case and product export models to improve readability and testability and ensure exceptions bubble up to the job.

I propose that the next step to improve performance should be to defer the searching to the background job rather than in the request cycle. This will avoid an unnecessary delay in feedback to the user and improve the replay-ability of the export jobs.


nasirkhanbeis (approved):
Tested on review app after injecting more data, verified full export and search result too. looks good as expected.

PR #1640, opened 2021-10-28
Remove user

KyleMacPherson:
https://trello.com/c/mPNfQOt4/1174-add-remove-user-feature-to-team-admins

## Description
Allows team admin to remove users from their team (and PSD as a whole). Prior to this PR the user would need to contact the team to have a user removed.

Note: This PR does not include any work to allow a deleted user to be re-added, this will come in a follow up PR.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis (approved):
Tested this on the review app, looks all good as expected. 
Verified admin and non-admin users. 
Verified case re-assigned to the team when the user is removed. Activity log is reflected accordingly. 
Re-add the same user will be implemented in a separate ticket.

stdstaff (approved):

PR #1645, opened 2021-11-03
Disable inactive users

slorek:
https://trello.com/c/EyLirH8d/1159-remove-users-who-have-been-inactive-for-a-certain-period-of-time

This change specifically locks users who have signed in to the service at least once but have been inactive for a configurable period of time.

The thresholds have been set to 3 months of activity, and the job will run at 1am daily. No email will be sent until the user attempts to log in.

We could also consider deleting accounts which have remained locked for a period of time.

The copy of the email and message requires checking for suitability prior to deployment.

Additionally:

* Finally remove the `User.current` hell which is now redundant, and the request_store gem that enabled it.
* Added a missing feature spec to ensure users who are deleted are signed out.


app/models/user.rb
```diff
@@ -173,6 +192,7 @@ def has_filled_out_account_setup_form_and_verified_number?
     name.present? && mobile_number_verified?
   end
 
+<<<<<<< HEAD
```
KyleMacPherson, line 195: ✂️

KyleMacPherson (approved):

stdstaff (approved):

PR #1649, opened 2021-11-08
Change cases search sort options to a dropdown

stdstaff:
https://trello.com/c/AR5zgCsB/1170-update-sort-by-on-cases-page3

## Description
Update the Cases page to a newer visual design. This moves "sort by" to a dropdown on the right hand side, and changes the sort logic.

- On the initial page load of the cases page, select 'blank', but this implies "recent"
- Never show the "relevant" option
- Maintain state (keep whatever sort the user has perviously selected when applying new filters)
- If the user has 'blank' set and there's been a keyword search, the sort should still display the 'blank' option, but this should switch to mean "relevant"

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis (approved):
Tested this on the review app, Sort by dropdown works as expected. 
Also, verified the non-js version, combination with other filters and keyword search, 
sort option is retained in the session which I think is good.

spec/models/search_params_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  context "when sort_by is blank" do
+    let(:search_params) { { sort_by: SearchParams::BLANK } }
+
+    describe "sort_by_option" do
+      it "returns recent by default" do
+        expect(model_instance.sort_by_option).to eq(SearchParams::RECENT)
+      end
+
+      it "returns relevant if there is a query present" do
```
KyleMacPherson, line 14: [Minor] I would add another context for clarity here

```context "when sort_by is blank" do
        context "when there is a query present" do
        end

        context "when there is no search query present" do
        end
end```

KyleMacPherson (approved):

slorek (changes requested):
I think we are missing a scenario in the feature spec describing the sorting behaviour when keywords are entered.

I'd also like to review the behaviour with Darren and Toby before this is introduced as I feel the changes may be confusing.

spec/models/search_params_spec.rb
stdstaff: :+1: https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/1649/commits/5ddb635b9c966b7eec9dbf146478841ef217c679

stdstaff:
This logic has been reviewed internally and it is felt further investigation is required into how best to present sort options to service users, and into how to implement this technically. Marking this PR "DO NOT MERGE" pending the results of these discussions.

KyleMacPherson (approved):

slorek (approved):

nasirkhanbeis (approved):
Tested this on the review app, looking good as expected. Nice work Nick!

PR #1679, opened 2021-12-07
Case filters changes

KyleMacPherson:
https://trello.com/c/xtXEcwak/1229-revamp-of-the-filters-on-the-cases-page

## Description
The key change in this PR is the switch of the filters from checkboxes in which multiple options can be chosen from each filter group, to radio buttons in which only one option can be chosen from each group.

I have also refactored some key files, such as `investigation_search_helper.rb` in order to try to make the code clearer and more understandable.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


stdstaff (approved):

slorek (changes requested):
Couple of issues spotted during manual testing on the review app:

* Disabling JavaScript doesn't expand the 'more options' in the filters column
* The download link is still underneath the search results, the ticket says this should be moved to the left column

slorek (approved):

PR #1682, opened 2021-12-09
Design revisions for the products page

stdstaff:
https://trello.com/c/lzwzowhr/1206-products-page-html-design-revision-and-refresh-8

## Description

- Add expandable details help text
- Make search bar more prominent and move above table
 - Should be clear-able with 'x' (see prototype behaviour)
- Add search results statement above table (see case search page implementation for scenarios)
- Add `Filters` header above `Hazard type`
- Change cards to a HTML data table
- Add table headers (Name, Type, Category)
- Add footers if there are 12 or more results (Name, Type, Category)
- Update button text to `Apply`
- Add `Reset` link in-line with button which clears fields
- Table Footer included when there are 12 or more products
- Pagination included on when there are more than 20 products/results
- Results overview statement is contextual. Different statement for page load, keywords, and filters

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis (approved):
The new page design looks good as expected. Just a minor one, saw no footer when I have exactly 12 rows. however, for 12+ more rows looks good.

app/views/products/_search_statement.html.erb
```diff
@@ -0,0 +1,13 @@
+<p class="govuk-body">
```
KyleMacPherson, line 1: [Minor] Might be good to add a test for the search statement

KyleMacPherson (approved):

PR #1707, opened 2021-12-29
Fix smoke test bug

KyleMacPherson:
https://trello.com/c/7J2xuOkC/1249-spike-investigate-smoke-test-failures

## Description
Smoke tests were failing due to not being able to find "Enter security code". This seemed to be because in our code we were breaking a loop based on the contents of a page which may not have loaded yet, so I have changed it to break the loop based on the path and changed the location of the break clause which fixes the issue.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -100,21 +100,21 @@ jobs:
         bundle config path vendor/bundle
         bundle check || bundle install --without development --jobs=4 --retry=3
 
-    # Temporarily disabled due to unreliability of login flow steps
-    #
-    # - name: Run Smoke Tests
-    #   if: success()
-    #   env:
-    #     SMOKE_USER: ${{secrets.SMOKE_USER}}
-    #     SMOKE_PASSWORD: ${{secrets.SMOKE_PASSWORD}}
-    #     SMOKE_RELAY_CODE_URL: ${{secrets.SMOKE_RELAY_CODE_URL}}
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -100,21 +100,21 @@ jobs:
...
-    #     SMOKE_PASSWORD: ${{secrets.SMOKE_PASSWORD}}
-    #     SMOKE_RELAY_CODE_URL: ${{secrets.SMOKE_RELAY_CODE_URL}}
-    #     SMOKE_RELAY_CODE_USER: ${{secrets.SMOKE_RELAY_CODE_USER}}
-    #     SMOKE_RELAY_CODE_PASS: ${{secrets.SMOKE_RELAY_CODE_PASS}}
-    #     IS_REVIEW_APP: "false"
-    #     SMOKE_TEST_URL: "https://staging.product-safety-database.service.gov.uk"
-
-    #   run: |
-    #     bundle exec rspec ./smoke_test/cases_page_spec.rb
+    Temporarily disabled due to unreliability of login flow steps
```
slorek, line 103: ✂️

slorek (approved):

.github/workflows/deploy.yml
KyleMacPherson: Damn! Thanks @slorek, good spot.

stdstaff (approved):

PR #1708, opened 2021-12-29
Switch from Elasticsearch to Opensearch for search infrastructure

slorek:
https://trello.com/c/gY6GmBXo/1245-spike-to-investigate-migrating-elasticsearch-instances-to-opensearch-in-all-environments

## Description
Due to a licensing change, GOV.UK PaaS are [retiring Elasticsearch](https://docs.cloud.service.gov.uk/deploying_services/opensearch/#migrating-from-elasticsearch) in favour of the open source Opensearch alternative.

This pull request implements the necessary code changes to support the infrastructure change. It _does not_ replace the usage of the `Elasticseearch` gems, which remain pinned to the old versions. This is because replacing these will require extensive code changes due to our reliance on the [elasticsearch-rails](https://github.com/elastic/elasticsearch-rails) gem which uses ActiveRecord callbacks to keep the index up to date. We should be managing these updates in our new service classes, but it's likely that not all cases where updates happen will be covered yet and we'll need to add a lot of tests to cover this, so it's better to manage this as a separate piece of work.

## Development Notes
You need to change your environment variable from `ELASTICSEARCH_URL` to `OPENSEARCH_URL`.

If running Opensearch via the [Docker Hub image](https://hub.docker.com/r/opensearchproject/opensearch), note that this is configured to use TLS and HTTP basic auth by default. This caused problems in the GitHub Actions environment with the Rspec suite hanging. Disabling the Opensearch security plugin using the undocumented `DISABLE_SECURITY_PLUGIN` environment variable fixed this problem. It is recommended to disable this plugin as per the `docker-compose.yml` file to maintain consistency with the GitHub environment.

## Deployment Notes

This deployment will need to be performed _out of hours_ as it will result in at least 30 minutes where no search results will appear while the indexes are being rebuilt.

Before deployment, provision the new service instance on both `staging` and `prod` spaces:

```
cf create-service opensearch small-ha-1 psd-opensearch-1
```

Then following deployment run the indexing task on both spaces:

```
cf run-task psd-web --command "export \$(./env/get-env-from-vcap.sh) && rake opensearch:index" -k 2G
```

Ensure the task status is shown as `SUCCEEDED` and test the service manually.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


KyleMacPherson (approved):
Tested locally and all looks to be working well

nasirkhanbeis (approved):
Tested OpenSearch on the review app, looking good as expected. 
Verified all possible search scenarios and did full regression testing.
Need to verify shared  OpenSearch instances on more than one review app to ensure search results are not broken

stdstaff (approved):

PR #1709, opened 2021-12-30
Fix exports bug caused by linking to blob path rather than proxy path

KyleMacPherson:
https://trello.com/c/PlVQdcWP/1261-investigate-failing-business-product-exports

## Description
Prior to this change user's could not download their business or product export spreadsheets. They could generate the spreadsheet, receive an email with a link to the download page but the download was not working, it was instead redirecting to the home page.

This issue is due to us using a blob path rather than a proxy path, and was caused when work was done to authorise attachments by overwriting activestorage controllers. By changing to a proxy path for these download links this issue is resolved in this PR.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


slorek (approved):

nasirkhanbeis (approved):

PR #1724, opened 2022-01-10
Move failed log in logging to trigger when authentication fails rathe…

KyleMacPherson:
…r than when form has errors

## Description

Prior to this change, the logging for failed sign in error was triggered when the sign in form had errors, however after seeing the logging go live it became clear that this logging should be triggered when authentication fails. This PR corrects this misplaced logging.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -40,7 +40,7 @@ module Users
           return render :new
         end
 
-        Rails.logger.info "Forgotten password email sent to user: #{user.id}"
+        Rails.logger.info "Forgotten password email sent to user: #{user.try(:id)}"
       end
     end
 
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -39,6 +39,8 @@ module Users
           resource.errors.merge!(reset_password_form.errors)
           return render :new
         end
+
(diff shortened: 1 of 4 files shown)


stdstaff (approved):

app/controllers/users/sessions_controller.rb
```diff
@@ -47,6 +47,8 @@ def handle_authentication_success
     end
 
     def handle_authentication_failure(user)
+      Rails.logger.info "Failed sign in attempt for email address: #{user.try(:email)}"
```
slawosz, line 50: You should not use data that can identify user in logs (or any other place).

app/controllers/users/passwords_controller.rb
```diff
@@ -39,6 +39,8 @@ def create
           resource.errors.merge!(reset_password_form.errors)
           return render :new
         end
+
+        Rails.logger.info "Forgotten password email sent to user: #{user.try(:id)}"
```
slawosz, line 43: I would rather do something like: "[forgotten_password] user_id: #{user.try(:id)}" its easier to search for. You could also add other tag, like "[user_event][forgotten_password]..." so by serching for user_event it would be easy to find all types of events.

slawosz (approved):

PR #1726, opened 2022-01-11
Keep the 'more options' filters open when in use

stdstaff:
https://trello.com/c/B6uSZK1k/1255-more-options-filter-drop-down-behaviour

## Description

Currently if a user uses any of the filters from the expanded "More options" section on the Cases page, they collapse again when the filters are applied. This change keeps the "More options" section expanded if any of the filters inside have been changed from their default values. If they have all been changed back to their default values, it will collapse again.

I have rearranged the order of the feature spec to make it consistent with the layout of the page, and added expectations for the new behaviour.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


KyleMacPherson (approved):

slorek (approved):

nasirkhanbeis (approved):

PR #1738, opened 2022-01-17
Introduce a unique 'PSD ref' Product reference

stdstaff:
https://trello.com/c/K4K4b5sP/1273-introduce-a-unique-product-id-for-products-in-the-psd-database

## Description
Introduces a unique "PSD ref" for each product, formed of "psd-" and its internal ID. Surfaces this informaton in the table on the Products page. This change will help users identify individual products in cases where other data such as barcode is not unique. This does occur for example when a counterfeit product has affixed a genuine product's barcode.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


spec/models/product_spec.rb
```diff
@@ -1,7 +1,15 @@
 require "rails_helper"
 
-RSpec.describe Product do
+RSpec.describe Product, :with_opensearch do
```
slorek, line 3: This will significantly slow down the specs so would suggest removal and only adding to those examples where we are testing search functionality.

```suggestion
RSpec.describe Product do
```

slorek (approved):

spec/models/product_spec.rb
stdstaff: I've added this because the instance needs to be created in order to have an `id`, and creation indexes the model with opensearch. Is there an alternative approach?

stdstaff: This appears to disrupt `it_behaves_like "a batched search model"` in the example above unfortunately, this shared example aready includes `:with_opensearch`...

```
1) Product behaves like a batched search model #search_in_batches when searching in batches returns the expected objects
   Failure/Error: described_class.import refresh: :wait_for

   NoMethodError:
     private method `select' called for nil:NilClass
   Shared Example Group: "a batched search model" called from ./spec/models/product_spec.rb:4
```

slorek: Ah OK, I didn't realise `with_opensearch` was already included there and there weren't any other examples in this spec.

In that case, no significant performance deficit introduced, but ideally for a slightly faster run and to avoid slowing down any future examples we add to this spec, I suggest we could refactor like so:

```ruby
  describe "#psd_ref" do
    let(:id) { 123 }
    let(:product) { build :product, id: id }

    it "returns a reference formed with 'psd-' and the product's ID" do
      expect(product.psd_ref).to eq("psd-#{id}")
    end
  end
```

KyleMacPherson (approved):

spec/models/product_spec.rb
stdstaff: Perfect, I'll switch to building but not saving 👍

PR #1745, opened 2022-01-18
Strip whitespace from business search queries

KyleMacPherson:
https://trello.com/c/7bPQx24l/1244-business-page-search-bug

## Description
Prior to this change, when a user entered a search query with leading or trailing whitespace it would not be stripped and would impact the results of the search. This was evident particularly when user's searched by company number, in which case the search would not work as expected if it included leading or trailing whitespace.

This change strips leading and trailing whitespace from all business search queries.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


slorek (approved):

stdstaff (approved):

nasirkhanbeis (approved):

PR #1748, opened 2022-01-18
Display matching team name on case result highlights

slorek:
https://trello.com/c/V9VoZQbC/1167-case-search-result-page-returning-team-ids-instead-of-names

## Description
Displays the matching team name in case search result highlights rather than their ID.

Note Opensearch is currently configured to wrap highlights in its result set in `<em>` tags. This can be changed, but would involve a much larger refactoring, so I elected to strip them out in order to resolve the team name. An exception will be thrown if the team cannot be resolved from the ID. This should be safe since we do not delete teams for various reasons which would break things elsewhere.

I nalso changed the label 'teams with access' to 'team added to the case' to be consistent with the label in the filter list.

### DIFF
--- a/app/helpers/investigations/display_text_helper.rb
+++ b/app/helpers/investigations/display_text_helper.rb
@@ -61,7 +61,7 @@ module Investigations::DisplayTextHelper
 
     highlight[1].each do |result|
       unless should_be_hidden?(source, investigation)
-        best_highlight[:content] = get_highlight_content(result)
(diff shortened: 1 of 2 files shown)


stdstaff (approved):

KyleMacPherson (approved):

nasirkhanbeis (approved):
Tested this on the review app, looking good as expected!

PR #1768, opened 2022-02-01
Business pages bug when more than 10k records

KyleMacPherson:
https://trello.com/c/KXsl3e8Q/1295-bug-on-prod-total-businesses-count-on-the-businesses-page-is-incorrect

## Description
This change:
- corrects the count on business/product/cases page when no filters/search queries are used.
- directs the user to the final page of records if they try to access page >500.

Note: I have not added feature tests for scenarios with >10,000 records as in my opinion that these scenarios will slow down our test suite without providing value necessary to do so but I am definitely open to if the team prefers.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis (approved):
Tested businesses count over 10k which is showing correctly. 
Also, verified cases/businesses/products download looking good.

slorek (commented):
Thanks - I think we should add a _request_ spec for testing the overriding of the page number where the request is for page numbers beyond page 500. This will give us an opportunity to document the reason for this override in the test as well.

Like you say a feature test creating 10,000 records in Elasticsearch will probably be overkill and very slow, but I wonder if there is another way 🤔  Maybe by stubbing `Business.count` and testing that is the value which is displayed?

KyleMacPherson:
> Thanks - I think we should add a _request_ spec for testing the overriding of the page number where the request is for page numbers beyond page 500. This will give us an opportunity to document the reason for this override in the test as well.
> 
> Like you say a feature test creating 10,000 records in Elasticsearch will probably be overkill and very slow, but I wonder if there is another way 🤔 Maybe by stubbing `Business.count` and testing that is the value which is displayed?

I have added feature tests for both of these scenarios. I did try to add request specs but in doing so I needed to stop stubbing opensearch which caused other issues.

slorek (approved):

PR #1784, opened 2022-02-10
Stop filter params from being saved in session

KyleMacPherson:
https://trello.com/c/tMTrGT7l/1308-drop-the-filter-session-storage

## Description
Prior to this change the last used filters on the cases page were stored in the session and were used again the next time the user visited the cases page. This change removes this behaviour, meaning that each time the user visits the cases page the default filters are applied.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


app/helpers/investigation_search_helper.rb
```diff
@@ -6,14 +6,6 @@ def search_query(user)
     OpensearchQuery.new(@search.q, filter_params(user), @search.sorting_params, nested: nested_filters(user))
   end
 
-  def set_search_params
-    params_to_save = params.dup
-    params_to_save.delete(:sort_by) if params[:sort_by] == SortByHelper::SORT_BY_RELEVANT
-    @search = SearchParams.new(query_params)
-
-    store_previous_search_params
-  end
```
KyleMacPherson, line 15: It doesn't look like this method does anything different to `set_search_params` in the search_helper controller other than `store_previous_search_params` and as such is no longer needed but I might be mistaken.

slorek (approved):

PR #1799, opened 2022-02-17
Remove IE8 support and separate PostCSS config

stdstaff:
https://trello.com/c/mPwQM8U6/1297-remove-ie8-support-from-the-project

## Description

- IE8 is End of Life and we do not currently test with or support it
- The minimum IE version the service actively supports is IE11, in line with [the service manual](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in-from-june-2021)
- Removing the various workarounds and custom stylesheets which were used to support IE8

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis (approved):
Tested on the review app(regression test), looking good as expected.

PR #1847, opened 2022-04-06
Redact data on restricted cases export unless user has access to it

KyleMacPherson:
https://trello.com/c/JRrGkReH/1362-spike-investigate-how-to-implement-alldataexporter-role-restrictions-based-on-team

## Description
Redact information from the case export if the case is restricted and the current user is not on the team that owns the case.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/models/case_export.rb
+++ b/app/models/case_export.rb
@@ -126,7 +126,7 @@ private
   end
 
   def serialize_case(investigation, team)
-    restrict_info = !current_user_is_on_owner_team?(team, investigation) && investigation.is_private?
+    restrict_info = !current_user_is_on_owner_team?(team, investigation) && investigation.is_private
 
     restrict_info ? restricted_data(investigation) : non_restricted_data(investigation)
   end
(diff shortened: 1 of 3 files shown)


app/models/case_export.rb
```diff
@@ -143,4 +147,43 @@ def serialize_case(investigation)
...
+      investigation.date_closed,
+      investigation.risk_validated_at,
+      investigation.creator_user&.team&.name,
+      country_from_code(investigation.notifying_country, Country.notifying_countries),
+      investigation.reported_reason
+    ]
+  end
+
+  def restrict_info?(team, investigation)
+    !current_user_is_on_owner_team?(team, investigation) && investigation.is_private
```
stdstaff, line 183: Could this be a policy?

```rb
Pundit.policy!(user, investigation).view_restricted?
```

stdstaff (approved):

PR #1853, opened 2022-04-14
Prevent double-click/back when removing a user

stdstaff:
https://trello.com/c/x0BPJvqm/1240-prevent-double-click-on-remove-user-journey-and-backlink-to-return-back-to-the-team-page

## Description

When an admin removes a user from their team, if they clicked "save and continue" twice, or if they used the browser's back button after removing a user and tried to remove them a second time, a server error was produced. This PR prevents double-click on the button, and displayes an alert to the admin if they try to remove a user whom has already been removed.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -58,8 +58,11 @@ class UsersController < ApplicationController
     return render(:remove) if @remove_user_form.invalid?
 
     if @remove_user_form.remove == "yes"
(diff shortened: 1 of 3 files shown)


KyleMacPherson (approved):

nasirkhanbeis (approved):
Tested this on the review app, looking good as expected!

PR #1884, opened 2022-05-16
Delete unsafe files job

KyleMacPherson:
https://trello.com/c/5wZznipW/1393-delete-unsafe-attachments-after-validation-has-completed

## Description
Prior to this change, unsafe files in our system were not deleted. We have validation on file upload, however due to this running in a background job if the antivirus check is not run in time sometimes the antivirus validation will be skipped.

This change introduces a cron job that deletes any offending files every 15 minutes and informs the user via email.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/jobs/delete_unsafe_files_job.rb
+++ b/app/jobs/delete_unsafe_files_job.rb
@@ -1,5 +1,5 @@
 class DeleteUnsafeFilesJob < ApplicationJob
-  def self.perform
+  def perform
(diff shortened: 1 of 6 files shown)


slorek (changes requested):
Please add an index on the `active_storage_blobs` table on the `metadata` column to improve the performance of this query.

slorek (approved):

PR #1899, opened 2022-05-25
Correct heading in the cases pages table footer

stdstaff:
https://trello.com/c/W7tJiicf/1399-on-the-cases-pages-wrong-column-label-used-in-the-table-footer

## Description

Fixes a bug in the table footer for the Cases pages. On the Cases pages: 'Your cases' and 'Team cases' in the footer of the table, the second label currently reads "Case owner" when it should read "Case created".


nasirkhanbeis:

slorek (approved):

PR #1915, opened 2022-06-09
Introduce a new lefthand nav for the cases detail pages

stdstaff:
https://trello.com/c/qzsoT0EX/1405-left-nav-for-cases

## Description
This PR brings the top case navigation on the current service to the left hand side, and also provides hyperlinks to navigate the cases landing page.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


slorek (approved):
Suggest renaming the `_transitional_case_nav.html.erb` partial as invariably these things tend to stick around longer than we expect and context gets forgotten ;)

nasirkhanbeis (approved):
Tested this on the review app,
The functionality works as expected and the page design looks good as given in the transition designs.
Verified h1s and HTML titles with all case types.
Verified the action bar which is untouched, looks good.
Tested left nav links on different supported browsers and on smaller screens.

PR #1944, opened 2022-06-29
Config for using multiple sentry services across different environments

KyleMacPherson:
https://trello.com/c/ZnlYY5CO/1376-rate-limit-apps-environments-on-sentry

## Description
This change aims to ensure that our production environment uses a different sentry dsn key to our staging and int (review apps) environments. I created three services, one for each environment, and this PR adds the correct configuration to ensure that each service is bound the correct apps depending on the environment.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


manifest.yml
```diff
@@ -23,7 +23,7 @@ applications:
     - psd-pghero-env
     - psd-queue-6
     - psd-rails-env
-    - psd-sentry-env
+    - ((sentry-service-name))
```
stdstaff, line 26: I think you'd need to update `deploy.sh` to pass in an extra `--var` if you want to do this? https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/deploy.sh#L18

KyleMacPherson: You're right, I forgot to commit that file 😳 Good spot, thanks Nick. Corrected.

stdstaff (approved):

PR #1961, opened 2022-07-26
Change text/link from 'close case' to 're-open' case if the case is c…

KyleMacPherson:
https://trello.com/c/Zpy8v3OO/1487-when-a-case-is-closed-header-link-should-change-to-re-open

## Description
This change fixes a bug which causes the link at the top of the case overview page to show "Close case" and take the user to the close case page, even when the case is already closed. The expected behaviour is that the link will take the user to the re-open case page in this case.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/spec/features/change_case_status_spec.rb
+++ b/spec/features/change_case_status_spec.rb
@@ -27,7 +27,14 @@ RSpec.feature "Changing the status of a case", :with_opensearch, :with_stubbed_m
 
     # Navigate via the case overview table
     visit "/cases/#{investigation.pretty_id}"
-    first(:link, "Close").click
+    # x = all("div.govuk-grid-row").first
+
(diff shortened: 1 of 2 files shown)


nasirkhanbeis (approved):
looks good as expected!

spec/features/change_case_status_spec.rb
```diff
@@ -27,7 +27,14 @@
 
     # Navigate via the case overview table
     visit "/cases/#{investigation.pretty_id}"
-    first(:link, "Close").click
+    # x = all("div.govuk-grid-row").first
```
stdstaff, line 30:
```suggestion
```

stdstaff (approved):

PR #1952, opened 2022-07-18
Case summaries no longer auto-populate; new summary edit form design

stdstaff:
https://trello.com/c/RYMK5q7Q/1473-cases-page-edit-the-summary

## Description

- Case summaries now no longer auto-populate with product risk details (they are now blank by default)
- The summary edit form is now visually larger and has a new character count

## Checklist:
- [ ] Have you documented your changes in the pull request description?
- [ ] Does the change present any security considerations?
- [ ] Is any gem functionality overloaded? Eg: Devise controller methods being overloaded.
- [ ] Has acceptance criteria been tested by a peer?
- [ ] Has the CHANGELOG been updated? (If change is worth telling users about.)

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

### Accessibility testing (author)
- [ ] Works keyboard only
- [ ] Tested with one screen reader
- [ ] Zoom page to 400% - content still visible
- [ ] Disable CSS - does content make sense and appear in a logical order?

PR #1971, opened 2022-08-02
Allow blank case summaries; revert to longer summaries

stdstaff:
#1952 introduced a situation where a case can be generated with no summary (description), but the user not be given an opportunity to supply one, which would then fail validation.

Since the intention is to allow a summary to be blank, and for one to not be automatically generated, this PR removes the `presence` validation.

The PR also reverts the model change from a 10,000 character limit to an 800 character limit. We need an effective strategy to manage existing cases which have summaries over 800 characters before we can impose a shorter length. For now, the user is prompted visually to enter no more than 800 characters, but the backend validation will permit them to go up to 10,000 untill we have a solution for this case.


KyleMacPherson (approved):

nasirkhanbeis (approved):
looks all good as expected!

PR #1984, opened 2022-08-09
Remove the character count and lower limit on case summaries

stdstaff:
https://trello.com/c/f5lVcGj8/1507-remove-character-count-for-the-summary-field

## Description

#1952 and #1971 made some changes to case summaries, reducing their length. We have decided to revert these changes pending further research into how summaries are being used by our users. This PR restores the summary character limit to 10,000 characters, but removes the new character count design as this was determined to be a distraction.


slorek (approved):

nasirkhanbeis (approved):
Tested this on the review app, looks good as expected!

PR #1994, opened 2022-08-15
Update a handful of back links in the case area

stdstaff:
https://trello.com/c/PWmQLO6s/1445-back-links-in-a-case-area-require-updating

## Description

- The back link when adding a test result will return to the case summary page
- The back link when when a product needs to be added to a case before adding an accident or incident will return to the case summary page
- The back link when when a product needs to be added to a case before adding a risk assessment will return to the case summary page
- The back link when editing a phone call will return to the phone call's details


slorek (approved):

nasirkhanbeis (approved):
Tested this on the review app, looks good as expected. Nice work Nick 👍

PR #1999, opened 2022-08-17
Export metrics for open/closed case totals

stdstaff:
https://trello.com/c/hycOmu2q/1498-add-number-of-open-and-number-of-closed-cases-to-kibana-dashboard

## Description

Adds `total_number_of_open_cases` and `total_number_of_closed_cases` metrics to the regular stats logging job. These metrics can then be used on the service stats dashboard.


app/jobs/log_db_metrics_job.rb
```diff
@@ -2,6 +2,8 @@ class LogDbMetricsJob < ApplicationJob
   def perform
     stats = {
       total_number_of_cases: Investigation.count,
+      total_number_of_open_cases: Investigation.where.not(is_closed: true).count,
```
stdstaff, line 5: Note: I chose `.not` over `is_closed: false` because whilst `is_closed` has a default of `false`, it doesn't have a not-null constraint, so technically it could be `false` or `null` to signify openness.

I also added both cases (rather than subtracting `closed` from `total` to get open, as it's not always possible to do mathematical operations against metric series, so having all three metrics is more flexible.

KyleMacPherson (approved):

PR #2001, opened 2022-08-17
New ts flow

KyleMacPherson:
https://trello.com/c/AQigpt6X/1490-new-trading-standards-officer-tso-pages-for-creating-a-case

## Description
This is the first change to go into our new feature branch which will contain all the new case flow work.

This PR changes the way that TS users create cases. It removes any business or product information from the case creation process and shows a simplified wizard which the user can work through to create a case. More details can be found in the above trello card.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis (approved):
Tested this new ts flow, looks all good as expected.
Suggestion:  we need to introduce case name field for OPSS create allegation journey too before we release this to prod

PR #2012, opened 2022-08-23
Add the concept of products being owned by a team

stdstaff:
https://trello.com/c/9FcZdj4F/1526-creating-a-product-record-in-the-database-rules

## Description

- Adds a database column, attribute, relationships and foreign key to represent the Team which owns a Product
- Products have no owning team by default

```rb
Product.owning_team
# => <Team ...>
Team.owned_products
# => [...]
```

### DIFF
--- a/db/migrate/20220822125733_add_owning_team_to_products.rb
+++ b/db/migrate/20220822125733_add_owning_team_to_products.rb
@@ -1,0 +1,6 @@
+class AddOwningTeamToProducts < ActiveRecord::Migration[6.1]
+  def change
+    add_reference :products, :owning_team, type: :uuid, null: true, index: false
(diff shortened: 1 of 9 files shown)


slorek (approved):

db/migrate/20220822125733_add_owning_team_to_products.rb
```diff
@@ -0,0 +1,6 @@
+class AddOwningTeamToProducts < ActiveRecord::Migration[6.1]
+  def change
+    add_reference :products, :owning_team, type: :uuid, null: true, index: false
+    add_foreign_key :products, :teams, column: :owning_team_id, validate: false
```
slorek, line 4: We should add an index on the `owning_team_id` column.

PR #2015, opened 2022-08-25
Remove older CSV render export in favour of XLSX export

stdstaff:
Autenticated users with the `exporter` role were able to generate two kinds of exports: A CSV export by appending `.csv` to product and businesses index URLs, and an XLSX export using the generated export feature.

The former was no longer linked anywhere in the service and was not being used. This PR removes it in favour of the more complete XLSX export which is linked on those pages.


slorek (approved):

nasirkhanbeis (approved):

PR #2020, opened 2022-08-26
Implement product ownership assignment on edit and link

stdstaff:
https://trello.com/c/sAyPjgIu/1542-editing-a-product-record-database-rules

Implements the following rules around product ownership/creation:

- If the team owns the product record and edits the product, there is no change to the product record ownership
- If a product record has no owner, the team which edits it becomes the product record owner
- If the team does not own the product record (as it is already owned), they can not edit it
- If a product record has no owner, when it is linked to a case, the team which owns the case becomes the owner of the product


app/services/update_product.rb
```diff
@@ -1,13 +1,18 @@
...
 
-  delegate :product, :product_params, to: :context
+  delegate :product, :product_params, :updating_team, to: :context
 
   def call
     context.fail!(error: "No product supplied") unless product.is_a?(Product)
     context.fail!(error: "No product params supplied") unless product_params.is_a?(Hash)
+    context.fail!(error: "No updating team supplied") unless updating_team.is_a?(Team)
+    context.fail!(error: "You can not explicitly set a Product's owning team during update") if product_params.key?("owning_team") || product_params.key?("owning_team_id")
+    context.fail!(error: "The updating team does not own the product") unless product.owning_team.nil? || product.owning_team == updating_team
```
slorek, line 11: 🤔 Suggestion: I understand the intent, but maybe enforcing this rule is beyond the remit of the service. It's a duplication of the policy which should be the thing which determines permissions. Sometimes there is a need to use services for utility tasks outside the bounds of user interactions. I'd be inclined to enforce this only in the controller. At least, we could re-use the `ProductPolicy` implementation. What do you think?

stdstaff: Discussed and agreed that the service will not enforce product ownership, and that updates will be permitted through the service by any team.

slorek (approved):

PR #2046, opened 2022-09-21
Investigation product data migration

KyleMacPherson:
https://trello.com/c/I3imrFNG/1520-data-migration-of-4-pieces-of-case-specific-info-from-product-to-case

## Description
Migrate investigation product data from the products table to the investigation products table and remove references to these removed columns

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -12,13 +12,6 @@ class Product < ApplicationRecord
     "unsure" => "unsure"
   }
 
-  enum affected_units_status: {
-    "exact" => "exact",
-    "approx" => "approx",
-    "unknown" => "unknown",
-    "not_relevant" => "not_relevant"
-  }
-
   enum has_markings: {
     "markings_yes" => "markings_yes",
     "markings_no" => "markings_no",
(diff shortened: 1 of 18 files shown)


app/models/product.rb
```diff
@@ -4,19 +4,14 @@ class Product < ApplicationRecord
...
     "genuine" => "genuine",
     "unsure" => "unsure"
   }
 
-  enum affected_units_status: {
-    "exact" => "exact",
-    "approx" => "approx",
-    "unknown" => "unknown",
-    "not_relevant" => "not_relevant"
-  }
```
stdstaff, line 18: We probably need a new model class, `InvestigationProduct` that has these enums, and has the `redacted_export_with` declaration inside, so these fields continue to be exported from their new table.

stdstaff (approved):

lib/tasks/rapex_import.rake
```diff
@@ -55,7 +55,6 @@ module DataImportHelper
       name: "#{brand(notification)}, #{name}, #{field_from_notification(notification, 'type_numberOfModel')}",
       description: field_from_notification(notification, "description"),
       category: field_from_notification(notification, "category"),
-      batch_number: field_from_notification(notification, "batchNumber_barcode"),
```
slorek, line 58: I think we should add this to the `create_investigation_product` method in this task.

slorek (changes requested):
I think we need to handle the fields differently in the `AddProductToCase` service and the products in the seeds file.

nasirkhanbeis (approved):
Had a quick look at the review app, regression tested cases and product creation. looks all good.

slorek (approved):

PR #2051, opened 2022-09-23
Remove redundant Rapex data import

slorek:
https://trello.com/c/cO5mZtMb/1604-remove-redundant-rapex-import-code

This removes the redundant Rapex import task which has never been used in production.

Doing so renders the `Source`, `ReportSource` and `UserSource` models redundant. Instead I have stored the `added_by_user_id` against the relevant models (naming is consistent with that used elsewhere in the application).

Also fixed some very minor bugs in certain services where the notification emails included names with the team names appended where this was not necessary or consistent.


stdstaff (approved):

stdstaff (approved):

nasirkhanbeis (approved):
Sanity tested on the review app looks all good as expected. (verified scenarios where services have changed and all key scenarios)

PR #2052, opened 2022-09-26
Case specific info

KyleMacPherson:
https://trello.com/c/CFxQxHOo/1516-case-landing-page-case-specific-info-cap106

## Description
Show the case specific information on the investigation page.
- empty fields with an explanation text if no products are added.
- products in order from oldest to newest if products are added 

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/helpers/investigations_helper.rb
+++ b/app/helpers/investigations_helper.rb
@@ -113,6 +113,8 @@ module InvestigationsHelper
   end
 
   def units_affected(investigation_product)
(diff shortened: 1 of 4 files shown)


nasirkhanbeis (approved):
Tested this on the review app from the functionality perspective,  all good as expected.
Please note, edit links are not implemented this pr as it makes sense to implement them in the edit story card, so permissions etc will be tested in the same.

PR #2076, opened 2022-10-07
Link an existing product to a case

stdstaff:
https://trello.com/c/JwiOlXZk/1544-add-a-product-record-to-a-case-cap016

## Description

Introduces a new flow for linking existing products to a case. Users will be prompted to enter a PSD reference number and given the opportunity to review any match before adding the found product to their case.


app/controllers/investigations/products_controller.rb
```diff
@@ -4,21 +4,39 @@ class Investigations::ProductsController < ApplicationController
...
     @breadcrumbs = build_breadcrumb_structure
   end
 
   def new
-    render plain: "TODO: Link an existing product / create a product within a case flow."
+    authorize @investigation, :update?
+    @find_product_form = FindProductForm.new
+  end
+
+  def create
```
slorek, line 18: Suggestion: I think swapping the action names between `link` and `create` may be appropriate, since this action is an intermediary page and is not doing any actual creation.

app/forms/find_product_form.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  validates :reference, numericality: { only_integer: true, greater_than: 0, message: "Enter a PSD product record reference number" }
+  validate :must_find_a_product
+  validate :must_not_find_a_product_already_linked
+
+  def product
+    product_id = reference.presence && reference.try(:to_i)
+    return nil unless product_id && product_id.positive?
+
+    Product.find_by id: product_id
```
slorek, line 20: This will avoid repeat calls to the database. Don't need to worry about tracking changes to `reference` as it's not changed after initialisation:
```suggestion
    return @product if @product.present?
    
    product_id = reference.presence && reference.try(:to_i)
    return nil unless product_id && product_id.positive?

    @product = Product.find_by id: product_id
```

app/forms/find_product_form.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+private
+
+  def tidy_reference
+    return if reference.blank?
+
+    reference.sub!(/\Apsd-/i, "")
+    reference.strip!
```
slorek, line 29:
```suggestion
    return if reference.blank?

    reference.sub!(/\Apsd-/i, "")
    reference.strip!
    
    @product = nil
```

app/forms/find_product_form.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def must_find_a_product
+    return if errors.include?(:reference)
+
+    errors.add(:reference, "An active product record matching psd-#{reference} does not exist") if product.blank?
```
slorek, line 35: Suggestion: Add a `TODO` here for when product retirement is implemented.

app/services/add_product_to_case.rb
```diff
@@ -0,0 +1,53 @@
...
+           :product,
+           to: :context
+
+  def call
+    context.fail!(error: "No investigation supplied") unless investigation.is_a?(Investigation)
+    context.fail!(error: "No user supplied") unless user.is_a?(User)
+    context.fail!(error: "No product supplied") unless product.is_a?(Product)
+
+    InvestigationProduct.transaction do
+      (context.fail!(error: "The product is already linked to the case") and return false) if investigation.products.include?(product)
```
slorek, line 16: Suggestion: Add a `TODO` here for when product retirement is implemented.

app/services/add_product_to_case.rb
```diff
@@ -0,0 +1,53 @@
...
+  def change_product_owner_if_unowned
+    product.update!(owning_team: investigation.owner_team) if product.owning_team.nil?
+  end
+
+  def create_audit_activity_for_product_added
+    AuditActivity::Product::Add.create!(
+      added_by_user: user,
+      investigation:,
+      title: product.name,
+      product:
```
slorek, line 38: These two properties should be stored in `metadata` as they're deprecated columns.

app/services/add_product_to_case.rb
stdstaff: 👍 Added `TODO`

nasirkhanbeis (approved):
Tested this new feature on the review app as OPSS and TS-user, works seamless. Nice work Nick. 
I need to re-test the product id search when we implemented product retire/ soft delete so it doesn't show up in the product id search.

app/controllers/investigations/products_controller.rb
stdstaff: Makes sense, I've renamed the route that finds the potential product `find` and the route which actually creates the association `create`...

app/forms/find_product_form.rb
stdstaff: 👍 implemented.

PR #2088, opened 2022-10-18
Implement a Pundit policy for documentable models

stdstaff:
https://trello.com/c/q2zGzEMU/1605-users-shouldnt-be-able-to-add-images-to-a-product-if-they-dont-have-edit-permissions-for-the-product-cap018b

## Description

Only users who are in the team which owns a product should be able to see the “Add image” and “Add an image to the product” links on the product show page, or reach the controller actions.

- Changes the visibility of "Add an image" links so these are only visible to users who can edit the product
- Refactors the existing controller and permissions checks to use a Pundit policy class


slorek (changes requested):
Looks good, but let's add some tests for scenarios where the user does not have permission to add images to a product.

nasirkhanbeis (approved):
Note, we need to discuss product edit scope as currently collaborator with edit won't be able to edit product record.
Product record edit permissions are locked to the owning team only.

slorek (approved):

PR #2099, opened 2022-10-26
Create case from product record

KyleMacPherson:
https://trello.com/c/9hL8XPMo/1559-create-a-case-directly-from-the-product-record-cap022

## Description
This ticket forces TS users to open cases from a product. This means that a case opened by a TS user must always have a product associated with it which has caused quite a few tests to need to be changed. For this reason I have made some changes to the investigation factory, defaulting the case creator to be an OPSS user so as to not force us to create products for every investigation by default.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


nasirkhanbeis (approved):
Tested this on the review app, the functionality is looking good as expected. The design is not implemented yet which will be done in a separate ticket.

PR #2100, opened 2022-10-26
Rules surrounding removing a product from a case

stdstaff:
https://trello.com/c/1DgA1PIU/1528-removing-a-product-record-from-a-case-rules-2a

## Description

- Implements rules around product ownership when removing a product from a case
- Products can only be removed from a case if they do not relate to supporting information within the case (note, new design for this screen using `govukErrorSummary` replacing HMCTS banner)
- Attributes on the relationship between products and cases (e.g. batch numbers, units affected) are removed when the product is removed
- A removed product becomes unowned if there are no other cases owned by its owning team which link to it
- Products can no longer be removed from closed cases


app/controllers/investigations/products_controller.rb
```diff
@@ -40,13 +40,17 @@ def create
   end
 
   def remove
-    authorize @investigation, :update?
+    authorize @investigation, :remove_product?
+
+    @supporting_information = @product.supporting_information.select { |si| si.investigation == @investigation }
```
slorek, line 45: This should be quicker.
```suggestion
    @supporting_information = @product.supporting_information.where(investigation_id: @investigation.id)
```

app/services/remove_product_from_case.rb
```diff
@@ -9,7 +9,14 @@ def call
     context.fail!(error: "No investigation supplied") unless investigation.is_a?(Investigation)
     context.fail!(error: "No user supplied") unless user.is_a?(User)
 
-    investigation.products.delete(product)
+    # TODO: Versioned/historic products can't be removed from a case. Ensure
+    #   this is caught once product versioning is implemented.
+
```
slorek, line 14: FYI: `product.version` must be `nil`.

app/services/remove_product_from_case.rb
```diff
@@ -21,7 +28,17 @@ def call
...
+    # If the product was owned by the team who owns this case, and is not linked
+    # to any other cases owned by this team, it become unowned.
+    if product.owning_team == investigation.owner_team && product.investigations.none? { |inv| inv.owner_team == product.owning_team }
+      product.update! owning_team: nil
+    end
+  end
+
   def create_audit_activity_for_product_removed
+    # TODO: Refer to the correct version of the product record once versioning
+    #   is implemented.
```
slorek, line 41: FYI: Set `timestamp` on the `InvestigationProduct` record.

slorek (approved):

app/controllers/investigations/products_controller.rb
stdstaff: Unfortunately `supporting_information` returns an `Array` of differing objects rather than a relation: https://github.com/OfficeForProductSafetyAndStandards/product-safety-database/blob/master/app/models/product.rb#L73-L75 (_undefined method `where' for []:Array (NoMethodError)_) 😞

PR #2101, opened 2022-10-26
Product versioning

slorek:
https://trello.com/c/ySc2jwsU/1640-editing-a-product-versioning-of-data

Implements the [PaperTrail](https://github.com/paper-trail-gem/paper_trail) library to allow versioning of `Product` records.

This will allow us to display a snapshot of a case's product data accurate at the time the case is closed.

1. When a product is edited, the prior version of the product data is stored automatically.
1. When viewing a product from the products tab, we always display the latest (current) version.
1. it is possible to view a product's previous versions by supplying a timestamp for the desired historic state in Unix timestamp format e.g. `/products/:id/:timestamp`
1. When viewing a previous version of the product, editing of all product data is disabled.
1. When viewing a previous version of the product, the timestamp is shown in Unix timestamp format alongside the PSD reference: `psd-123_1519211809934`
1. Product search (not case search) is only be performed on current versions of products. Old versions are not searchable via the products search.
1. The product export only exports the latest version of the product. Older versions are not be exported.
1. The redacted export also exports the `versions` table.


app/models/product.rb
```diff
@@ -91,7 +93,14 @@ def name_for_sorting
...
 
-  def psd_ref
-    "psd-#{id}"
+  def psd_ref(timestamp = nil)
+    ref = "psd-#{id}"
+
+    # Timestamp to append is not necessarily the same as when the version was created.
+    if version.present? && timestamp.present?
+      ref << "_#{timestamp}"
+    end
```
KyleMacPherson, line 102: This is an interesting solution. I would expect us to use `updated_at` from this version of the product to generate a psd_ref so that the psd_ref of a product version is always the same. Any reason not to do that?

nasirkhanbeis (approved):
Tested this on the review app, looking good as expected. Nice work!

KyleMacPherson (approved):

stdstaff (approved):

PR #2114, opened 2022-11-01
Fix invalid DOM IDs on sub-nav links

stdstaff:
https://trello.com/c/IR7vDyBV/1635-html-id-bug

## Description

Removes DOM IDs from the links on the case navigation menu. Some of the generated IDs were not valid HTML as they had spaces in them, and these IDs were never actually used.


KyleMacPherson (approved):

nasirkhanbeis (approved):
Had a quick look, no impact on the left nav after the removal of DOM IDs. All good.

PR #2128, opened 2022-11-14
Add Complainant_Reference to case export

stdstaff:
https://trello.com/c/B21OZf08/1658-include-trading-standards-reference-in-the-cases-export

## Description

Adds the following existing field (Complainant Reference) to the Case Export.

image

- Adds a new column to case export Excel sheets "Complainant_Reference"
- This will display as "Restricted" for restricted cases
- Will be null/blank if not set
- Changes the casing of "Reported_as" to be "Reported_As" in order to be consistent with the casing of other fields

### DIFF
--- a/app/models/case_export.rb
+++ b/app/models/case_export.rb
@@ -104,7 +104,8 @@ private
                      Date_Validated
                      Case_Creator_Team
                      Notifying_Country
-                     Reported_as]
+                     Reported_As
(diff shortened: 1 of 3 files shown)


nasirkhanbeis (approved):
looks good as expected.

KyleMacPherson (approved):

PR #2139, opened 2022-11-23
Implement expiration of products after 18 months

stdstaff:
https://regulatorydelivery.atlassian.net/jira/software/projects/PST1/boards/53?selectedIssue=PST1-29

## Description

 - A product record which has not been linked to an open case for more than 18 months can not be added to a case - attempts in doing so return an error warning (on the 'add a product record to your case' page).
 - For TSO users (not OPSS users), attempting to navigate directly to the product record - which is "expired/retired/soft-deleted" should return the product not found page. (OPSS users can open retired product records).
 - A retired product product is not editable
 - A case cannot be created from a retired product


app/policies/product_policy.rb
```diff
@@ -1,9 +1,22 @@
...
   def update?
-    record.version.nil? && (record.owning_team.nil? || record.owning_team == user.team)
+    return false if record.version.present?
+    return false if record.retired? && !user.is_opss?
+
+    record.owning_team.nil? || record.owning_team == user.team
+  end
+
+  def can_spawn_case?
+    !user.is_opss? && record.not_retired?
```
KyleMacPherson, line 20: [Minor] I think the `!user.is_opss` is just based on a design decision on the different case creation flows for opss and non-opss users, I don't think it is a theoretical policy that opss users should not be able to spawn a case from a product so I would probably just have this is `record.not_retired?`. Feel free to ignore though, looks like it works well for our needs here 👍

KyleMacPherson (approved):

app/policies/product_policy.rb
stdstaff: Thanks! I have discussed this with the product owner and subsequently altered the permissions; no user will be able to update a retired product or spawn a case from one.

nasirkhanbeis (approved):
Tested this on the review app, orphan products older than 18 months are being expired nightly as expected.  
Verified this with a stage DB copy on the review app.
Note: retired products are still indexed and searchable which will be addressed in a separate ticket.

db/migrate/20221117160717_add_retired_at_to_products.rb
```diff
@@ -0,0 +1,8 @@
+class AddRetiredAtToProducts < ActiveRecord::Migration[6.1]
+  disable_ddl_transaction!
+
+  def change
+    add_column :products, :retired_at, :datetime
```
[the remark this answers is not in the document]
stdstaff: An index is added on the line below, will this suffice?

stdstaff:
Thanks — visibility in the search results is to be addressed in an upcoming ticket, however I do not have clarity around the other points raised — I have forwarded these to product managment and will cover them here, thanks for raising!

PR #2240, opened 2023-02-02
New design for the Product page

stdstaff:
https://regulatorydelivery.atlassian.net/browse/PSD-1171

## Description

- A redesigned Product show page
- Uses an Accordion to separate product details from Images and Cases
- Adds a left-hand navigation bar to the product details area
- Creates a consistent header look and feel across cases and products


nasirkhanbeis (approved):

PR #2251, opened 2023-02-07
Added cases in new style to product page

acallaghan:
https://regulatorydelivery.atlassian.net/browse/PSD-1172

Adds a redesigned cases display for the product page

## Checklist:

### General testing (author)
- [x] Test without JavaScript
- [x] Test on small screen

- [x] Works keyboard only


spec/features/product_versioning_spec.rb
```diff
@@ -44,7 +44,7 @@
     visit "/cases/#{investigation.pretty_id}/products"
 
     expect(page).to have_selector("h3.govuk-heading-m", text: product.name)
-    expect(page).to have_selector("dd.govuk-summary-list__value", text: "psd-#{product.id}_#{investigation.date_closed.to_i} - The PSD reference for this version of the product record - as recorded when the case was closed: #{investigation.date_closed.to_s(:govuk)}.")
+    expect(page).to have_selector("dd.govuk-summary-list__value", text: "psd-#{product.id}_#{investigation.date_closed.to_i} - The PSD reference for this version of the product record - as recorded when the case was closed: #{investigation.date_closed.to_fs(:govuk)}.")
```
stdstaff, line 47: We don't use `to_fs` anywhere else in the codebase, but we do use the longer form `to_formatted_s` quite a bit, so it might be preferable to keep the convention?

```suggestion
    expect(page).to have_selector("dd.govuk-summary-list__value", text: "psd-#{product.id}_#{investigation.date_closed.to_i} - The PSD reference for this version of the product record - as recorded when the case was closed: #{investigation.date_closed.to_formatted_s(:govuk)}.")
```

app/views/products/_cases.html.erb
```diff
@@ -0,0 +1,15 @@
...
+    </p>
+  </div>
+</div>
+
+<% if @product.investigations.any? %>
+  <dl class="govuk-summary-list opss-summary-list-mixed opss-summary-list-mixed--sm-font opss-summary-list-mixed--4-cols psd-case-card">
+    <% @product.investigations.each do |investigation| %>
+      <%= investigation.details_summary_list %>
+    <% end %>
+  </dl>
```
stdstaff, line 14: Rather than rendering HTML and using a HTML-outputting helper, it's preferrable to use the design system helpers, which take Ruby objects:
```rb
<&= govukSummaryList(classes: "opss-summary-list-mixed ...", rows: [...]) %> 
```
When the design system changes, the helpers library takes care of generating the correct HTML.

acallaghan: So I had to customise the output here as the current `govukSummaryList` helper doesn't support adding multiple value fields per row, and didn't want to add it just for this. I talked to Darren, and he wanted to add it like in the designs

spec/features/product_versioning_spec.rb
acallaghan: 👍🏻

app/views/products/_cases.html.erb
stdstaff: Ah right, in that case you could consider editing https://github.com/OfficeForProductSafetyAndStandards/product-safety-database/blob/master/app/views/components/_govuk_summary_list.html.erb, which is our service-wide override for the component. The `govukSummaryList` helper will render this partial instead of the stock one. You could add support for `rows: [ { values: [ { text: "xxx" }, { html: "yyy" } ] } ]` or something similar.

The service already uses a non-standard Summary List with secondary text, it may be worth checking if the multiple-values design pattern will be used more than once and taking a view on whether there's value adding it to the component override.

app/assets/stylesheets/_width_overrides.scss
```diff
@@ -41,34 +41,6 @@
...
-  }
-
-  div.psd-case-card a:focus:before,
-  div.psd-product-card a:focus:before {
-    outline: none;
-    background-color: none;
-    box-shadow: none;
-    text-decoration: none;
-  }
-
```
stdstaff, line 71: If you're sure that case cards are no longer rendered anywhere, please remove e.g. `app/views/investigations/_case_card.html.erb`, `_highlight_card.html.erb`, `_restricted_card.html.erb`, `_case_card_status_area.html.erb` as appropriate. It'll be nice to finally tidy away these old partials 🙂

acallaghan: i may push these back until the next PR - as that is more of a 'tidy things up PR' than this one is

app/decorators/investigation_decorator.rb
```diff
@@ -46,6 +46,24 @@ def risk_level_description
...
 
+  def details_summary_list
+    tag_class_name = is_closed? ? "opss-tag--risk3" : "opss-tag--plain"
+    action = h.tag.span("Case #{status}", class: "opss-tag #{tag_class_name}")
+
+    h.tag.div class: "govuk-summary-list__row" do
+      list = []
+      if object.is_private?
+        list << h.tag.dt("Allegation restricted", class: "govuk-summary-list__key")
+      else
```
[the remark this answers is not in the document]
acallaghan: I've update this now -

AndrewCowle (approved):

PR #2257, opened 2023-02-13
Change the edit permissions for unowned products

stdstaff:
https://regulatorydelivery.atlassian.net/browse/PSD-1341

## Description

Changes the edit permissions for unowned products, and behaviour around product ownership:

- Products may only be edited by users in their owning team. If the product doesn't have an owning team, then they may only be edited by users whose teams have open cases which are linked to the product
- Newly created products do not have an owner

Has also expanded to include some visual changes required to support this:

- Normalises the product show page summary list with the list from the case products list (these are now visually the same)
- Addresses a snag with the wording for the PSD reference secondary text in the above list
- Addresses a snag with missing secondary text for "Market date"


KyleMacPherson (approved):

nasirkhanbeis (approved):

PR #2259, opened 2023-02-14
Remove case restrictions

acallaghan:
Removes the ability for cases to be restricted on PSD. Existing cases that are open cannot be made restricted, but previously restricted cases can still be made public.

https://regulatorydelivery.atlassian.net/browse/PSD-1062


AndrewCowle (approved):

nasirkhanbeis (approved):
Tested this on the review app, looking all good as expected.

PR #2272, opened 2023-02-16
Versioning of images

AndrewCowle:
JIRA ticket: https://regulatorydelivery.atlassian.net/browse/PSD-1164

## Description

Add a new `DocumentUpload` model for `Product` that stores metadata about and handles the validations for the attached file, and points to the actual file in Active Storage using `has_one_attached`. This replaces the existing direct relationship via `Document` for product images.

Other models that also use the same relationship are not changed in this PR.

A data migration is provided that moves all product images from the existing to the new model relationship.


test/models/document_upload_test.rb
```diff
@@ -0,0 +1,7 @@
+require "test_helper"
+
+class DocumentUploadTest < ActiveSupport::TestCase
+  # test "the truth" do
+  #   assert true
+  # end
+end
```
stdstaff, line 7: Propose removing as the project uses RSpec

test/fixtures/document_uploads.yml
```diff
@@ -0,0 +1,7 @@
+# Read about fixtures at https://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
+
+one:
+  metadata: 
+
+two:
+  metadata:
```
stdstaff, line 7: Propose removing as the project uses factories

app/models/document_upload.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  has_one_attached :file_upload
+
+  store_accessor :metadata, :title
+  store_accessor :metadata, :description
+  store_accessor :metadata, :created_by
+
+  attribute :existing_file_upload_file_id, :string
+
+  validates :file_upload, attached: true, size: { less_than: 100.megabytes, message: "must be smaller than 100MB" }
```
stdstaff, line 14: This seems reasonable but is this a current service requirement?

app/models/product.rb
```diff
@@ -49,7 +48,7 @@ def as_indexed_json(*)
     )
   end
 
-  has_many_attached :documents
+  has_many_attached :documents # Deprecated, to be migrated and removed
```
stdstaff, line 51: We use `bin/rails notes` now and then to pick up code tasks like this...

```suggestion
  # TODO: Remove this once attatchments have been migrated
  has_many_attached :documents
```

app/views/document_uploads/_image_tag.html.erb
```diff
@@ -0,0 +1,6 @@
+<% if image.file_upload.variable? %>
+  <% class_name ||= 'app-image' %>
+  <%= image_tag image.file_upload.variant(resize_to_limit: dimensions), class: class_name, alt: image.title %>
+<% else %>
+  <span>Preview not available: <%= image.title %></span>
```
stdstaff, line 5: The Service Designer may be able to provide markup for this case.

app/views/document_uploads/_placeholder.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="app-document-preview__placeholder" aria-hidden="true">
+  <% if document_upload_spreadsheet?(document) %>
+    <%= image_tag "document_placeholder_spreadsheet.png", alt: "" %>
+  <% else %>
+    <%= image_tag "document_placeholder_document.png", alt: "" %>
+  <% end %>
+  <span class="app-document-preview__filetype">
+    <%= document_upload_file_extension(document)  %>
```
stdstaff, line 8:
```suggestion
    <%= document_upload_file_extension(document) %>
```

app/views/document_uploads/index.json.jbuilder
```diff
@@ -0,0 +1 @@
+json.array! @document_uploads, partial: "document_uploads/document_upload", as: :document_upload
```
stdstaff, line 1: The JSON API has been deprecated and is never used, we have been removing it so no need to include `.jbuilder` views for `index` or `show`.

KyleMacPherson:
Looking good to me. Might be worth adding tests to `spec/features/product_versioning_spec.rb` to assert that a user viewing an investigation associated with a versioned product sees the attached images as they were at the time of the version.

nasirkhanbeis (approved):
Tested this on review app and latest fixes, looking all good as expected.Happy to merge this into feature branch.

- Verify when I edit an image description or title, it creates a new product version - Passed
- Verify when I edit an image, all cases should where the product is linked should show the updated image title and description - Passed
- Verify image title/description is not updated for  timestamped version - Passed
- Verify when I remove the image, then all cases where the product is linked should should show updated images. - Passed
- Verify when I remove the image and upload a new one, all cases where the product is linked should show an updated image - Passed
- Verify I should not be able to edit the image of a timestamped version product - Passed
- Verify product export should include only live products and images - Passed
- Verify when I re-open a closed case then I should not be able to edit product images - Passed
- Verify when I add the same product again and my team is the owner then I should be able to edit images
- Verify when I add an image for the same product, then timestamped version product should not be updated. - Passed
- Verify when I remove an image for the same product, then timestamped version product should not be updated - Passed
- Verify images on the version product are clickable and can open via the product image tab.  - Passed
- Verify when I edit a legacy product image, it creates a new version product. - Passed

PR #2276, opened 2023-02-17
Add/Remove product indexing bug

KyleMacPherson:
https://regulatorydelivery.atlassian.net/jira/software/c/projects/PSD/boards/57?modal=detail&selectedIssue=PSD-1349

## Description
When adding a freshly created product to a case, or removing any product from a case, the investigation related fields are not being updated in open search which results in the product not being properly filtered by the investigation.owner_id and therefore not showing up correctly in the you product/team product pages.

It seems that the product.__elasticsearch__.update_document is not meeting our needs here because it looks for the changed attributes of the instance of product and performs a partial update on these changed attributes, however in this case no product attributes were changing and update_document was not updating the fields related to the investigation. To fix this I have used the index_document method which calls `as_indexed_json` and to re-index all the fields, including the investigation owner_id method which is necessary to fix these bugs.

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/services/add_product_to_case.rb
+++ b/app/services/add_product_to_case.rb
@@ -22,7 +22,9 @@ class AddProductToCase
 
     change_product_owner_if_unowned
 
-    product.__elasticsearch__.update_document
+    investigation.reload.__elasticsearch__.update_document
+
+    product.reload.__elasticsearch__.update_document
 
     context.activity = create_audit_activity_for_product_added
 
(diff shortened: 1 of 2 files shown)


stdstaff (approved):

nasirkhanbeis (approved):

PR #2280, opened 2023-02-20
Fix a bug where products are selectable multiple times when adding risk assessments

stdstaff:

## Description

The `risk_assessed_products` table has a `unique` index on `product_id` and `investigation_id`. The introduction of versioned products on a case produces a problem: A user can add the same `Product` to a case multiple times (by adding it, closing the case, reopening the case, and adding again). This is intended behaviour, however the Risk Assessment creation/edit flow has not been updated to take this into account. Accordingly, in this scenario the same `product_id`/checkbox was presented on the form twice. Selecting _both_ checkboxes for the same logical product caused a 500 error.

Work to address the relationship of supporting information to products is ongoing, however in the meantime, this PR mitigates the issue by deduplicating the list of products, so that each product is only presented to the user once for selection.

I have also made the product list on other supporting information forms (accidents/incidents, corrective actions, and test results) unique. These didn't cause errors but in this scenario the user would be presented with the same option in a dropdown list more than once which would have been confusing.

### DIFF
--- a/app/forms/risk_assessment_form.rb
+++ b/app/forms/risk_assessment_form.rb
@@ -81,7 +81,7 @@ class RiskAssessmentForm
   end
 
   def products
-    investigation.products
+    investigation.products.uniq
     .pluck(:name, :id).collect do |row|
       {
         text: row[0],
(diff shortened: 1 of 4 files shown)


KyleMacPherson (approved):

stdstaff:
Note: The service team will shortly supersede this fix with a larger piece of work, looking at the relationships between supporting information and versioned products: https://regulatorydelivery.atlassian.net/browse/PSD-1359.

PR #2288, opened 2023-02-24
Refactor country picker select JS

acallaghan:
Refactors the country picker JS - the form uses a select for progressive enhancement.
 
This old lib was fetching a huge graph of json objects and including it into the bundle just for this autocomplete. We're already using `accessible-autocomplete` elsewhere, and it can lift out the country names from the select values just fine. So all data is just in the DOM instead of the DOM and in the JS. This saves ~400Kb of JS in the compiled bundle

- Switch libs, removed old country autocomplete lib
- Moved data file for `Country` model to `app/assets`, out of node_modules
- Tidied up location picker JS, remove jQuery and ported to typescript

https://regulatorydelivery.atlassian.net/browse/PSD-1367


AndrewCowle (approved):

nasirkhanbeis (approved):
test this on the review app, looking good as expected.

PR #2290, opened 2023-02-27
Fix sub-nav on the case's products page, where the product appears twice

stdstaff:
https://regulatorydelivery.atlassian.net/browse/PSD-1372

## Description

Fixes a bug which occurs in the sub-nav of a case's products page when the same product is included twice within a case. This PR provides unique IDs for these HTML anchors.


KyleMacPherson (approved):

nasirkhanbeis (approved):

PR #2293, opened 2023-02-28
Remove jQuery (part 1)

rubenarakelyan:
JIRA ticket: https://regulatorydelivery.atlassian.net/browse/PSD-1368

## Description

* Removes jQuery dependency from all JavaScript files at the top level of `app/javascript`
* Removes unused JavaScript
* Updates some syntax


app/views/help/about.html.erb
```diff
@@ -470,8 +470,6 @@
   </div>
 </div>
 
-<%= render "close_page_button" %>
```
rubenarakelyan, line 473: These buttons are never shown as far as I can see

app/views/investigations/creation_flow/about_enquiry.html.erb
```diff
@@ -12,8 +12,8 @@
       <fieldset class="govuk-fieldset">
         <%= form.govuk_date_input :date_received, legend: "When was it received?" %>
         <p class="govuk-body">
-          Set to: <%= link_to('Today', 'javascript:void(0)', id: :today) %>
```
rubenarakelyan, line 15: We should not link to `javascript` links like this - accepted practice is to use `#` instead.

stdstaff (approved):
LGTM from the code perspective, please ask QA to test these components across [the standard browser matrix](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices), and that the accessibility features are all still functioning as expected.

acallaghan (approved):

nasirkhanbeis (approved):
Regression tested this on the review app, looking good as expected.

PR #2295, opened 2023-02-28
Updated change log

bfanshawe:
Description of what significant change will occur.

Update to the change log.

## Description
Update to include what significant change will occur with decoupling work

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## 2020-02-28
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## 2020-02-28
+- Significant service update made to change the way users create products and cases. Products and cases are now two separate entities with associated information, enabling one product to be included in multiple different cases.
```
stdstaff, line 5:
```suggestion
- Significant service update made to change the way users create products and cases. Products and cases are now two separate entities with associated information, enabling one product to be included in multiple different cases.
```
Removes the extra space at the end of the line...

stdstaff (approved):

PR #2298, opened 2023-03-02
Fix investigation_product associated models & audit trail

acallaghan:
Previously the code linked the following data to `product`, instead of `investigation_product`
- Test results 
- Corrective actions
- Risk assessments
- Accident/Incident - 'Record an accident'
- Accident/Incident - 'Record an incident'

This meant that the data model and schema for each of these classes (subclasses of `Investigation`) was wrong, and meant that multiple corrective actions for the same product could be added to the same case (if it was closed and reopened)

The audit trail also was tightly coupled to this association, and also had to change at the same time. The audit trail keeps the data of associations in the base class (through db association) but also in a JSON metadata blob. 

The database schema has changed for both `investigations` and `audit_trails` - and the migration looks up the `investigation_product_id` for each `product_id` via an `investigations` table join.

https://regulatorydelivery.atlassian.net/browse/PSD-1370

## Remaining issues

- [ ] Product XLSX export is failing. I can't figure out how this has stopped working
- [ ] Metadata structure assertion in `spec/services/remove_product_from_case_spec.rb:72`. The `investigation_product.attributes` is returning inconsistent values. Previously this was using a `JSON.parse` to convert first, and this might need to be changed
- [ ] `#psd_ref` is being called on an attachment decorator in a view. I need help to work out why this spec has broken, and whether it's an issue with the changes or the spec.
- [ ] The metadata structure needs to be migrated for `AuditTrail` -> every `product_id` or `product_ids`, including some nested in other objects has to be migrated to use `investigation_product_id(s)`. This *has* to be done as a data migration after this branch is deployed but before users start using PSD


app/models/audit_activity/risk_assessment/risk_assessment_added.rb
```diff
@@ -26,7 +22,7 @@ def risk_assessment_file
   end
 
   def products_assessed
-    Product.find(metadata["risk_assessment"]["product_ids"])
+    InvestigationProduct.find(metadata["risk_assessment"]["investigation_product_ids"])
```
KyleMacPherson, line 25: This method still needs to return the products_assessed. After your change it will return investigation_products. We will need to iterate through the investigation_products and find the corresponding products to return from this method.

PR #2299, opened 2023-03-02
Fix investigation_product associated models & audit trail 

acallaghan:
Previously the code linked the following data to `product`, instead of `investigation_product`
- Test results 
- Corrective actions
- Risk assessments
- Accident/Incident - 'Record an accident'
- Accident/Incident - 'Record an incident'

This meant that the data model and schema for each of these classes (subclasses of `Investigation`) was wrong, and meant that multiple corrective actions for the same product could be added to the same case (if it was closed and reopened)

The audit trail also was tightly coupled to this association, and also had to change at the same time. The audit trail keeps the data of associations in the base class (through db association) but also in a JSON metadata blob. 

The database schema has changed for both `investigations` and `audit_trails` - and the migration looks up the `investigation_product_id` for each `product_id` via an `investigations` table join.

https://regulatorydelivery.atlassian.net/browse/PSD-1370

## Remaining issues

- [ ] Product XLSX export is failing. I can't figure out how this has stopped working
- [ ] Metadata structure assertion in `spec/services/remove_product_from_case_spec.rb:72`. The `investigation_product.attributes` is returning inconsistent values. Previously this was using a `JSON.parse` to convert first, and this might need to be changed
- [ ] `#psd_ref` is being called on an attachment decorator in a view. I need help to work out why this spec has broken, and whether it's an issue with the changes or the spec.
- [ ] The metadata structure needs to be migrated for `AuditTrail` -> every `product_id` or `product_ids`, including some nested in other objects has to be migrated to use `investigation_product_id(s)`. This *has* to be done as a data migration after this branch is deployed but before users start using PSD

### DIFF
--- a/app/decorators/test/result_decorator.rb
+++ b/app/decorators/test/result_decorator.rb
@@ -10,7 +10,7 @@ class Test < ApplicationRecord
                       "Test result"
                     end
 
-      "#{result_text}: #{investigation_product.product.name}"
+      "#{result_text}: #{investigation_product.name}"
     end
 
     def result_text
@@ -71,11 +71,11 @@ class Test < ApplicationRecord
     def is_attached_to_versioned_product?
       !!investigation_closed_at
     end
-  
+
     def investigation_closed_at
       object.investigation_product.investigation_closed_at
     end
-  
+
     def psd_ref
       object.investigation_product.psd_ref
     end
--- a/app/helpers/investigations/accident_or_incidents_helper.rb
+++ b/app/helpers/investigations/accident_or_incidents_helper.rb
@@ -2,10 +2,10 @@ module Investigations
   module AccidentOrIncidentsHelper
     def accident_or_incident_summary_list_rows(accident_or_incident)
       product_value = if accident_or_incident.investigation_product.investigation_closed_at
-        { text: "#{accident_or_incident.investigation_product.name} (#{accident_or_incident.investigation_product.psd_ref})" }
-      else
-        { html: link_to("#{accident_or_incident.investigation_product.name} (#{accident_or_incident.investigation_product.psd_ref})", product_path(accident_or_incident.investigation_product.product)) }
-      end
+                        { text: "#{accident_or_incident.investigation_product.name} (#{accident_or_incident.investigation_product.psd_ref})" }
+                      else
+                        { html: link_to("#{accident_or_incident.investigation_product.name} (#{accident_or_incident.investigation_product.psd_ref})", product_path(accident_or_incident.investigation_product.product)) }
+                      end
 
       rows = [
         { key: { text: "Date of #{accident_or_incident.type.downcase}" }, value: { text: accident_or_incident.date_of_activity } },
--- a/app/helpers/investigations/accident_or_incidents_helper.rb
+++ b/app/helpers/investigations/accident_or_incidents_helper.rb
@@ -2,9 +2,9 @@ module Investigations
   module AccidentOrIncidentsHelper
     def accident_or_incident_summary_list_rows(accident_or_incident)
       product_value = if accident_or_incident.investigation_product.investigation_closed_at
-        { text: "#{accident_or_incident.investigation_product.product.name} (#{accident_or_incident.investigation_product.psd_ref})" }
-      else 
-        { html: link_to("#{accident_or_incident.investigation_product.product.name} (#{accident_or_incident.investigation_product.psd_ref})", product_path(accident_or_incident.investigation_product.product)) }
+        { text: "#{accident_or_incident.investigation_product.name} (#{accident_or_incident.investigation_product.psd_ref})" }
+      else
+        { html: link_to("#{accident_or_incident.investigation_product.name} (#{accident_or_incident.investigation_product.psd_ref})", product_path(accident_or_incident.investigation_product.product)) }
       end
 
       rows = [
--- a/app/models/audit_activity/product/destroy.rb
+++ b/app/models/audit_activity/product/destroy.rb
@@ -3,9 +3,28 @@ class AuditActivity::Product::Destroy < AuditActivity::Product::Base
     { reason:, product: product.attributes }
   end
 
+  def metadata
+    migrate_metadata_structure
+  end
+
 private
 
   def subtitle_slug
     "Product removed"
+  end
+
+  def migrate_metadata_structure
+    metadata = self[:metadata]
+
+    return metadata if already_in_new_format?
+
+    JSON.parse({
+      "reason" => self[:title],
+      "product" => product.attributes
+    }.to_json)
+  end
+
+  def already_in_new_format?
+    self[:metadata]&.key?("product")
   end
 end
--- a/app/models/audit_activity/product/destroy.rb
+++ b/app/models/audit_activity/product/destroy.rb
@@ -1,30 +1,12 @@
 class AuditActivity::Product::Destroy < AuditActivity::Product::Base
-  def self.build_metadata(product, reason)
-    { reason:, product: product.attributes }
+  def self.build_metadata(investigation_product, reason)
+    {
+      reason:,
+      investigation_product: investigation_product.attributes
+    }
   end
-
-  def metadata
-    migrate_metadata_structure
-  end
-
-private
 
   def subtitle_slug
     "Product removed"
-  end
-
-  def migrate_metadata_structure
-    metadata = self[:metadata]
-
-    return metadata if already_in_new_format?
-
-    JSON.parse({
-      "reason" => self[:title],
-      "product" => product.attributes
-    }.to_json)
-  end
-
-  def already_in_new_format?
-    self[:metadata]&.key?("product")
   end
 end
--- a/app/models/audit_activity/product/destroy.rb
+++ b/app/models/audit_activity/product/destroy.rb
@@ -3,28 +3,9 @@ class AuditActivity::Product::Destroy < AuditActivity::Product::Base
     { reason:, product: product.attributes }
   end
 
-  def metadata
-    migrate_metadata_structure
-  end
-
 private
 
   def subtitle_slug
     "Product removed"
-  end
-
-  def migrate_metadata_structure
-    metadata = self[:metadata]
-
-    return metadata if already_in_new_format?
-
-    JSON.parse({
-      "reason" => self[:title],
-      "product" => product.attributes
-    }.to_json)
-  end
-
-  def already_in_new_format?
-    self[:metadata]&.key?("product")
   end
 end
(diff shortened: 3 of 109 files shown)


db/migrate/20230223145950_move_risk_assessed_products_product_refs_to_investigation_products.rb
```diff
@@ -0,0 +1,21 @@
...
+      add_reference :risk_assessed_products, :investigation_product, index: { algorithm: :concurrently }
+      execute "UPDATE risk_assessed_products SET investigation_product_id = ip.id FROM investigations i JOIN risk_assessments ra ON i.id = ra.investigation_id JOIN risk_assessed_products rap ON ra.id = rap.risk_assessment_id JOIN products p ON rap.product_id = p.id JOIN investigation_products ip ON i.id = ip.investigation_id AND p.id = ip.product_id WHERE risk_assessed_products.product_id = p.id;"
+      remove_reference :risk_assessed_products, :product, index: { algorithm: :concurrently }
+      add_index :risk_assessed_products, %i[risk_assessment_id investigation_product_id], unique: true, algorithm: :concurrently, name: "index_risk_assessed_products"
+    end
+  end
+
+  def down
+    safety_assured do
+      # add_reference :risk_assessed_products, :product, index: { algorithm: :concurrently }
```
KyleMacPherson, line 15: We will need to re-instate this down method

app/decorators/test/result_decorator.rb
```diff
@@ -67,5 +67,17 @@ def event_type
 
       supporting_information_type
     end
+
+    def is_attached_to_versioned_product?
```
KyleMacPherson, line 71: These methods are included in multiple decorators, would be good to refactor to DRY up this code

app/models/audit_activity/product/destroy.rb
```diff
@@ -1,30 +1,12 @@
...
-    migrate_metadata_structure
-  end
-
-private
-
   def subtitle_slug
     "Product removed"
   end
-
-  def migrate_metadata_structure
```
KyleMacPherson, line 16: I don't think we can remove this method otherwise the legacy metadata will be different to the "new" metadata

app/views/investigations/activities/accident_or_incident/_accident_or_incident_added.html.erb
```diff
@@ -14,6 +14,3 @@
 
 <%= link_to "View #{activity.type}", investigation_accident_or_incident_path(activity.investigation, activity.metadata["accident_or_incident_id"]), class: "govuk-link psd-block-link govuk-link--no-visited-state" %>
 
-<% if activity.product_id %>
```
KyleMacPherson, line 17: Are we happy to no longer link to a product?

app/models/investigation_product.rb
```diff
@@ -11,11 +11,27 @@ class InvestigationProduct < ApplicationRecord
...
+  has_many :risk_assessments, through: :risk_assessed_products
+
+  belongs_to :added_by_user, class_name: :User, optional: true
+  belongs_to :owning_team, class_name: "Team", inverse_of: :owned_products, optional: true
+
   redacted_export_with :id, :affected_units_status, :batch_number, :created_at,
                        :customs_code, :investigation_id, :number_of_affected_units,
                        :product_id, :updated_at
 
+  delegate :name, to: :product
```
KyleMacPherson, line 28: I think using `product_name`, might be a bit clearer.

spec/models/test/result_spec.rb
```diff
@@ -11,12 +11,12 @@
...
 
-    context "when missing a product" do
-      let(:test_result) { build(:test_result, product: nil) }
+    context "when missing an investigation_product" do
+      let(:test_result) { build(:test_result, investigation_product: nil) }
 
       it "is invalid and includes an error message", :aggregate_failures do
         expect(test_result).not_to be_valid
-        expect(test_result.errors.messages[:product]).to include("Select the product which was tested")
+        expect(test_result.errors.messages[:investigation_product]).to include("must exist")
```
KyleMacPherson, line 19: I don't think the error messages should have changed here.

app/models/audit_activity/test/test_result_updated.rb
```diff
@@ -61,8 +64,8 @@ def show_new_failure_details?
 
   def new_product
     @new_product ||=
-      if updates["product_id"]
-        Product.find(updates["product_id"].second)
+      if updates["investigation_product_id"]
+        InvestigationProduct.find(updates["investigation_product_id"].second)
```
KyleMacPherson, line 68: Feels a bit weird to return an investigation_product from a method called `new_product`. I think we should decide on the desired behaviour and either change the method name or return a product depending on what we decide.

app/models/audit_activity/risk_assessment/risk_assessment_updated.rb
```diff
@@ -105,7 +105,7 @@ def subtitle_slug
   end
 
   def products_assessed
-    Product.find(metadata["product_ids"])
+    InvestigationProduct.find(metadata["investigation_product_ids"])
```
KyleMacPherson, line 108: `new_product_ids`, `new_products` and `products_assessed` all return investigation_products now which i think could be a bit surprising. I think we should decide on what the intended behaviour of these methods are and  either change the method name, or return products as the names suggest

db/migrate/20230223144450_move_all_products_on_audit_trails_to_use_investigation_products.rb
```diff
@@ -0,0 +1,19 @@
+class MoveAllProductsOnAuditTrailsToUseInvestigationProducts < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+
+  def up
+    safety_assured do
+      add_reference :activities, :investigation_product, index: { algorithm: :concurrently }
+      execute "UPDATE activities SET investigation_product_id = investigation_products.id FROM investigation_products WHERE investigation_products.investigation_id = activities.investigation_id AND investigation_products.product_id = activities.product_id;"
```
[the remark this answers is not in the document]
acallaghan: https://github.com/OfficeForProductSafetyAndStandards/product-safety-database/pull/2299/commits/f7a0e892774381bd7a36d34dd1563c23d6ed6a3c

app/decorators/test/result_decorator.rb
acallaghan:
```
  def is_attached_to_versioned_product?
    !!investigation_closed_at
  end

  def investigation_closed_at
    object.investigation_product.investigation_closed_at
  end

  def psd_ref
    object.investigation_product.psd_ref
  end
```
These are repeated but not introduced by my changes, i don't think it's in scope here. We plan to refactor how activities work once we have user feedback

app/models/audit_activity/product/destroy.rb
acallaghan: there are no records left of `AuditActivity::Product::Destroy` in production where metadata is blank, so safe to remove here

app/models/audit_activity/risk_assessment/risk_assessment_updated.rb
acallaghan: I wanted to keep the interface to these classes as close as possible to how they were, otherwise even more view, decorator and form code will have to change. I'm happy to refactor, but i think it can wait & be another ticket

app/services/migrate_metadata_for_audit_trails.rb
```diff
@@ -0,0 +1,53 @@
+class MigrateMetadataForAuditTrails
```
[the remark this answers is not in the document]
acallaghan: We had an issue with migration specs with some of Ruben's code - keeping it separated here makes it easier to test

app/decorators/test/result_decorator.rb
KyleMacPherson: Yep those are the ones! Not sure what you mean, these are introduced in this PR so I think it is scope.

app/models/audit_activity/product/destroy.rb
KyleMacPherson: Ok nice!

app/models/audit_activity/risk_assessment/risk_assessment_updated.rb
KyleMacPherson: Well returning products would keep it exactly the same as it was before, and be quite straightforward.

app/views/investigations/activities/accident_or_incident/_accident_or_incident_added.html.erb
acallaghan: Yeah it was a discussion we had with @DH02 to make the interface simpler & this work easier

KyleMacPherson: 👍 I think we might have missed out `_test_result_updated.html.erb. Looks like we still link to product there, which I think might cause issues if the test result is linked to a versioned product (as I don't think we are not meant to link to versioned products).

app/helpers/investigations/accident_or_incidents_helper.rb
```diff
@@ -1,9 +1,15 @@
 module Investigations
   module AccidentOrIncidentsHelper
     def accident_or_incident_summary_list_rows(accident_or_incident)
+      product_value = if accident_or_incident.investigation_product.investigation_closed_at
+                        { text: "#{accident_or_incident.investigation_product.name} (#{accident_or_incident.investigation_product.psd_ref})" }
+                      else
+                        { html: link_to("#{accident_or_incident.investigation_product.name} (#{accident_or_incident.investigation_product.psd_ref})", product_path(accident_or_incident.investigation_product.product)) }
+                      end
+
```
KyleMacPherson, line 9: Some inconsistent behaviour across the supporting information types. Corrective actions never link to products, but accident and incidents, tests and risk assessments link to the product unless it is an old version.

app/services/migrate_metadata_for_audit_trails.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+  end
+
+  def migrate_corrective_action_updated!
+    AuditActivity::CorrectiveAction::Update.all.each do |object|
+      product_id = object.metadata["updates"]["product_id"]
+      object.metadata["updates"]["investigation_product_id"] = investigation_product_id(investigation: object.investigation, product_id:)
+      object.save!
+    end
+  end
```
KyleMacPherson, line 43: I think we need a migration for `AuditActivity::CorrectiveAction::Add`, `AuditActivity::Test::Result` and `AuditActivity::AccidentOrIncident::AccidentOrIncidentAdded` too

app/helpers/investigations/accident_or_incidents_helper.rb
acallaghan: I've made a jira ticket for this https://regulatorydelivery.atlassian.net/browse/PSD-1402

nasirkhanbeis (approved):
Tested this on the review app, looking all good as expected. 
Also, tested legacy cases with supporting information and audit log.

PR #2346, opened 2023-03-31
Remove 2fa step from review apps, separate smoke test

acallaghan:
## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


slawosz:
Does this removes 2FA from all review apps? If so, its not a way to go as review app has to be exactly the same as production.

PR #2425, opened 2023-05-18
OPSS funded test reports (PSD-1573)

acallaghan:
Create OPSS funded test reports 

https://regulatorydelivery.atlassian.net/browse/PSD-1573

- Adds a form asking the user if the test report was funded by OPSS
- If no, they go to the normal test report form page as it is now
- If yes, they are asked more info about the test certificate, and when it was issued. This is then captured and stored on the test report.

### DIFF
--- a/app/controllers/investigations/test_results/funding_source_controller.rb
+++ b/app/controllers/investigations/test_results/funding_source_controller.rb
@@ -24,6 +24,8 @@ private
     investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
     authorize investigation, :update?
     @investigation = investigation.decorate
+  rescue ActiveRecord::RecordNotFound
+    render_404_page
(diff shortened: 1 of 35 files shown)


app/controllers/investigations/test_results/funding_source_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+      end
+    else
+      render :new
+    end
+  end
+
+private
+
+  def set_investigation
+    investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
```
tobyprivett, line 24: question: would the `find_by!` method raise an exception if nothing is found. Curious how the exception would be handled?

tobyprivett (approved):

kennyevil (approved):

PR #2520, opened 2023-08-09
Refactor search - PSD-1803

acallaghan:
https://regulatorydelivery.atlassian.net/browse/PSD-1803
https://regulatorydelivery.atlassian.net/browse/PSD-1804

## Description

- Replaces all custom search code with `searchkick` gem
- Removes all mention on elasticsearch (in code at least, infra will need to be changed)
- Refactored specs setup to always stub opensearch, tag to opt-out allows for reindexing
- Removes `opensearch` indexes for `Product` and `Business` - both of these are DB only for now
- Refactors Investigation search & filtering helpers, now using `searchkick`
- 'Fuzziness' rules replaces with similar misspelling rules given by `searchkick`

### Code design decisions

- Adds new serializers for `Investigation` and related objects (like Teams, Users, Products etc)
- Serializers are used to generate the hashes that opensearch use for its indexes
- All manual reindexes have been removed in preference of `searchkick` callbacks (this needs to be tested like a lot)
- The amount indexed on `Investigation` greatly reduced - this will be changed shortly once search UI is redesigned 
- Changed Product & Business CSV exports to only use the DB, no `opensearch` involvement 
- Kept `opensearch` for Investigation CSV export - we need to decide if the CBAS/RIIT team need a different style of Case export (one that is on a cron, and just exports everything instead of hitting the search engine)

## Checklist:
- [x] Manual testing on review app
- [x] Decide how/when we are deploying this - A Thursday afternoon/early evening. Re-index will degrade service but not fully bring it down
- [x] If deploying, pick a time that we can reindex after - index will be empty
- [x] Write code decision statement (me and Andrew discussed this at the beginning)


spec/features/search_cases_spec.rb
```diff
@@ -297,6 +295,7 @@
         end
 
         it "only shows the case when searching by product details from the time the case was closed" do
+          pending "This is completely mental and should probably not have gone live. This is matching cases to the old & updated attributes of the products?"
```
rubenarakelyan, line 298: 😂

spec/features/filter_cases_spec.rb
```diff
@@ -178,7 +178,7 @@
         choose "Others", id: "case_owner_others"
         click_button "Apply"
         expect(page).not_to have_listed_case(investigation.pretty_id)
-        expect(page).to have_listed_case(other_user_investigation.pretty_id)
+        expect(page).not_to have_listed_case(other_user_investigation.pretty_id)
```
rubenarakelyan, line 181: Is this intentional?

rubenarakelyan (approved):

spec/features/search_cases_spec.rb
```diff
@@ -65,9 +64,6 @@
 
     # Full product name should be shown
     expect(page).to have_text("MyBrand washing machine")
-
-    # The part of the product name which matches the search term should be highlighted
-    expect(page).to have_selector("em", text: "MyBrand")
```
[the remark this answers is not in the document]
acallaghan: Yeah that's right -there's a plan from Dimitry to re-add it soon

spec/features/filter_cases_spec.rb
acallaghan: Hmm, good catch! Ill take a look at that

PR #2872, opened 2024-02-01
Restrict to new reporting role - PSD-2398

acallaghan:
https://regulatorydelivery.atlassian.net/browse/PSD-2398

## Description
- Restricts reports portal to new 'reporting' role, instead of 'opss'
- Fixes 2fa, due to incorrect concerns being loaded in reporting/base_controller class
- Fixes seeds after notification refactor

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -118,7 +118,7 @@ usages = UnexpectedEvent.usages.values
     is_date_known: true
   }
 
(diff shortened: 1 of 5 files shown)


rubenarakelyan (approved):

rubenarakelyan:
The sentry concern file name need to be renamed to remove `reporting` then I think this should work

PR #3003, opened 2024-04-05
PSD-2481 - Test report and Risk assessment hyperlinks

Nashar101:
This Branch addresses the Jira ticket PSD-2481: https://regulatorydelivery.atlassian.net/jira/software/c/projects/PSD/boards/57?selectedIssue=PSD-2481

Description
I have changed the text adjacent to Risk assessments in the Notification page to a hyperlink that when clicked would redirect the user straight to the show page of that risk assessment

Screen-shots or screen-capture of UI changes

image

image

### DIFF
--- a/app/views/notifications/show.html.erb
+++ b/app/views/notifications/show.html.erb
@@ -166,7 +166,7 @@
     <%= govuk_button_link_to "Add business", new_investigation_business_types_path(@notification), secondary: true %>
     <h2 class="govuk-heading-m">Evidence</h2>
     <% @notification.investigation_products.decorate.each do |investigation_product| %>
-      <% risk_assessment = Product.find(investigation_product.product_id).risk_assessments %>
+      <% risk_assessment = Investigation.find_by(pretty_id: @notification.pretty_id).risk_assessments %>
       <%= risk_assessment.first.id %>
       <% risk_text = "" %>
       <% if risk_assessment.nil? %>
--- a/app/views/notifications/show.html.erb
+++ b/app/views/notifications/show.html.erb
@@ -166,6 +166,13 @@
     <%= govuk_button_link_to "Add business", new_investigation_business_types_path(@notification), secondary: true %>
     <h2 class="govuk-heading-m">Evidence</h2>
     <% @notification.investigation_products.decorate.each do |investigation_product| %>
+      <% risk_assessment_id = Product.find(investigation_product.product_id).risk_assessments.find_by(investigation_id: investigation_product.id) %>
+      <% risk_text = "" %>
+      <% if risk_assessment_id.nil? %>
+        <% risk_text = "Not provided" %>
+      <% else %>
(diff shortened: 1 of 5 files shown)


alan-at-work (commented):
Interesting way of approaching this. Do we have test coverage on this to make sure the links are being built correctly?

rubenarakelyan:
Why not just change what `formatted_risk_assessments` does to include links? That method is only used in two places, both of which are part of the notification task list workflow. The method is a helper and should pass a complete string to the view to display, so that you don't have to construct a string inside the view itself.

Nashar101:
> Why not just change what `formatted_risk_assessments` does to include links? That method is only used in two places, both of which are part of the notification task list workflow. The method is a helper and should pass a complete string to the view to display, so that you don't have to construct a string inside the view itself.

I have removed the hyperlink function and added the changes to the 'formatted_risk_assessments' function

Nashar101:
> Interesting way of approaching this. Do we have test coverage on this to make sure the links are being built correctly?

Created Rspec for new feature

alan-at-work (approved):
Nice work

app/views/notifications/show.html.erb
```diff
@@ -153,7 +153,7 @@
             },
             {
               key: { text: "Address" },
-              value: { text: formatted_business_address(investigation_business.business.locations.find_by(name: "Registered office address")).html_safe.presence || "Not provided" }
+              value: { text: "Not provided" || "Not provided" }
```
rubenarakelyan, line 156: Is this intentional?

app/views/notifications/show.html.erb
```diff
@@ -166,6 +166,14 @@
     <%= govuk_button_link_to "Add business", new_investigation_business_types_path(@notification), secondary: true %>
     <h2 class="govuk-heading-m">Evidence</h2>
     <% @notification.investigation_products.decorate.each do |investigation_product| %>
+      <% risk_assessment = Investigation.find_by(pretty_id: @notification.pretty_id).risk_assessments %>
+      <%= risk_assessment.first.id %>
+      <% risk_text = "" %>
+      <% if risk_assessment.nil? %>
+        <% risk_text = "Not provided" %>
```
rubenarakelyan, line 173: Can this be done in the `formatted_risk_assessments` method rather than the view?

app/views/notifications/show.html.erb
alan-at-work: I did wonder about this, and why it was duplicated.

app/views/notifications/show.html.erb
Nashar101: > Can this be done in the `formatted_risk_assessments` method rather than the view?

I have added the requested changes, the conditional statement is now part of the helper method

app/views/notifications/show.html.erb
Nashar101: Unintentional mistake will revert

alan-at-work (approved):

PR #3010, opened 2024-04-16
feat/PSD-2482-view-prod-summary-from-notif-summary

tahir-khalid:
## Description
As a PSD user I want to view product from a notification summary post submission so that I can check Product details without navigating away from the summary page

### Before:
image

### After:

image

### DIFF
--- a/app/helpers/notifications/create_helper.rb
+++ b/app/helpers/notifications/create_helper.rb
@@ -65,7 +65,7 @@ module Notifications
         end
     end
 
-    def number_of_affected_units(investigation_products)
+    def number_of_affected_units(investigation_products, is_link: false)
       investigation_products.map { |investigation_product|
         units = case investigation_product.affected_units_status
                 when "exact"
@@ -79,8 +79,20 @@ module Notifications
                 else
                   "Not provided"
                 end
-        "#{investigation_product.product.decorate.name_with_brand}: #{units}"
+        if is_link
+          investigation_product_units_with_link(investigation_product, units)
+        else
+          investigation_product_units(investigation_product, units)
+        end
       }.join("<br>")
+    end
+
+    def investigation_product_units_with_link(investigation_product, units)
+      "#{link_to investigation_product.product.decorate.name_with_brand, product_path(investigation_product.product_id) }  #{units}"
+    end
+
+    def investigation_product_units(investigation_product, units)
+      "#{investigation_product.product.decorate.name_with_brand}: #{units}"
     end
 
     def investigation_products_options
@@ -110,7 +122,17 @@ module Notifications
     end
 
     def formatted_business_address(location)
-      [location.address_line_1, location.address_line_2, location.city, location.county, location.postal_code, country_from_code(location.country)].map(&:presence).compact.join("<br>")
+      address_parts = [
+        location&.address_line_1,
+        location&.address_line_2,
+        location&.city,
+        location&.county,
+        location&.postal_code,
+        location&.country.nil? ? nil : country_from_code(location.country)
+      ]
+
+      formatted_address = address_parts.compact.join("<br>")
+      formatted_address.presence || "Address not available"
     end
 
     def formatted_business_contact(contact)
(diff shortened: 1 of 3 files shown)


app/helpers/notifications/create_helper.rb
```diff
@@ -79,10 +79,22 @@ def number_of_affected_units(investigation_products)
                 else
                   "Not provided"
                 end
-        "#{investigation_product.product.decorate.name_with_brand}: #{units}"
+        if is_link
```
alan-at-work, line 82: `investigation_product_units_with_link` and `investigation_product_units` isn't that clear from their names. 
Could we rename them to better reflect their purpose?

Maybe extract these out and do 

```
def with_link(investigation_product, units_text)
  "#{link_to investigation_product.product.decorate.name_with_brand, product_path(investigation_product.product_id) }  #{units_text}"
end
```

```
def without_link(investigation_product, units_text)
  "#{investigation_product.product.decorate.name_with_brand}: #{units_text}"
end
```

app/helpers/notifications/create_helper.rb
```diff
@@ -65,7 +65,7 @@ def countries_options
         end
     end
 
-    def number_of_affected_units(investigation_products)
+    def number_of_affected_units(investigation_products, is_link: false)
```
alan-at-work, line 68: I would start to look a cleaning this up to make it more maintainable and easier to read?

Maybe something like this...

```
def number_of_affected_units(investigation_products, is_link: false)
  investigation_products.map do |investigation_product|
    units_text = units_text_for_product(investigation_product)
    is_link ? with_link(investigation_product, units_text) : without_link(investigation_product, units_text)
  end.join("<br>")
end
```

app/helpers/notifications/create_helper.rb
```diff
@@ -65,7 +65,7 @@ def countries_options
...
       investigation_products.map { |investigation_product|
         units = case investigation_product.affected_units_status
```
alan-at-work, line 70: Bring this out too 

```
def units_text_for_product(investigation_product)
  case investigation_product.affected_units_status
  when "exact"
    investigation_product.number_of_affected_units
  when "approx"
    "Approximately #{investigation_product.number_of_affected_units}"
  when "unknown"
    "Unknown"
  when "not_relevant"
    "Not relevant"
  else
    "Not provided"
  end
end
```

alan-at-work (commented):
Looks good, some ideas which could clean up the code, WDYT?

alan-at-work:
Also need to use rubocop to help fix the rouge whitespace

alan-at-work:
rebased with main

alan-at-work (approved):
Changes look good for this initial bit of work.

app/helpers/notifications/create_helper.rb
```diff
@@ -65,24 +65,36 @@ def countries_options
...
+        "Unknown"
+      when "not_relevant"
+        "Not relevant"
+      else
+        "Not provided"
+      end
+    end
+
+    def with_link(investigation_product, units_text)
+      "#{link_to "#{investigation_product.product.psd_ref} - #{investigation_product.product.decorate.name_with_brand}", product_path(investigation_product.product_id), target: '_blank', rel: 'nofollow noopener'}  #{units_text}"
```
rubenarakelyan, line 91: There should be a colon here to match with the "no link" version

docker/env.psd
```diff
@@ -1,6 +1,6 @@
 # Note: These are DEVELOPMENT values only - not used in production
 
-DATABASE_URL=postgres://postgres:password@db:5432/
+DATABASE_URL=postgresql://postgres:password@db:5432
```
rubenarakelyan, line 3: Don't think this should be here...

rubenarakelyan (changes requested):

docker/env.psd
tahir-khalid: IDE being crap have restored the file.

PR #3021, opened 2024-04-29
[PSD-2536] - updated the message for risk level validation

meeravnagesh:
When a risk is validated by IMT , the messaged shown will be "The notification risk level has been validated"
image


alan-at-work (approved):
Looks good

PR #3028, opened 2024-05-07
PSD-2577 - Merge country data

Nashar101:
This ticket adds the columns Overseas Regulator and Country to the Notifications export excel spreadsheet between the Notifying country column and Trading Standards Region column
I found a way to export the notifications from local PSD by using the rails console in a CSV format, then displaying that CSV file in an excel spreadsheet so I could be sure that the changes were made.

Before
image

After
image


Nashar101:
This is ready to merge, i found a way to export the notifications from local PSD by using the rails console in a CSV format, then displaying that CSV file in an excel spreadsheet so I could be sure that the changes were made.

This ticket adds the columns Overseas Regulator and Country to the Notifications export excel spreadsheet
Before
image

After
image

Nashar101:
> Can you please update the description with the PR is for as the work doesn't seem to match the ticket.
> 
> Also on a side note, we need to make sure we are using the branching format `feat/PSD-123-my-new-feature` for example this would have been `refactor/PSD-2577-merge_country_data` for example.

I have changed the description of the PR

Nashar101:
@alan-at-work @meeravnagesh please review

PR #3062, opened 2024-06-10
PSD-2569: Refactor Products Form

Nashar101:
Description
Refactored Products Form to use GOVUKDesignSystemFormBuilder instead of  ApplicationFormBuilder and made changes to the product related rspecs where necessary

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -3,7 +3,6 @@ default: &default
   encoding: unicode
   url: <%= ENV.fetch('DATABASE_URL', 'postgres://postgres@localhost:5432') %>
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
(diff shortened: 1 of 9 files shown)


config/database.yml
```diff
@@ -3,6 +3,7 @@ default: &default
   encoding: unicode
   url: <%= ENV.fetch('DATABASE_URL', 'postgres://postgres@localhost:5432') %>
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  password: "password"
```
alan-at-work, line 6: Not sure about adding this in, this should be setup in you .env file

Nashar101: My bad, that is not supposed to be there will remove it

alan-at-work (approved):

PR #3068, opened 2024-06-11
refactor/PSD-2660-business_form

Nashar101:
Description: I have refactored all the Business associated forms in this branch as the form partials would sometimes be called together in one html.erb view, this meant that if one form used the ApplicationFormBuilder system and another partial form used the GOVUKDesignSystemFormBuilder system, the forms would fail to render.

this pull request does the ticket work of PSD-2660, PSD-2661, PSD-2662, and PSD-2664.


alan-at-work (approved):
Nice work mate. Looks good, get this past QA.

PR #3074, opened 2024-06-19
refactor/PSD-2663-investigations_form

Nashar101:
Description: Refactored forms located in the investigations views to use the GOVUKDesignSystemFormBuilder builder for the form.
Because some form models are not descendants of ApplicationRecord and are ActiveModels and they have date attributes, I had to handle the date field in the forms as the date field was sending an array of 3 items and they did not know how to handle it.
Fixed the model specs to reflect these changes.

This pull request addresses PSD-2665, PSD-2669, PSD-2670, PSD-2671, PSD-2752, PSD-2672.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -3,7 +3,6 @@ default: &default
   encoding: unicode
   url: <%= ENV.fetch('DATABASE_URL', 'postgres://postgres@localhost:5432') %>
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
-  password: "password"
 development:
   <<: *default
   database: psd_development
--- a/config/database.yml
+++ b/config/database.yml
@@ -3,6 +3,7 @@ default: &default
   encoding: unicode
   url: <%= ENV.fetch('DATABASE_URL', 'postgres://postgres@localhost:5432') %>
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  password: "password"
 development:
   <<: *default
   database: psd_development
--- a/app/controllers/investigations/accident_or_incidents_controller.rb
+++ b/app/controllers/investigations/accident_or_incidents_controller.rb
@@ -65,7 +65,7 @@ module Investigations
 
     def accident_or_incident_params
       accident_or_incident_date_params = params[:accident_or_incident_form]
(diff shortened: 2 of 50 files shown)


alan-at-work:
This is really great work @Nashar101, but I think it will take a little bit to get through as a review :) please bare with :D

config/database.yml
tahir-khalid, line 1: Do we need the password in the database config file (I have been told off by Meera for leaving it in there lol)?

tahir-khalid (changes requested):
Approving

config/database.yml
```diff
@@ -3,6 +3,7 @@ default: &default
   encoding: unicode
   url: <%= ENV.fetch('DATABASE_URL', 'postgres://postgres@localhost:5432') %>
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  password: "password"
```
alan-at-work: remove please :) and rebase

config/database.yml
Nashar101: No, I don't know why it was changed I did not add it to files to commit, I will rectify this change

app/decorators/audit_activity/corrective_action/update_decorator.rb
```diff
@@ -15,7 +15,9 @@ def new_date_decided
   end
 
   def new_legislation
-    return unless (legislation = metadata.dig("updates", "legislation", 1))
+    legislation = metadata.dig("updates", "legislation", 1)
+    Rails.logger.debug { "Legislation Metadata: #{metadata}" } # Debugging output
```
alan-at-work, line 19: Why are you including these files? I am suspecting a rebase was not done properly?

app/models/notification_export.rb
```diff
@@ -35,7 +34,9 @@ def to_spreadsheet
 
     notification_ids.each_slice(FIND_IN_BATCH_SIZE) do |batch_notification_ids|
       find_notifications(batch_notification_ids).each do |notification|
-        sheet.add_row(serialize_notification(notification.decorate), types: :text)
+        row_data = serialize_notification(notification.decorate)
```
alan-at-work, line 37: Why are you including these files? I am suspecting a rebase was not done properly?

alan-at-work (changes requested):
There seems to be a lot of files her from other commits other than your own work. We need to look at why this is happening as you should only have you work in your PR.

app/models/notification_export.rb
Nashar101: I do not know why it was not part of the files which had conflicts, the files which had conflicts were the ones which i modified I did not touch this file.

alan-at-work:
There are so many issues with this branch, not your fault or code. But I think we need to look to break these down into smaller chunks of work. I will have a think on how we can do this :) 

Then it would be best to do each part as its own branch/PR

PR #3080, opened 2024-06-25
refactor/PSD-2829 -investigations _ts _investigations_form

Nashar101:
Description
This pull request makes changes to the PSD 1.0 case creation journey to use the GOVUKSystemDesignFormBuilder as the default builder for the forms, it also makes changes to the specs to reflect the HTML changes caused by the new form builder.


alan-at-work (approved):
All looks to be ok.

PR #3084, opened 2024-06-26
Refactor/PSD-2671: Update to Correspondence Email form

Nashar101:
Description
Refactored Correspondence Email form to use the GOVUK form builder instead of Application form builder, added error presenter helper and added spec for GOVUK error presenter


alan-at-work (approved):
Looks good

PR #3090, opened 2024-06-28
chore/PSD-2864-Adding_missing_validation_to_form

Nashar101:
previous team forgot to add maximum length validation to notification user title of notification name in the notification tasklist journey, if the user entered a notification name longer than 100 characters they would be given a 422 error.

### DIFF
--- a/app/forms/change_notification_details_form.rb
+++ b/app/forms/change_notification_details_form.rb
@@ -9,7 +9,7 @@ class ChangeNotificationDetailsForm
   attribute :current_user
   attribute :notification_id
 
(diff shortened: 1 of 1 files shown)


app/forms/change_notification_details_form.rb
```diff
@@ -9,7 +9,7 @@ class ChangeNotificationDetailsForm
   attribute :current_user
   attribute :notification_id
 
-  validates :user_title, :reported_reason, presence: true
+  validates :user_title, :reported_reason, presence: true, length: { maximum: 100 }
```
meeravnagesh, line 12: pull out the user title as a separate line

meeravnagesh (approved):

PR #3148, opened 2024-08-02
refactor/PSD-2519:rspec_and_coverage_updates

Nashar101:
Description
Increased spec coverage for helpers


.byebug_history
```diff
@@ -0,0 +1,16 @@
+c
```
alan-at-work, line 1: Need to remove this file as it shouldn't really get committed :)

app/helpers/notifications_helper.rb
```diff
@@ -4,10 +4,10 @@ def show_edit_link?
   end
 
   def collaborator_access(collaborator)
-    case collaborator
```
alan-at-work, line 7: Not sure this is the best way to do this, you should not need to convert it to a string?

If the values are symbols (how it was done), you should use symbols directly for comparison. Here's how you can do it:

```ruby
def collaborator_access(collaborator)
  case collaborator
  when Collaboration::Access::Edit
    "Edit"
  when Collaboration::Access::ReadOnly
    "View"
  else
    "Unknown Access Level"
  end
end
```

If collaborator is a string and the constants are symbols, you can use to_sym to convert the string to a symbol for comparison:

```ruby
def collaborator_access(collaborator)
  case collaborator.to_sym
  when Collaboration::Access::Edit
    "Edit"
  when Collaboration::Access::ReadOnly
    "View"
  else
    "Unknown Access Level"
  end
end
```

spec/features/add_and_remove_a_business_spec.rb
```diff
@@ -384,7 +384,6 @@ def record_a_corrective_action
 
     within_fieldset "Which business is responsible?" do
       choose "Trading name (Retailer)"
-      # TODO: add test here once business selection is possible
```
alan-at-work, line 387: This has not been finished, so why remove the TODO?

spec/features/assert_formatted_product_and_notification_methods_spec.rb
```diff
@@ -252,7 +252,6 @@
 
       within_fieldset "Which business is responsible?" do
         choose "Trading name (Retailer)"
-        # TODO: add test here once business selection is possible
```
alan-at-work, line 255: This has not been finished, so why remove the TODO?

spec/features/visit_risk_assessment_from_notification_task_list_page_spec.rb
```diff
@@ -243,7 +243,6 @@
 
     within_fieldset "Which business is responsible?" do
       choose "Trading name (Retailer)"
-      # TODO: add test here once business selection is possible
```
alan-at-work, line 246: This has not been finished, so why remove the TODO?

spec/features/add_and_remove_a_business_spec.rb
Nashar101: It is done the business selection is the choose "Trading name" there

spec/features/assert_formatted_product_and_notification_methods_spec.rb
Nashar101: It is done the business selection is the choose "Trading name" there

spec/features/visit_risk_assessment_from_notification_task_list_page_spec.rb
Nashar101: It is done the business selection is the choose "Trading name" there

.byebug_history
Nashar101: I did not commit this, will remove

app/helpers/notifications_helper.rb
alan-at-work: fair enough 👍

spec/features/add_and_remove_a_business_spec.rb
alan-at-work: fair enough 👍

spec/features/assert_formatted_product_and_notification_methods_spec.rb
alan-at-work: fair enough 👍

spec/features/visit_risk_assessment_from_notification_task_list_page_spec.rb
alan-at-work: fair enough 👍

alan-at-work (approved):

PR #3152, opened 2024-08-05
fix/psd-2559-ar-timeout-error

tahir-khalid:
PSD-2559 subtask: https://regulatorydelivery.atlassian.net/browse/PSD-2603 (details of the change and reasoning in this ticket).

## Description
Adjusted connection pool settings to try to improve performance and avoid connection timeout errors.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -11,10 +11,16 @@ test:
   <<: *default
   database: psd_test<%= ENV['TEST_ENV_NUMBER'] %>
 
-
 production:
   adapter: postgresql
   url: <%= ENV['RAILS_DATABASE_URL'] %>
+  # Sets the connection pool size, using the RAILS_MAX_THREADS environment variable if available, otherwise defaults to 15
+  pool: <%= ENV.fetch('RAILS_MAX_THREADS', 15) %>
+  # Sets the maximum time (in milliseconds) to wait for any database operation to complete (the entire scope of operation)
+  timeout: 5000
+  # Sets the maximum time (in seconds) to wait when establishing a new database connection
   connect_timeout: 2
   variables:
+    # Sets a PostgreSQL-specific timeout for individual SQL statements,
+    # using the STATEMENT_TIMEOUT environment variable if available, otherwise defaults to 500 milliseconds
     statement_timeout: <%= ENV["STATEMENT_TIMEOUT"] || "500ms" %>
--- a/config/database.yml
+++ b/config/database.yml
@@ -14,13 +14,8 @@ test:
 production:
   adapter: postgresql
   url: <%= ENV['RAILS_DATABASE_URL'] %>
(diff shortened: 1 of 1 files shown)


config/database.yml
```diff
@@ -11,10 +11,16 @@ test:
   <<: *default
   database: psd_test<%= ENV['TEST_ENV_NUMBER'] %>
 
-
 production:
   adapter: postgresql
   url: <%= ENV['RAILS_DATABASE_URL'] %>
+  # Sets the connection pool size, using the RAILS_MAX_THREADS environment variable if available, otherwise defaults to 15
+  pool: <%= ENV.fetch('RAILS_MAX_THREADS', 15) %>
```
alan-at-work, line 18: you have added this, do you know is this is set on prod?

config/database.yml
```diff
@@ -11,10 +11,16 @@ test:
...
+  # Sets the maximum time (in milliseconds) to wait for any database operation to complete (the entire scope of operation)
+  timeout: 5000
+  # Sets the maximum time (in seconds) to wait when establishing a new database connection
   connect_timeout: 2
   variables:
-    statement_timeout: <%= ENV["STATEMENT_TIMEOUT"] || "500ms" %>
+    # Sets a PostgreSQL-specific timeout for individual SQL statements,
+    # using the STATEMENT_TIMEOUT environment variable if available, otherwise defaults to 500 milliseconds
+    statement_timeout: <%= ENV["STATEMENT_TIMEOUT"] || "500ms" %>
```
alan-at-work, line 26: not sure on this change as you have just added some comments, are they needed? and do you know what its set to on prod?

config/database.yml
tahir-khalid: I based this on my review of the `manifest.yml` file there is an env section that defines it.  I will remove if its not present in PROD.

config/database.yml
tahir-khalid: Comments was for clarity will remove them and I am unable to confirm what is in PROD I based my findings on what I discovered in the codebase.

config/database.yml
alan-at-work: If you are to up the pool it needs to be done server side, especially if you want to use the `RAILS_MAX_THREADS`
Also comments are not needed. :)

config/database.yml
alan-at-work: Yeah it's not really needed, so should be removed.

config/database.yml
tahir-khalid: Removed left the statement in there as it should fallback to the 15 value if that ENV var is not set for now then we can raise another PR in future to set those values once we have calculated what they are.

config/database.yml
tahir-khalid: done

alan-at-work (approved):

PR #3155, opened 2024-08-07
refactor/PSD-3321:Make_change_to_overseas_regulator_change_notifying_country

Nashar101:
Made changes to Changes Overseas Regulator service call to also change the notifications notifying country in accordance to this tickets parent ticket step 2.

### DIFF
--- a/app/services/change_notification_overseas_regulator.rb
+++ b/app/services/change_notification_overseas_regulator.rb
@@ -9,7 +9,7 @@ class ChangeNotificationOverseasRegulator
     context.fail!(error: "No user supplied") unless user.is_a?(User)
 
     assign_overseas_regulator
-    return if notification.changes.none?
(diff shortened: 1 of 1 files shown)


alan-at-work (commented):
Good but we need to remove fields within the DB. Let's chat.

app/services/change_notification_overseas_regulator.rb
```diff
@@ -37,7 +37,11 @@ def activity_class
 
   def assign_overseas_regulator
     country = is_from_overseas_regulator ? overseas_regulator_country : nil
-    notification.assign_attributes(is_from_overseas_regulator:, overseas_regulator_country: country)
+    if country.present?
```
meeravnagesh, line 40: Agreed solution is to remove the notifying_country from DB

PR #3157, opened 2024-08-08
Fix/psd 2874 bulk upload notif status

tahir-khalid:
[PSD-2874](https://regulatorydelivery.atlassian.net/browse/PSD-2874)

## Description
Ensured notification is in submitted state until its actually submitted at the end of the journey.

image

image

### DIFF
--- a/app/services/create_notification.rb
+++ b/app/services/create_notification.rb
@@ -25,7 +25,7 @@ class CreateNotification
       notification.build_owner_collaborations_from(user)
 
       notification.save!
-      byebug
+
       AddProductToNotification.call!(notification:, product:, user:, skip_email: true) if product
 
       AddPrismRiskAssessmentToNotification.call!(notification:, product:, prism_risk_assessment:, user:) if prism_risk_assessment
--- a/app/services/create_notification.rb
+++ b/app/services/create_notification.rb
@@ -12,7 +12,7 @@ class CreateNotification
     notification.creator_user = user
     notification.creator_team = team
     notification.notifying_country = team.country
-    notification.state = "submitted" unless from_task_list
+    notification.state ||= "draft"
 
     ActiveRecord::Base.transaction do
       # This ensures no other pretty_id generation is happening concurrently.
@@ -25,7 +25,7 @@ class CreateNotification
       notification.build_owner_collaborations_from(user)
 
       notification.save!
-
+      byebug
       AddProductToNotification.call!(notification:, product:, user:, skip_email: true) if product
 
       AddPrismRiskAssessmentToNotification.call!(notification:, product:, prism_risk_assessment:, user:) if prism_risk_assessment


app/services/create_notification.rb
```diff
@@ -12,7 +12,7 @@ def call
     notification.creator_user = user
     notification.creator_team = team
     notification.notifying_country = team.country
-    notification.state = "submitted" unless from_task_list
```
meeravnagesh, line 15: please check if the create_notification service is called from anywhere else, as we have changed the logic , it might not work in other places

Nashar101:
> [PSD-2874](https://regulatorydelivery.atlassian.net/browse/PSD-2874)
> 
> ## Description
> Ensured notification is in submitted state until its actually submitted at the end of the journey.
> 
> ## Screen-shots or screen-capture of UI changes
> image
> 
> image

do you mean draft state until it is actually submitted ?

app/services/create_notification.rb
tahir-khalid: Restored logic and added new condition.

app/services/create_notification.rb
```diff
@@ -12,7 +12,12 @@ def call
...
+
+    notification.state = if from_task_list
```
meeravnagesh, line 16: dont think this is the right place as this is the Service, we need to change the point where this Service is being called for Bulk product upload flow

tahir-khalid: Added draft in bulk upload and ensuring state is correct in the service class,

app/services/create_notification.rb
```diff
@@ -38,6 +39,14 @@ def call
 
 private
 
+  def get_notification_state
+    notification.state = notification.state ||= if from_task_list
```
meeravnagesh, line 43: we dont need  the if condition and the draft status here as you are sending the status from bulk product upoad

meeravnagesh: notification.state = "submitted" unless from_task_list || its_product_bulk_upload - something like this

app/services/create_notification.rb
```diff
@@ -38,6 +39,14 @@ def call
 
 private
 
+  def get_notification_state
+    notification.state = "submitted" unless from_task_list || its_product_bulk_upload
+  end
+
+  def its_product_bulk_upload
+    user.can_bulk_upload_products?
```
alan-at-work, line 47: Why check the if the user can bulk upload? I think you should be checking if the flow is a `product_bulk_upload`, not the `user.can_bulk_upload_products?`

app/services/create_notification.rb
alan-at-work: Please let @meeravnagesh resolve her comment, not you doing it.

app/views/help/accessibility.html.erb
alan-at-work, line 1: This should not be in here, has someone merged and not rebased

alan-at-work:
The branch needs to be cleaned up, there are way too may commits which are the same.

tahir-khalid:
New PR created.

PR #3158, opened 2024-08-09
refactor/PSD-3321:Make_change_to_overseas_regulator_field_change_notifying_country_field

Nashar101:
Changed all references to overseas_regulator country in the code now change notifying_country instead. 
Made changes to field names for forms in order to reflect these changes, made changes to form to now use GOVUKDesignSystemFormBuilder instead of applicationformbuilder.
Made changes to activity partial so that it displays UK based countries in human form not the country code, this bug has been in the code for a long time.
Made spec changes

### DIFF
--- a/app/controllers/investigations/overseas_regulator_controller.rb
+++ b/app/controllers/investigations/overseas_regulator_controller.rb
@@ -32,7 +32,7 @@ module Investigations
     end
 
     def overseas_regulator_params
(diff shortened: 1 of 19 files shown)


alan-at-work (commented):
All looks good, are we 100% sure we are removing this now?

Nashar101:
> All looks good, are we 100% sure we are removing this now?

Yeah this is step 2 of the ticket, this needs to be done before I mogve onto the ticket where i create the migration

alan-at-work (approved):

PR #3159, opened 2024-08-09
Implemented fix and updated spec

tahir-khalid:
[PSD-2874](https://regulatorydelivery.atlassian.net/browse/PSD-2874)

## Description
Ensured notification is in submitted state until its actually submitted at the end of the journey.

image

image

[PSD-2874]: https://regulatorydelivery.atlassian.net/browse/PSD-2874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/create_notification.rb
+++ b/app/services/create_notification.rb
@@ -43,6 +43,10 @@ private
     notification.state = "submitted" unless from_task_list || its_product_bulk_upload
   end
 
+  def its_product_bulk_upload
+    user.can_bulk_upload_products?
+  end
+
   def generate_pretty_id
     "#{date.strftime('%y%m')}-#{latest_case_number_this_month.next}"
   end
--- a/app/services/create_notification.rb
+++ b/app/services/create_notification.rb
@@ -12,7 +12,8 @@ class CreateNotification
     notification.creator_user = user
     notification.creator_team = team
     notification.notifying_country = team.country
-    notification.state = "submitted" unless from_task_list
+
+    get_notification_state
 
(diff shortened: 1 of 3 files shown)


app/services/create_notification.rb
```diff
@@ -38,6 +39,14 @@ def call
 
 private
 
+  def get_notification_state
+    notification.state = "submitted" unless from_task_list || its_product_bulk_upload
+  end
+
+  def its_product_bulk_upload
```
meeravnagesh, line 46: the role user.can_bulk_upload_products? is different . it gives the ability to the user to access bulk upload, we need to check if the flow from which the this class has been reached is through bulk upload, this check does not ensure that. You can reach this class from the task list flow as well, when the user has both roles notification_task_list_user and product_bulk_uploader

meeravnagesh (approved):

app/services/create_notification.rb
meeravnagesh: please check

meeravnagesh (changes requested):

tahir-khalid:
Renamed branch raising a new PR

app/services/create_notification.rb
tahir-khalid: Implemented solution using the call stack to determine if the source matches the class name in this case create bulk upload.

PR #3167, opened 2024-08-15
refactor/PSD-3240:Update-to-product-export-helper

Nashar101:
Description

Merged multiple if statements about the same parameter into one conditional, removed compact_blank statement method from @search.countries, manually tested every single country and there will never be a blank entry to the list, made NOTIFICATION_TYPES list into a hash


Nashar101:
the reason why the notification_types method returns [Investigation::Notification, nil] is due to the API calls being made, if you remove the nil from the list the API specs fail.

alan-at-work (approved):

PR #3168, opened 2024-08-16
chore/PSD-3325-amend_ga_cookie_name

tahir-khalid:
PSD-3325

## Description
Updated cookie implementation.


Nashar101:
do rubocop -A then rebase

alan-at-work (commented):
This won't and will actually break the GA code. 

Look at prod the GA code will show, but only when 
`Rails.env.production?` is true and when `analytics_cookies_accepted?` is true

If both of these are met then the GA code is active on the page. We need access to GA to be able to look into this issue more I think.

tahir-khalid:
Needs further review.

PR #3200, opened 2024-10-01
Applied change to product recall form (additional IMT options)

tahir-khalid:
(https://regulatorydelivery.atlassian.net/browse/PSD-3561)

## Description
As an IMT user can select one from additional Notified By options and these appear in the PDF and the Markdown text.

image

image

### General testing (author)
- [x] Test without JavaScript
- [x] Test on small screen

- [x] Works keyboard only

### DIFF
--- a/app/views/products/recalls/product-details.html.erb
+++ b/app/views/products/recalls/product-details.html.erb
@@ -105,37 +105,20 @@
           hint: {text: "Only applicable if you answered yes to online marketplace"},
           small: true
       %>
-      <% if current_user.team.name = "OPSS Incident Management" %>
-        <%= form.govuk_collection_radio_buttons :notified_by,
-          [
-            OpenStruct.new(id: "OPSS Enforcement", name: "OPSS Enforcement"),
-            OpenStruct.new(id: "HSE", name: "HSE"),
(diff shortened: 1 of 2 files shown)


app/views/products/recalls/product-details.html.erb
```diff
@@ -105,17 +105,37 @@
           hint: {text: "Only applicable if you answered yes to online marketplace"},
           small: true
       %>
-      <%= form.govuk_collection_radio_buttons :notified_by,
+      <% if current_user.team.name = "OPSS Incident Management" %>
```
meeravnagesh, line 108: the check is not needed as the product recall option is available only for IMT team

app/views/products/recalls/product-details.html.erb
```diff
@@ -105,17 +105,37 @@
...
+            OpenStruct.new(id: "Environmental Health Northern Ireland",  name: "Environmental Health Northern Ireland"),
+            OpenStruct.new(id: "Ofcom",  name: "Ofcom")
           ],
           :id,
           :name,
           legend: { text: "Notified by", size: "s" },
           small: true
-      %>
+        %>      
+      <% else %>
```
meeravnagesh, line 125: No need of the else part

meeravnagesh (changes requested):

PR #3229, opened 2024-10-15
chore/PSD-3525_gtm_cookie_investigation

tahir-khalid:
Ensured GTM script is firing at correct time and cookies_policy is being set.

### DIFF
--- a/app/controllers/concerns/cookies_concern.rb
+++ b/app/controllers/concerns/cookies_concern.rb
@@ -45,13 +45,12 @@ module CookiesConcern
     cookies[:accept_analytics_cookies] = { value: accept_analytics_cookies, expires: 1.year.from_now }
     cookies[:cookie_preferences_set] = { value: accept_analytics_cookies, expires: 1.year.from_now }
 
-    if accept_analytics_cookies == true
-      cookies[:cookies_policy] = {
-        value: get_policy.to_json,
-        expires: 1.year.from_now,
-        httponly: true
-      }
-    end
+    cookies[:cookies_policy] = {
+      value: get_policy.to_json,
+      expires: 1.year.from_now,
+      httponly: true,
+      raw: true
+    }
   end
 
 private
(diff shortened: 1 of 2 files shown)


app/views/layouts/application.html.erb
```diff
@@ -14,12 +14,12 @@
     <%= javascript_include_tag "application", type: "module", nonce: true %>
     <%= stylesheet_link_tag "application" %>
     <%= csrf_meta_tags %>
-    <% if Rails.env.production? && analytics_cookies_accepted? %>
+    <% if !Rails.env.development? %>
```
alan-at-work, line 17: maybe use `<% unless Rails.env.development? %>` also why are we not checking the `analytics_cookies_accepted` now?

app/views/layouts/application.html.erb
```diff
@@ -14,12 +14,12 @@
...
       <%= render "shared/ga_head" %>
     <% end %>
   </head>
   <body class="govuk-template__body app-body-class">
-    <% if Rails.env.production? && analytics_cookies_accepted? %>
+    <% if !Rails.env.development? %>
```
alan-at-work, line 22: as above maybe use `<% unless Rails.env.development? %>`

app/controllers/concerns/cookies_concern.rb
```diff
@@ -49,7 +49,8 @@ def set_analytics_cookies(accept_analytics_cookies)
       cookies[:cookies_policy] = {
```
alan-at-work, line 49: Seems if you're not calling this is it really needed?

alan-at-work (changes requested):

app/views/layouts/application.html.erb
tahir-khalid: Reading John's email the issue seems to be the GTM container doesn't respond unless analytics cookie is accepted so the intention here is bypass that trigger the container regardless.

alan-at-work: that means you're not setting the cookie, I think they need to remove that and trust that we are dealing with it or update their code to use the one we have set.

app/controllers/concerns/cookies_concern.rb
tahir-khalid: Adjust code to ensure cookies policy is created.

tahir-khalid:
No longer required resolved in GTM 360 @ DBT

tahir-khalid:
Closing as this is not needed now they used another cookie to trigger the container.

PR #3246, opened 2024-10-28
fix/PSD-3121_issue_with_prism_journey

tahir-khalid:
## Description

PSD-3121 - resolved issue where product id would be nil in some scenarios.

### DIFF
--- a/prism/app/controllers/prism/triage_controller.rb
+++ b/prism/app/controllers/prism/triage_controller.rb
@@ -8,18 +8,25 @@ module Prism
     def index; end
 
     def serious_risk
-      @prism_risk_assessment = Prism::RiskAssessment.new
+      @serious_risk_form = Form::SeriousRisk.new(
+        risk_assessment: Prism::RiskAssessment.new
+      )
+      @prism_risk_assessment = @serious_risk_form.risk_assessment.presence
     end
 
     def serious_risk_choose
-      @prism_risk_assessment = Prism::RiskAssessment.new(serious_risk_params.merge(created_by_user_id: current_user.id).except(:investigation_id, :product_id, :product_ids))
-      if @prism_risk_assessment.save(context: :serious_risk)
-        if serious_risk_params[:investigation_id].present? && serious_risk_params[:product_ids].present?
-          associated_investigation = @prism_risk_assessment.associated_investigations.create!(investigation_id: serious_risk_params[:investigation_id])
-          serious_risk_params[:product_ids].each { |product_id| associated_investigation.associated_investigation_products.create!(product_id:) }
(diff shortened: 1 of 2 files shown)


prism/app/controllers/prism/triage_controller.rb
```diff
@@ -35,6 +35,10 @@ def serious_risk_choose
     def serious_risk_rebuttable; end
 
     def serious_risk_rebuttable_choose
+      if @prism_risk_assessment.associated_products.blank? && @prism_risk_assessment.associated_products.count.zero? && session[:serious_risk_params_product_id].present?
```
meeravnagesh, line 38: @prism_risk_assessment.&associated_products.blank? does not check for empty which is same as count zero ? 
do we need 2 conditions

prism/app/controllers/prism/triage_controller.rb
```diff
@@ -89,6 +93,10 @@ def continue_with_risk_assessment
 
   private
 
+    def set_serious_risk_params
```
meeravnagesh, line 96: With Prism being a separate module in understand there are limitations , but I have seen some problems associated with session parameter usage, explore if any other way to not loose the Product Id

prism/app/controllers/prism/triage_controller.rb
```diff
@@ -13,13 +13,13 @@ def serious_risk
 
     def serious_risk_choose
       @prism_risk_assessment = Prism::RiskAssessment.new(serious_risk_params.merge(created_by_user_id: current_user.id).except(:investigation_id, :product_id, :product_ids))
-
       if @prism_risk_assessment.save(context: :serious_risk)
         if serious_risk_params[:investigation_id].present? && serious_risk_params[:product_ids].present?
           associated_investigation = @prism_risk_assessment.associated_investigations.create!(investigation_id: serious_risk_params[:investigation_id])
           serious_risk_params[:product_ids].each { |product_id| associated_investigation.associated_investigation_products.create!(product_id:) }
         elsif serious_risk_params[:product_id].present?
           @prism_risk_assessment.associated_products.create!(product_id: serious_risk_params[:product_id])
```
meeravnagesh, line 21: if the code is already creating the associated Product , then why are we creating another one in line 39.

meeravnagesh (changes requested):

prism/app/controllers/prism/triage_controller.rb
tahir-khalid: Refactored.

prism/app/controllers/prism/triage_controller.rb
tahir-khalid: Reviewing will discuss any issues found.

PR #3253, opened 2024-11-04
Fixed Risk assessment flows (Serious, Rebuttable)

tahir-khalid:
## Description

Issue With PRISM risks assessment controller - associated products would not be preserved between screens/views.   Resolved by using the Serious Risk form as a cache store.


alan-at-work:
From what I can see, you just need to make sure the product ID is returned when the form is submitted, and it fails. 
It looks like the product ID is missing as a parameter in the URL, so you should be able to return to the URL which has the param in for example....

`https://staging.product-safety-database.service.gov.uk/prism/triage/serious-risk?product_id=12345`
fails and returns...
`https://staging.product-safety-database.service.gov.uk/prism/triage/serious-risk`

ideally we need to return to this url `https://staging.product-safety-database.service.gov.uk/prism/triage/serious-risk?product_id=12345` if it fails so the app has the product ID to use.

alan-at-work (changes requested):
From what I can see, you just need to make sure the product ID is returned when the form is submitted, and it fails. 
It looks like the product ID is missing as a parameter in the URL, so you should be able to return to the URL which has the param in for example....

`https://staging.product-safety-database.service.gov.uk/prism/triage/serious-risk?product_id=12345`
fails and returns...
`https://staging.product-safety-database.service.gov.uk/prism/triage/serious-risk`

ideally we need to return to this url `https://staging.product-safety-database.service.gov.uk/prism/triage/serious-risk?product_id=12345` if it fails so the app has the product ID to use.

prism/app/controllers/prism/triage_controller.rb
```diff
@@ -9,6 +9,7 @@ def index; end
 
     def serious_risk
       @prism_risk_assessment = Prism::RiskAssessment.new
+      cache_prism_risk_assessment_associated_products
```
alan-at-work, line 12: Not sure you need to cache anything.

alan-at-work (changes requested):

prism/app/controllers/prism/triage_controller.rb
meeravnagesh: Have you addresses this comment from Alan, now that you are having the product_id as a hidden feild , do you still need to cache

prism/app/controllers/prism/triage_controller.rb
```diff
@@ -45,7 +57,7 @@ def serious_risk_rebuttable_choose
           redirect_to risk_assessment_tasks_path(@prism_risk_assessment)
         else
           @prism_risk_assessment.update!(triage_complete: true)
-          session[:prism_risk_assessment_id] = @prism_risk_assessment.id
+          session[:prism_risk_assessment_id] = @prism_risk_assessment.i
```
meeravnagesh, line 60: what is this @prism_risk_assessment.i - is it a short hand for something?

prism/app/controllers/prism/triage_controller.rb
```diff
@@ -89,12 +101,48 @@ def continue_with_risk_assessment
 
   private
 
+    def cache_prism_risk_assessment_associated_products
```
meeravnagesh, line 104: why do we need to cache ?

meeravnagesh (changes requested):

prism/app/controllers/prism/triage_controller.rb
alan-at-work: is this a typo?

meeravnagesh (approved):

alan-at-work (approved):

PR #3270, opened 2024-11-19
Implemented Phonelib, fixed breaking tests (multiple sms calls were n…

tahir-khalid:
## Description

Sentry reporting an error: InvalidPhoneError: Not a UK mobile number (Notifications::Client::BadRequestError)

Error occurring due to number not be formatted and validated correctly.

## Solution:
- Implemented `Phonelib` for parsing
- Added strict option in `SendSMS` to allow test numbers to be used in specs
- Updated and fixed buggy tests
- Increased coverage for `SendSMS` class

### DIFF
--- a/app/services/send_sms.rb
+++ b/app/services/send_sms.rb
@@ -3,17 +3,56 @@ class SendSMS
     otp_code: "091c8861-8532-4907-abc0-f89632c34f09"
   }.freeze
 
+  Phonelib.default_country = "GB"
+  UK_PREFIX = "+44".freeze
+  UK_LEADING_ZERO = /\A0/
+
   attr_reader :client
 
-  def initialize
+  def initialize(strict: true)
     @client = Notifications::Client.new(Rails.configuration.notify_api_key)
+    Phonelib.strict_check = strict
   end
 
-  def self.otp_code(mobile_number:, code:)
-    new.client.send_sms(
-      phone_number: mobile_number,
-      template_id: TEMPLATES[:otp_code],
-      personalisation: { code: }
-    )
+  def self.otp_code(mobile_number:, code:, strict: true)
+    new(strict:).otp_code(mobile_number:, code:)
+  end
+
+  def otp_code(mobile_number:, code:)
+    if (formatted_number = validate_and_format_phone_number(mobile_number))
+      client.send_sms(
+        phone_number: formatted_number,
+        template_id: TEMPLATES[:otp_code],
+        personalisation: { code: }
+      )
+    end
+  end
+
+private
+
+  def validate_and_format_phone_number(phone_number)
+    return nil if phone_number.nil?
+
+    normalized_number = normalize_uk_number(phone_number)
+    return nil unless valid_phone_number?(normalized_number)
+
+    Phonelib.parse(normalized_number).international
+  end
+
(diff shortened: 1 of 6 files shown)


app/services/send_sms.rb
```diff
@@ -3,17 +3,59 @@ class SendSMS
...
+    return nil if phone_number.nil?
+
+    normalized_number = normalize_uk_number(phone_number)
+    return nil unless valid_phone_number?(normalized_number)
+
+    Phonelib.parse(normalized_number).international
+  end
+
+  def normalize_uk_number(phone_number)
+    return nil if phone_number.nil?
```
meeravnagesh, line 43: the code will not reach this point if the phone_number is nil, unnecessary if condition ,

app/services/send_sms.rb
```diff
@@ -3,17 +3,59 @@ class SendSMS
...
+    if phone_number.match?(UK_LEADING_ZERO)
+      return UK_PREFIX + phone_number.gsub(UK_LEADING_ZERO, "")
+    end
+
+    # If number doesn't start with + or 0, assume it needs +44
+    UK_PREFIX + phone_number
+  end
+
+  def valid_phone_number?(phone_number)
+    return false if phone_number.nil?
```
meeravnagesh, line 56: the code will not reach this point if the phone_number is nil, unnecessary if condition ,

app/services/send_sms.rb
```diff
@@ -3,17 +3,59 @@ class SendSMS
...
+
+    parsed_number = Phonelib.parse(phone_number)
+    parsed_number.valid? && parsed_number.country_code == "44" # Ensure it's a UK number
```
meeravnagesh, line 59: repeating the check , on country code , you have made sure its 44 at the end of normalise method, why do we need to check it again here

app/services/send_sms.rb
```diff
@@ -3,17 +3,59 @@ class SendSMS
...
-  def self.otp_code(mobile_number:, code:)
-    new.client.send_sms(
-      phone_number: mobile_number,
-      template_id: TEMPLATES[:otp_code],
-      personalisation: { code: }
-    )
+private
+
+  def validate_and_format_phone_number(phone_number)
+    return nil if phone_number.nil?
```
meeravnagesh, line 34: the Sentry error is on None and not Nil , hope you code also addresses the case where the Active record is returning None . Its not Cleary evident from you Rspec, please correct me if I am worng

meeravnagesh (changes requested):

app/services/send_sms.rb
tahir-khalid: Defensive coding I had created this function previously and not taken into account that the nil check is already being done further up, thank you.

app/services/send_sms.rb
tahir-khalid: Defensive coding I had created this function previously and not taken into account that the nil check is already being done further up, thank you.

app/services/send_sms.rb
tahir-khalid: This was a redundancy check have removed as its not necessary.

meeravnagesh (approved):

alan-at-work (approved):

PR #3272, opened 2024-11-20
Fix/psd 3930 action view template error

tahir-khalid:
## Description

Duplicate checks controller show view was encountering an error when the file object was nil added defensive code in the view and created spec file for coverage.

### DIFF
--- a/app/views/products/duplicate_checks/show.html.erb
+++ b/app/views/products/duplicate_checks/show.html.erb
@@ -55,7 +55,7 @@
                       </span>
                     </summary>
                       <div class="govuk-details__text">
-                        <%= image_tag(@image.file_upload, alt: sanitize(@image.file_upload.filename), class: 'opss-details-img') %>
+                        <%= image_tag(@image.file_upload, alt: sanitize(@image.file_upload&.filename&.to_s || 'Image not available'), class: 'opss-details-img') if @image&.file_upload %>
                       </div>
                   </details>
                 <% end %>


spec/views/show_html_spec.rb
```diff
@@ -0,0 +1,102 @@
+require "rails_helper"
```
meeravnagesh, line 1: we don't have specs specifically for  htmls , probably this has to be a feature test -spec\features which had tests related to these

spec/support/test_models.rb
```diff
@@ -0,0 +1,10 @@
+# spec/support/test_models.rb
```
meeravnagesh, line 1: remove the comment -comment says test_model the class name is Test_upload_model

app/views/products/duplicate_checks/show.html.erb
```diff
@@ -55,7 +55,7 @@
                       </span>
                     </summary>
                       <div class="govuk-details__text">
-                        <%= image_tag(@image.file_upload, alt: sanitize(@image.file_upload.filename), class: 'opss-details-img') %>
+                        <%= image_tag(@image.file_upload, alt: sanitize(@image.file_upload&.filename&.to_s || 'Image'), class: 'opss-details-img') if @image&.file_upload %>
```
meeravnagesh, line 58: would it be not nice to say " Image not available" as alt rather than Image - questions like what image which image will arise if its only word Image

meeravnagesh (changes requested):

spec/support/test_models.rb
tahir-khalid: Reviewing features/adding_a_duplicate_product_spec.rb to see if I can add a feature test in there.  The spec I wrote specifically tests the image control and nil and related behaviours.

meeravnagesh (approved):

PR #3292, opened 2024-12-11
fix/PSD-3997_concurrent_session_logins

tahir-khalid:
## Description

Concurrent Session Logins - The application was found to allow concurrent session logins.


config/locales/devise.security_extension.by.yml
```diff
@@ -0,0 +1,50 @@
+by:
+  errors:
```
meeravnagesh, line 2: why do we need different - so many locales , is PSD available in different languages?

config/initializers/devise_security.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+Devise.setup do |config|
+  # ==> Security Extension
+  # Configure security extension for devise
```
meeravnagesh, line 5: why is everything commented ?

docker/db/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM postgres:11
+FROM postgres:17
```
meeravnagesh, line 1: why is the postgres upgraded ? do we know if the Gov pass supports version 17

meeravnagesh (changes requested):

config/locales/devise.security_extension.by.yml
tahir-khalid: These were installed by the gym I will test if they can be removed.

config/initializers/devise_security.rb
tahir-khalid: This is how the gem provides the configuration file (removed comments).

docker/db/Dockerfile
tahir-khalid: I had to upgrade to version 17 as I was unable to get the gem to work with version 11 and later versions.   I discussed this with @alan-at-work who advised it was okay to user the latest version.

If we need to stay on version 11 then we need to review an alternative solution to using a gem (more code involved and bit as bullet proof as the gem).

app/models/user.rb
```diff
@@ -237,6 +237,11 @@ def update_last_activity_time!
 
 private
 
+  def force_session_timeout?
```
meeravnagesh, line 240: is this a common way of implementing this method. the comment says # custom logic here, and the line is return true

config/initializers/devise_security.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
```
meeravnagesh, line 1: # frozen_string_literal: true - comment

config/initializers/devise_security.rb
```diff
@@ -0,0 +1,6 @@
...
+
+Devise.setup do |config|
+  # Configuration settings (optional):
+  # https://github.com/devise-security/devise-security?tab=readme-ov-file#configuration
```
meeravnagesh, line 5: do we need these comments

db/migrate/20241211101216_add_session_limitable_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddSessionLimitableToUsers < ActiveRecord::Migration[7.1]
```
meeravnagesh, line 1: if are introducing a new table should we inform CDS, ? @alan-at-work ?

docker/db/Dockerfile
meeravnagesh: @alan-at-work can you please confirm

app/models/user.rb
tahir-khalid: Removed as its not necessary the gem is taking care of the session.

config/initializers/devise_security.rb
tahir-khalid: Reference to the gem documentation if we ever want to enable any options for the configuration.

docker/db/Dockerfile
alan-at-work: can update to 17, it will be done as part of the migration. but you will also need to test to make sure the schema works. AND a way to migrate data over. @meeravnagesh @tahir-khalid

config/initializers/devise_security.rb
alan-at-work: rubocop sometimes adds this

db/migrate/20241211101216_add_session_limitable_to_users.rb
meeravnagesh: agreeing with Alan , as its not a new table

meeravnagesh (approved):

PR #3293, opened 2024-12-11
chore/PSD-3966_push_user_id_to_data_layer

tahir-khalid:
## Description
Provide `user_id` data layer value

### DIFF
--- a/app/views/shared/_ga_head.html.erb
+++ b/app/views/shared/_ga_head.html.erb
@@ -1,3 +1,10 @@
+<script>
+    window.dataLayer = window.dataLayer || [];
+    dataLayer.push({
+      'userId': <%= Digest::SHA256.hexdigest(current_user&.id.to_s) if current_user %>
+    });
+ </script>
+
 <!-- Google Tag Manager -->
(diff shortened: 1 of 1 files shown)


app/views/shared/_ga_head.html.erb
```diff
@@ -1,3 +1,9 @@
+<script>
+    window.dataLayer = window.dataLayer || [];
+    dataLayer.push({
+      'userId': <%= current_user&.id.to_json %>
```
alan-at-work, line 4: Do we really want to be showing the user id on here, seems a little insecure tbh

maybe hash it first?

```
dataLayer.push({
  'userId': <%= Digest::SHA256.hexdigest(current_user&.id.to_s) if current_user %>
});
```

alan-at-work (changes requested):

app/views/shared/_ga_head.html.erb
tahir-khalid: Updated with
```
dataLayer.push({
  'userId': <%= Digest::SHA256.hexdigest(current_user&.id.to_s.gsub('-', '')) if current_user %>
});
```

alan-at-work (approved):

PR #3301, opened 2024-12-17
CSP file ensuring the app scripts & styles are handled correctly and …

tahir-khalid:
## Checklist:

### General testing (author)
- [x] Test without JavaScript
- [x] Test on small screen

- [x] Works keyboard only

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -1,35 +1,37 @@
 # Be sure to restart your server when you modify this file.
 
-# govuk_domains = [
-#   "*.london.cloudapps.digital", # For preview apps
-#   "*.service.gov.uk",
-# ].freeze
-# google_analytics_domains = %w[www.google-analytics.com
-#                               ssl.google-analytics.com
-#                               stats.g.doubleclick.net
-#                               www.googletagmanager.com
-#                               www.region1.google-analytics.com
-#                               region1.google-analytics.com]
-# google_static_domains = %w[www.gstatic.com]
-
+# Trusted sources
 trusted_script_sources = [:self, "https://www.googletagmanager.com"]
 trusted_style_sources  = [:self, "https://fonts.googleapis.com"]
 trusted_font_sources = [:self, "https://fonts.gstatic.com", "*.gov.uk"]
 trusted_img_sources = %i[self data]
 
+# Configuration
 Rails.application.config.content_security_policy do |policy|
+  # Core security directives
   policy.default_src :self
   policy.object_src :none
   policy.child_src :self
   policy.frame_ancestors :none
   policy.upgrade_insecure_requests true
   policy.block_all_mixed_content true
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -8,6 +8,10 @@ class ApplicationController < ActionController::Base
   include CookiesConcern
   include BreadcrumbHelper
 
+  # before_action do
```
meeravnagesh, line 11: please remove this debug lines

config/initializers/content_security_policy.rb
```diff
@@ -1,35 +1,37 @@
...
+  # Script handling with nonce support
+  nonce_only = ->(request) { "'nonce-#{request.content_security_policy_nonce}'" }
+
+  # Script directives for modules & webpack support
+  policy.script_src(*trusted_script_sources, :unsafe_eval, &nonce_only)
+  policy.script_src_elem(*trusted_script_sources, :unsafe_eval, &nonce_only)
+  policy.script_src_attr(*trusted_script_sources)
+
+  # Style directive configured for inline styles
+  policy.style_src(*trusted_style_sources, :unsafe_hashes, "'unsafe-inline'")
```
meeravnagesh, line 28: double quotes and single quotes

config/initializers/content_security_policy.rb
```diff
@@ -1,35 +1,37 @@
...
+  # Additional resource directives
   policy.font_src(*trusted_font_sources)
   policy.img_src(*trusted_img_sources)
 end
 
-# Enable CSP nonce generation
+# CSP nonce configuration
 Rails.application.config.content_security_policy_nonce_generator = ->(_request) { SecureRandom.base64(16) }
-Rails.application.config.content_security_policy_nonce_directives = %w[script-src style-src]
+Rails.application.config.content_security_policy_nonce_directives = %w[script-src script-src-elem]
```
meeravnagesh, line 37: style-src not required is it?

app/controllers/application_controller.rb
tahir-khalid: Removed thank you.

config/initializers/content_security_policy.rb
tahir-khalid: No because we are explicitly providing this at the top of the file:
```
trusted_style_sources  = [:self, "https://fonts.googleapis.com"]
```
Then defining the policy directive further down:
```
  # Style directive configured for inline styles
  policy.style_src(*trusted_style_sources, :unsafe_hashes, "unsafe-inline")
```
Normally we wouldn't do this the `style-src` directive would be enough but because of the way the current `application.html.erb` loads the inline scripts and styles we need this pattern (hopefully with the rewrite we wouldn't and shouldn't need to be so explicit the implicit policy definition should be fine).

meeravnagesh (approved):

PR #3303, opened 2024-12-18
Enabled Flipper gem

tahir-khalid:
## Description

Enable Flipper feature flag gem.


db/migrate/20241218143042_add_welcome_flipper.rb
```diff
@@ -0,0 +1,11 @@
+class AddWelcomeFlipper < ActiveRecord::Migration[7.1]
```
meeravnagesh, line 1: need to create a flipper for SubmitDraftNotificationReminder or something like that

meeravnagesh (changes requested):

tahir-khalid:
Trash PR

PR #3314, opened 2025-01-06
PSD-3667_skip_to_main_content_link_misplaced

tahir-khalid:
## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/app/views/businesses/_filters.html.erb
+++ b/app/views/businesses/_filters.html.erb
@@ -1,7 +1,4 @@
 <section class="govuk-grid-column-one-quarter govuk-!-padding-right-1 opss-full-height__col">
-  <a href="#page-content" class="govuk-skip-link opss-skip-link">
-    Skip to the main content
-  </a>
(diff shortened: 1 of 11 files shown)


mirotriad (approved):
- I've reviewed Tahir's code and I was happy that his code should resolve this ticket and I haven't done any changes to his code changes
- To make sure this works as intended I've added new tests skip_links_spec.rb
- With these new tests I've discovered three issues:
1. There were still two skip links in product page which I resolved by removing duplicate skip link in _filters.html.erb
2. Skip link was not appearing as the first child of the body element as we had some JavaScript before it. I fixed this by moving the skip link in application.html.erb so it is now very first element in the body
3. I've discovered one more issue that I didn't resolve as it is beyond the scope of this ticket, but I've noted it and will discuss it with the team whether to add it to the backlog. The issue is that the updated_at field is missing from the search index.

meeravnagesh (approved):

PR #3317, opened 2025-01-07
feat/PSD-4024-Soft-delete-drafts-after-90-days

mirotriad:
# Implement Soft Delete for Draft Notifications After 90 Days

## Description
This PR implements automatic soft deletion of draft notifications that have not been updated in the last 90 days. This helps maintain data quality by removing stale draft notifications while preserving them in the database.

## Changes
- Added `SoftDeleteDraftNotificationsJob` to handle periodic cleanup of old draft notifications
- Implemented `soft_delete_old_drafts!` method in `Investigation::Notification` model
- Created `AutomaticallyClosedCase` audit activity to track soft deletions
- Added feature flag `submit_notification_reminder` to control the functionality
- Added comprehensive test coverage for the new functionality

## Implementation Details
- Notifications are considered stale if:
  - They are in `draft` state
  - Haven't been updated in the last 90 days
- Soft deletion:
  - Sets `deleted_at` timestamp
  - Sets `deleted_by` to "System"
  - Creates an audit activity record
- Feature can be toggled using the `submit_notification_reminder` feature flag

## Testing
- Added model specs for `Investigation::Notification#soft_delete_old_drafts!`
- Added job specs for `SoftDeleteDraftNotificationsJob`
- Tested with various notification states and ages
- Verified audit trail creation
- Confirmed proper handling of the feature flag


app/models/investigation/notification.rb
```diff
@@ -61,5 +61,29 @@ def case_type
     def case_created_audit_activity_class
       AuditActivity::Investigation::AddCase
     end
+
+    DRAFT_NOTIFICATION_AGE_LIMIT = 90.days
```
meeravnagesh, line 65: move this variable to the top of the class  with other class variables

app/models/investigation/notification.rb
```diff
@@ -61,5 +61,29 @@ def case_type
...
+
+    def self.soft_delete_old_drafts!
+      Rails.logger.info "Starting to soft delete old draft notifications"
+      processed_count = 0
+
+      where(state: "draft")
```
meeravnagesh, line 71: create a scope and use that here 
for example see BulkProductsUpload

Gemfile.lock
```diff
@@ -763,6 +763,7 @@ GEM
 PLATFORMS
   arm64-darwin-22
   arm64-darwin-23
+  arm64-darwin-24
```
meeravnagesh, line 766: what is this for?

app/models/investigation/notification.rb
mirotriad: Whoops. Sorted. Thanks!

app/models/investigation/notification.rb
mirotriad: Good idea. Please check the updated code. Thanks!

Gemfile.lock
mirotriad: This was added automatically by 'bundle install' because Bundler detected that I'm running the latest macOS (Darwin) Sequoia on ARM64 architecture (M chips). It is there just to ensure that gems are compiled and installed correctly.

meeravnagesh (approved):

PR #3321, opened 2025-01-10
PSD-3957-500_error

mirotriad:
# Fix nil date handling in EnquiryDecorator source details

## Problem
When viewing certain enquiries in the system, users encounter a 500 error if the `date_received` field is nil. This occurs because the code attempts to call `to_formatted_s` on a nil value.

Error:
```ruby
ActionView::Template::Error (undefined method `to_formatted_s' for nil:NilClass)
```

## Solution
Modified the `source_details_summary_list` method in `EnquiryDecorator` to conditionally display the received date only when it exists. This is achieved by using the existing `should_display_date_received?` method as a guard clause.

## Changes
- Added conditional check before displaying received date
- Added comprehensive test coverage for various date/type combinations
- Maintained existing behavior for cases with valid dates

## Testing
Added test cases covering:
- When both date_received and received_type are present
- When date_received is nil
- When received_type is nil
- When both fields are nil

## Notes
This change affects the display of enquiry details but does not modify any data or business logic.

### DIFF
--- a/app/decorators/investigation/enquiry_decorator.rb
+++ b/app/decorators/investigation/enquiry_decorator.rb
@@ -6,23 +6,13 @@ class Investigation < ApplicationRecord
       contact_details << h.tag.p(I18n.t("case.protected_details", data_type: "#{object.case_type} contact details"), class: "govuk-body-s govuk-!-margin-bottom-1 opss-secondary-text opss-text-align-right")
 
       rows = [
-        { key: { text: "Received date" }, value: { text: date_received.to_formatted_s(:govuk) } },
-        { key: { text: "Received by" }, value: { text: received_type.upcase_first } },
-        { key: { text: "Source type" }, value: { text: complainant.complainant_type } },
+        { key: { text: "Received date" }, value: { text: date_received&.to_formatted_s(:govuk) } },
+        { key: { text: "Received by" }, value: { text: received_type&.upcase_first } },
+        { key: { text: "Source type" }, value: { text: complainant&.complainant_type } },
         { key: { text: "Contact details" }, value: { text: contact_details } }
       ]
 
       h.govuk_summary_list(rows:, borders: false, classes: "opss-summary-list-mixed opss-summary-list-mixed--narrow-dt")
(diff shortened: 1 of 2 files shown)


app/decorators/investigation/enquiry_decorator.rb
```diff
@@ -5,12 +5,11 @@ def source_details_summary_list(view_protected_details: false)
...
       contact_details << h.tag.p(I18n.t("case.protected_details", data_type: "#{object.case_type} contact details"), class: "govuk-body-s govuk-!-margin-bottom-1 opss-secondary-text opss-text-align-right")
 
-      rows = [
-        { key: { text: "Received date" }, value: { text: date_received.to_formatted_s(:govuk) } },
-        { key: { text: "Received by" }, value: { text: received_type.upcase_first } },
-        { key: { text: "Source type" }, value: { text: complainant.complainant_type } },
-        { key: { text: "Contact details" }, value: { text: contact_details } }
-      ]
+      rows = []
+      rows << { key: { text: "Received date" }, value: { text: date_received.to_formatted_s(:govuk) } } if should_display_date_received?
```
meeravnagesh, line 9: Dont use the if condition , try using this syntax date_received**&**.to_formatted_s(:govuk) , the & will call the to_formatted only if there is any value

app/decorators/investigation/enquiry_decorator.rb
```diff
@@ -5,12 +5,11 @@ def source_details_summary_list(view_protected_details: false)
...
+      rows << { key: { text: "Received by" }, value: { text: received_type.upcase_first } } if should_display_received_by?
```
meeravnagesh, line 10: use &

meeravnagesh (changes requested):

app/decorators/investigation/enquiry_decorator.rb
mirotriad: I've updated the code and this should now be resolved.

meeravnagesh (approved):

PR #3327, opened 2025-01-16
PSD-4023  send_email_reminders_to_submit_drafts based on the days elapsed

meeravnagesh:
a new job to run and check the elapsed days on drafts to send emails to the users to remind them to submit the Notification

### DIFF
--- a/app/jobs/send_reminder_email_draft_notifications_job.rb
+++ b/app/jobs/send_reminder_email_draft_notifications_job.rb
@@ -23,6 +23,7 @@ private
       send_mails(drafts, days_old, remaining_days, last_reminder, last_line)
     end
   rescue StandardError => e
+    Sentry.capture_exception(e)
     Rails.logger.error "Failed to send reminder emails for drafts days old due to: #{e.message}"
   end
 
@@ -45,7 +46,7 @@ private
         pretty_id: draft.pretty_id,
         last_reminder: last_reminder,
         last_line: last_line
-      )
+      ).deliver_later
     end
   end
 
@@ -59,9 +60,9 @@ private
 
   def reminder_last_line(last_reminder)
     if last_reminder
-      "This will be the final reminder before the notification will be automatically deleted in 3 days, if the status remains on 'Draft'."
+      I18n.t(".final_reminder", scope: "mail.send_email_reminder_to_submit_draft")
     else
-      "If the status of this notification remains in 'Draft', another reminder email will be sent in the next coming days."
+      I18n.t(".another_email", scope: "mail.send_email_reminder_to_submit_draft")
     end
   end
 end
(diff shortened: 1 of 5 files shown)


app/jobs/send_reminder_email_draft_notifications_job.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def get_draft_title(draft)
+    draft.user_title? ? draft.user_title : "Untitled Draft Notification - Last updated #{draft&.updated_at&.to_formatted_s(:govuk)}"
+  end
+
+  def reminder_last_line(last_reminder)
+    if last_reminder
+      "This will be the final reminder before the notification will be automatically deleted in 3 days, if the status remains on 'Draft'."
+    else
+      "If the status of this notification remains in 'Draft', another reminder email will be sent in the next coming days."
```
mirotriad, line 64: If we moved all text strings into en.yml just like we do it for other mailer messages, would that be more DRY and more maintainable and easier to modify in the future? So if we need to modify text changes at some point we could do it through translations (en.yml) without touching the code.

app/jobs/send_reminder_email_draft_notifications_job.rb
```diff
@@ -0,0 +1,67 @@
...
+    .where(state: "draft")
+    .where(updated_at: days_old.days.ago.beginning_of_day..days_old.days.ago.end_of_day)
+  end
+
+  def send_mails(drafts, days_old, remaining_days, last_reminder, last_line)
+    drafts.each do |draft|
+      next unless (user = find_user(draft))
+
+      draft_title = get_draft_title(draft)
+      NotifyMailer.send_email_reminder(
```
mirotriad, line 40: Would it make sense to extract send_email_reminder into a separate method for better readability?

app/jobs/send_reminder_email_draft_notifications_job.rb
```diff
@@ -0,0 +1,67 @@
...
+        user: user,
+        remaining_days: remaining_days,
+        days: days_old,
+        title: draft_title,
+        pretty_id: draft.pretty_id,
+        last_reminder: last_reminder,
+        last_line: last_line
+      )
```
mirotriad, line 48: If we used `.deliver_later` it would send emails asynchronously, which is best practice for mailers

mirotriad (commented):
Acceptance criteria met, nothings breaks and it's a very good code. I have a few comments, but it is mostly just nitpicking - I'm eager to know what you think, so please respond to my comments, but even if you decide not to make any changes, I will happily hit approve :blush:

app/jobs/send_reminder_email_draft_notifications_job.rb
meeravnagesh: I don't agree with this , as i would be simply creating a over head with variables , to just call the Notify Mailer, which is what the current method is doing as well

mirotriad (approved):

PR #3365, opened 2025-02-14
feat/PSD-3911-add_supporting_documents_to_edit_mode

mirotriad:
Upgrade add/remove supporting documents for submitted notifications to PSD 2.0


app/controllers/notifications/add_supporting_documents_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+    before_action :set_remove_document, only: [:remove_upload]
+
+    breadcrumb "notifications.label", :your_notifications
+
+    def show
+      render "add_supporting_documents"
+    end
+
+    def update
+      validate_step
```
meeravnagesh, line 19: Validate_step is already a before action , so will it not run before update , line 19 not needed right?

app/controllers/notifications/add_supporting_documents_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+
+      if save_document
+        flash_and_redirect_success
+      else
+        render "add_supporting_documents"
+      end
+    end
+
+    def remove_upload
+      validate_step
```
meeravnagesh, line 31: same comment as above

app/controllers/notifications/add_supporting_documents_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+
+    def validate_step
+      return true if policy(@notification).update? && !@notification.is_closed?
+
+      render "errors/forbidden", status: :forbidden
+      false
+    end
+
+    def validate_view_access
```
meeravnagesh, line 55: validate view access , i understand the background , but we have not implemented this for test results or supporting images , this is something we need to understand as its not specified in the requirement

app/controllers/notifications/create_controller.rb
```diff
@@ -286,6 +286,7 @@ def show
 
           return redirect_to wizard_path(:record_a_corrective_action) if @investigation_products.blank?
 
+          @manage = true # Always show the fieldset when adding a corrective action
```
meeravnagesh, line 289: why is this change ? needed

app/services/notifications/document_service.rb
```diff
@@ -0,0 +1,24 @@
+module Notifications
```
meeravnagesh, line 1: why do we need a Service?

app/views/notifications/show.html.erb
```diff
@@ -214,8 +214,12 @@
         end
         summary_list.with_row do |row|
           row.with_key(text: "Supporting documents")
-          row.with_value(text: formatted_uploads(@notification.documents).html_safe.presence || "Not provided")
-          row.with_action(text: "Change", href: investigation_supporting_information_index_path(@notification, anchor: "other"), visually_hidden_text: "supporting documents") if show_edit_link?
+          row.with_value(text: if @notification.generic_supporting_information_attachments.any?
```
meeravnagesh, line 217: previously it was formatted_uploads(@notification.documents) , where as now its generic_supporting_information_attachments why is that so?

spec/factories/notifications.rb
```diff
@@ -0,0 +1,8 @@
+FactoryBot.define do
+  factory :supporting_document_notification, class: "Investigation::Notification" do
+    sequence(:pretty_id) { |n| "2502-#{n.to_s.rjust(4, '0')}" }
```
meeravnagesh, line 3: so the pretty_id is generated based on the date , 2502 is year+month , whats the reason for this , will it work if we run the test in March?
basically no hardcoding things like these

app/controllers/notifications/add_supporting_documents_controller.rb
mirotriad: Yes, you are absolutely right about this - I've fixed it

app/controllers/notifications/add_supporting_documents_controller.rb
mirotriad: fixed, just like above in update method

app/controllers/notifications/add_supporting_documents_controller.rb
mirotriad: Yeah, this is adding inconsistency and it is redundant. Fixed

app/models/investigation.rb
```diff
@@ -162,7 +162,7 @@ def supporting_information
   # array of IDs. This allows investigations to be versioned along with their associated image
   # uploads as they were at the time of the versioned investigation.
   def image_uploads
-    ImageUpload.where(id: image_upload_ids)
+    ImageUpload.where(upload_model: self)
```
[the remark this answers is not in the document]
mirotriad: This is something I thought about a lot last night and couldn't make up my mind, but your question helped me to resolve this - I don't know whether this doesn't break the old functionality, hence I decided to revert this change, despite "upload_model: self" is the better way, it could require larger refactoring and it is beyond the scope of this ticket.

app/services/notifications/document_service.rb
mirotriad: This was my attempt to improve the code quality during [refactoring](https://github.com/OfficeForProductSafetyAndStandards/product-safety-database/pull/3365/commits/72fd1bc2cf0ddcc3e65d864c15fca47b28914e29), but I reckon it's just an overhead. I've removed it now.

app/views/notifications/show.html.erb
mirotriad: I think this was changed due to my branch not being up to date with the main, which caused some issues I had to fix this way. But now I manually reverted this part, it seems to work okay.

spec/factories/notifications.rb
mirotriad: Good point - fixed.

app/controllers/notifications/create_controller.rb
```diff
@@ -286,6 +286,7 @@ def show
 
           return redirect_to wizard_path(:record_a_corrective_action) if @investigation_products.blank?
 
+          # @manage = true # Always show the fieldset when adding a corrective action
```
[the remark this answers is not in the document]
mirotriad: this shouldn't have been there - I've removed it now, thanks

app/controllers/notifications/create_controller.rb
```diff
@@ -1184,7 +1184,7 @@ def record_a_corrective_action_params
     end
 
     def record_a_corrective_action_details_params
-      allowed_params = params.require(:corrective_action_form).permit(:action, :has_online_recall_information, :online_recall_information, :business_id, :measure_type, :details, :related_file, :existing_document_file_id, :document, legislation: [], geographic_scopes: []).merge("date_decided(1i)" => params[:corrective_action_form]["date_decided(1i)"]).merge("date_decided(2i)" => params[:corrective_action_form]["date_decided(2i)"]).merge("date_decided(3i)" => params[:corrective_action_form]["date_decided(3i)"])
```
meeravnagesh, line 1187: this change is not needed , please check , the 3i is for the year on the screen

app/controllers/notifications/add_supporting_documents_controller.rb
```diff
@@ -0,0 +1,90 @@
+module Notifications
+  class AddSupportingDocumentsController < ApplicationController
+    include BreadcrumbHelper
+
+    before_action :authenticate_user!
```
meeravnagesh, line 5: not needed

app/controllers/notifications/add_supporting_documents_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  private
+
+    def set_notification
+      @notification = Investigation.find_by!(pretty_id: params[:notification_pretty_id])
+    rescue ActiveRecord::RecordNotFound
+      head :not_found
+    end
+
+    def validate_step
```
meeravnagesh, line 45: check the validate step , from add_supporting_images

meeravnagesh (approved):

PR #3371, opened 2025-02-21
fix/PSD-3673-actions_heading_and_table_repetition

mirotriad:
- Action headings updated visually from nothing to "Change or Delete" for draft notifications and "Update" for submitted notifications
- Screen readers now correctly announce each value with its heading


meeravnagesh (approved):

meeravnagesh (changes requested):
we will wait for Prod deployment

meeravnagesh (approved):

PR #3382, opened 2025-02-27
PSD-4705: prevent drafts from being accessed via URL

mirotriad:
Prevents unauthorized access to draft notifications via direct URL

- Adds policy check to restrict draft notification access to creators and owners only
- Implements proper redirects and error handling for unauthorized access attempts
- Improves code readability with clean method extraction and better comments
- Adds comprehensive test coverage for all access scenarios

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -10,7 +10,7 @@ class NotificationsController < ApplicationController
 
   # GET /notifications
   def index
-    # Find the most recent incomplete bulk products upload for the current user, if any
+    # Retrieve incomplete bulk products upload to allow users to continue their work
     @incomplete_bulk_products_upload = BulkProductsUpload.where(user: current_user, submitted_at: nil).order(updated_at: :desc).first
 
     @pagy, @answer  = pagy_searchkick(new_opensearch_for_investigations(20, paginate: true))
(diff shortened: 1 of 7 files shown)


app/controllers/notifications_controller.rb
```diff
@@ -101,11 +101,16 @@ def assigned_notifications
 
   # GET /notifications/:pretty_id
   def show
-    return redirect_to investigation_path(@notification) unless current_user.can_use_notification_task_list?
+    if @notification.draft?
+      handle_draft_notification
+      return
```
meeravnagesh, line 106: no need of return as the method is redirecting to different page, or there is a return after render

app/controllers/notifications_controller.rb
```diff
@@ -101,11 +101,16 @@ def assigned_notifications
...
+    end
+
+    redirect_to investigation_path(@notification) and return unless current_user.can_use_notification_task_list?
```
meeravnagesh, line 109: dont think we need return for a redirect

app/policies/investigation_policy.rb
```diff
@@ -37,9 +34,8 @@ def can_unrestrict?(user: @user)
     change_owner_or_status?(user:) && record.is_private? || user.has_role?(:super_user)
   end
 
-  # Ability to see most of the details of the case, with the exception of
```
meeravnagesh, line 40: dont update with the AI generated comments unless the old comments are useless

meeravnagesh (approved):

PR #3388, opened 2025-03-05
PSD-4571 omitting drafts from product export

meeravnagesh:
omitting the draft notifications from product export

### DIFF
--- a/app/models/product_export.rb
+++ b/app/models/product_export.rb
@@ -36,19 +36,29 @@ private
   end
 
   def add_product_to_sheets(product)
-    product.investigation_products.uniq(&:investigation_id).each do |investigation_product|
+    # Filter investigations once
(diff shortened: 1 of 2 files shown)


alan-at-work (changes requested):
Tests need updating :)

app/models/product_export.rb
```diff
@@ -300,4 +310,10 @@ def restricted_field(text)
 
     "Restricted"
   end
+
+  def is_draft_notifcation(investigation)
```
mirotriad, line 314: This method name would be okay in some other languages (eg Java), but according to Ruby conventions, method names shouldn't start with prefixes like "is_", "get_" or "set_" and when the method returns a boolean value, then it should end with a question mark, so I would suggest changing the name of this method to `draft_notification?`

mirotriad (changes requested):

mirotriad (approved):
:star2: :rocket:

alan-at-work (approved):
Nicely done

PR #3389, opened 2025-03-07
PSD-4770 some more virus image nil safe handling

meeravnagesh:
added nil safe checks for virus image handling


alan-at-work (changes requested):
Add some tests please :)

alan-at-work (approved):

PR #3408, opened 2025-03-14
fix/PSD-4704-urgent_email_headers_for_final_reminder

mirotriad:
## Description

- The previous implementation is removed with its tests as it is unsupported by GOV.UK Notify.


app/mailers/notify_mailer.rb
```diff
@@ -296,7 +296,13 @@ def unsafe_attachment(user:, record_type:, id:)
 
   def send_email_reminder(user:, remaining_days:, days:, title:, pretty_id:, last_reminder:, last_line:)
     set_template(TEMPLATES[:draft_notification_reminder])
-    set_reference("Draft Notification submission reminder")
+
```
meeravnagesh, line 299: what does this change do ?

meeravnagesh: we need to update the Gov Notify template to add [Urgent] in the subject

meeravnagesh (approved):

PR #3410, opened 2025-03-17
psd 4674 500 error while adding a prism risk assessment to a notification

meeravnagesh:
changes to fix the PRISM journey from submitted notification flow and associated rspec


alan-at-work (approved):

alan-at-work (changes requested):
Does this need tests updating?

alan-at-work (approved):

PR #3412, opened 2025-03-18
feat/psd-4840-flipper_for_2fa_in_staging

mirotriad:
## PR: Implement Flipper for 2FA in Staging Environment
- Added Flipper feature to control 2FA in the Staging environment
- Created a management script (bin/toggle_2fa_in_staging.rb) to easily toggle 2FA on/off
- Updated application configuration to use Flipper in Staging while keeping environment variable control in other environments
- Added documentation on how to use the feature for load testing purposes
- Added tests to verify the implementation works correctly

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -68,7 +68,28 @@ module ProductSafetyDatabase
 
     config.antivirus_url = ENV.fetch("ANTIVIRUS_URL", "http://localhost:3006/safe")
 
-    config.secondary_authentication_enabled = ENV.fetch("TWO_FACTOR_AUTHENTICATION_ENABLED", "true") == "true"
+    # In staging, we need a way to temporarily disable 2FA for load testing
+    # while maintaining it enabled by default for security
+    if Rails.env.staging?
+      # Start with 2FA enabled for security by default
+      config.secondary_authentication_enabled = true
+
+      # Use method aliasing to maintain the original accessor while providing
+      # dynamic Flipper-based control once the app is fully initialized
+      class << config
+        alias_method :original_secondary_authentication_enabled, :secondary_authentication_enabled
+
+        def secondary_authentication_enabled
+          # Fall back to original value during app boot before Flipper is available
+          return original_secondary_authentication_enabled unless defined?(Flipper)
+
+          Flipper.enabled?(:two_factor_authentication)
+        end
+      end
+    else
+      config.secondary_authentication_enabled = ENV.fetch("TWO_FACTOR_AUTHENTICATION_ENABLED", "true") == "true"
+    end
+
     config.whitelisted_2fa_code = ENV["WHITELISTED_2FA_CODE"]
     config.vcap_application = ENV["VCAP_APPLICATION"]
     config.two_factor_attempts = 10
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -49,3 +49,9 @@ RSpec/MultipleMemoizedHelpers:
 
 RSpec/MultipleExpectations:
   Max: 2
+
+Rails/UnknownEnv:
+  Environments:
+    - development
+    - staging
+    - production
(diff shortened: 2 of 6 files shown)


config/environments/staging.rb
```diff
@@ -0,0 +1,17 @@
+# The staging environment is meant to be very similar to production
+# but allows for testing features in a production-like environment
+# before they are deployed to actual production.
+
+require_relative "production"
+
+Rails.application.configure do
```
meeravnagesh, line 7: i dont think we should be changing anything else on staging , like here we are changing the log levels and mail deliveries

config/application.rb
```diff
@@ -68,7 +68,21 @@ class Application < Rails::Application
 
     config.antivirus_url = ENV.fetch("ANTIVIRUS_URL", "http://localhost:3006/safe")
 
-    config.secondary_authentication_enabled = ENV.fetch("TWO_FACTOR_AUTHENTICATION_ENABLED", "true") == "true"
+    # For staging environment, use Flipper feature to control 2FA
+    # For other environments, continue using the environment variable
+    if Rails.env.staging?
```
meeravnagesh, line 73: do we need this check on if env.staging here , is flipper not sufficient

config/environments/staging.rb
mirotriad: yeah, good catch - you are absolutely right about this. Removed.

config/application.rb
mirotriad: Yes, we need this check to ensure Flipper only affects staging. Without it, Flipper could potentially control 2FA in all environments, which would be outside the scope of this ticket.

bin/toggle_2fa_in_staging.rb
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env ruby
```
meeravnagesh, line 1: @alan-at-work  could you review this PR , and give us feedback on if we need a script bin/toggle_2fa_in_staging.rb to toggle the flipper , we have enabled flipper UI in staging , that would be enough right

alan-at-work: I think its overkill for what can be done via the command line IMHO

mirotriad: ### Pros

1. **Automation Friendly**: Can be integrated into scripts and CI/CD pipelines
2. **Environment Safety**: Contains safeguards against execution in production
3. **Command-line Access**: Useful for terminal-only sessions where UI access isn't available
4. **Clear Messaging**: Includes warnings about the temporary nature of disabling 2FA

### Cons

1. **Redundant**: Duplicates functionality available in the Flipper UI
2. **Maintenance Overhead**: Another piece of code to maintain
3. **Limited Scope**: Only manages this specific feature, unlike the UI
4. **Requires Rails Runner**: Cannot be executed as a standalone script

bin/toggle_2fa_in_staging.rb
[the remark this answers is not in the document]
mirotriad: script removed

config/routes.rb
```diff
@@ -19,7 +19,8 @@
     mount Sidekiq::Web => "/sidekiq"
   end
 
-  mount Flipper::UI.app(Flipper) => "/flipper" if Rails.env.development?
+  # Make Flipper UI available in both development and staging environments
+  mount Flipper::UI.app(Flipper) => "/flipper" if Rails.env.development? || Rails.env.staging?
```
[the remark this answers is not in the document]
mirotriad: code updated

config/application.rb
```diff
@@ -68,7 +68,22 @@ class Application < Rails::Application
 
     config.antivirus_url = ENV.fetch("ANTIVIRUS_URL", "http://localhost:3006/safe")
 
-    config.secondary_authentication_enabled = ENV.fetch("TWO_FACTOR_AUTHENTICATION_ENABLED", "true") == "true"
+    # For staging environment, use Flipper feature to control 2FA
```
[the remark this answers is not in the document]
mirotriad: updated

alan-at-work (approved):

PR #3438, opened 2025-04-02
fix/psd-5101-update_av_code

mirotriad:
# Update to ClamAV REST Integration

## What

Updated our antivirus scanning to use the recommended `/v2/scan-chunked` endpoint from the dit-clamav-rest service.

## Changes

* Modified `AntiVirusAnalyzer` to use the correct endpoint, headers, and content format
* Updated response handling to parse JSON response properly
* Fixed and expanded tests to verify correct behavior

## Why

Improves virus scanning efficiency, especially for larger files, by using the recommended API endpoint.

## Related PR

This implements the same changes that were made in the SCPN repository (PR #3580).

### DIFF
--- a/app/analyzers/anti_virus_analyzer.rb
+++ b/app/analyzers/anti_virus_analyzer.rb
@@ -8,12 +8,15 @@ class AntiVirusAnalyzer < ActiveStorage::Analyzer
       return { error: "No file provided" } if file.nil?
 
       begin
-        uri = URI(Rails.application.config.antivirus_url)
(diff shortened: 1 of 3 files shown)


app/analyzers/anti_virus_analyzer.rb
```diff
@@ -22,7 +24,7 @@ def metadata
           case response
           when Net::HTTPSuccess
             body = JSON.parse(response.body)
-            { safe: body["safe"] }
+            { safe: !body.fetch("infected", false) }
```
meeravnagesh, line 27: why is there a difference in SCPN code and PSD code

mirotriad: SCPN uses the `malware` field instead of `infected` with more complex error handling, while this is a simpler implementation focused on the safe boolean value.

meeravnagesh (approved):

PR #3440, opened 2025-04-07
feat/psd-5037-upload-product-taxonomy

rubenarakelyan:
https://regulatorydelivery.atlassian.net/browse/PSD-5037

## Description

Adds new functionality to the Support Portal to allow users with the `update_product_taxonomy` role to upload files that update product categories and subcategories.

Screenshot 2025-04-07 at 13 47 29

Screenshot 2025-04-07 at 13 47 17

Screenshot 2025-04-07 at 13 47 37

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only


support_portal/app/views/support_portal/product_taxonomy/index.html.erb
```diff
@@ -0,0 +1,50 @@
...
+            row.with_cell(text: "Bulk upload template file")
+          end
+        end
+        table.with_body do |body|
+          @records.each do |record|
+            body.with_row do |row|
+              row.with_cell(text: record.user.name)
+              row.with_cell(text: display_date_time(record.created_at))
+              row.with_cell(text: record.status)
+              row.with_cell(text: record.import_file.attached? ? "<a href=\"#{Rails.application.routes.url_helpers.rails_storage_proxy_path(record.import_file, only_path: true)}\" class=\"govuk-link\" rel=\"noreferrer noopener\" target=\"_blank\">#{record.import_file.filename}</a>".html_safe : "-")
```
alan-at-work, line 33: we could improve readability and remove potential security risks if we do this is a separate file?

_**_file_link.html.erb**_
```ruby
<% if file.attached? %>
  <%= link_to file.filename.to_s, 
              rails_storage_proxy_path(file, only_path: true),
              class: "govuk-link",
              rel: "noreferrer noopener",
              target: "_blank" %>
<% else %>
  -
<% end %>
```

then do

```ruby
<% row.with_cell do %>
  <%= render 'file_link', file: record.import_file %>
<% end %>
```

support_portal/app/views/support_portal/product_taxonomy/new.html.erb
```diff
@@ -0,0 +1,15 @@
+<% content_for :page_title, "Upload new product taxonomy file" %>
+<% @errors = @product_taxonomy_import.errors.any? %>
+<% @back_link_href = support_root_path %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @product_taxonomy_import, url: product_taxonomy_index_path, method: :post do |f| %>
+      <%= f.govuk_error_summary %>
+      <% if @errors %><h1 class="govuk-heading-l"><%= yield :page_title %></h1><% end %>
```
alan-at-work, line 9: Should this heading only be there when there are errors?

support_portal/app/views/support_portal/product_taxonomy/taxonomy.html.erb
```diff
@@ -0,0 +1,16 @@
+<% content_for :page_title, "Current product taxonomy" %>
+<% @back_link_href = product_taxonomy_index_path %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column">
+    <%=
```
alan-at-work, line 6: Could we not use ERB tags for each subcategory instead of building HTML strings. And avoid using `.html_safe` removing the potential security risk.

```ruby
<%= govuk_accordion do |accordion| %>
  <% @records.each do |category, subcategories| %>
    <% accordion.with_section(
         heading_text: "#{category} (#{subcategories.count})"
       ) do %>
       <% subcategories.each do |subcategory| %>
         <p><%= subcategory.name %></p>
       <% end %>
    <% end %>
  <% end %>
<% end %>
 ```

**Review by alan-at-work — CHANGES_REQUESTED** [review]
Looks good, just some comments for consideration.

File: app/models/product_taxonomy_import.rb [hunk 4]
```diff
@@ -0,0 +1,131 @@
...
+    return if import_file&.content_type == "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
+
+    errors.add(:import_file, :wrong_type, message: "The selected file must be an Excel file (XLSX)")
+  end
+
+  def file_is_in_expected_format
+    return unless import_file.attached?
+
+    import_file.open do |file|
+      workbook = RubyXL::Parser.parse(file.path)
```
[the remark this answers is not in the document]
rubenarakelyan: Fixed

support_portal/app/controllers/support_portal/product_taxonomy_controller.rb
```diff
@@ -0,0 +1,69 @@
+module SupportPortal
```
meeravnagesh, line 1: Will get Ali in the loop to have a UX review on the solution that we have chosen

support_portal/app/views/support_portal/product_taxonomy/new.html.erb
rubenarakelyan: Yeah, the layout generally renders the heading unless there is an error, in which case it is deferred to the individual view because the error messages need to be rendered above the page header.

alan-at-work (changes requested):

support_portal/app/services/support_portal/update_product_taxonomy.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    def call
+      context.fail!(error: "No product taxonomy import supplied") unless product_taxonomy_import.is_a?(ProductTaxonomyImport)
+      context.fail!(error: "The product taxonomy import does not have an import file") unless product_taxonomy_import.import_file.attached?
+
+      # Update database with new product taxonomy
+      product_taxonomy_import.import_file.open do |import_file|
+        workbook = RubyXL::Parser.parse(import_file.path)
+
+        ActiveRecord::Base.transaction do
```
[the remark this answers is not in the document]
rubenarakelyan: We don't intend to have any relationship between categories/subcategories and anything else. The idea is to just take the string of the chosen category/subcategory and insert it into the product record, which also means no changes to the current schema. Since these fields cannot be changed after the product is created, it means changes to the taxonomy won't affect older products and we don't need to keep a history in these tables - they are simply for populating the dropdowns at creation time.

db/migrate/20250401135335_change_product_taxonomy_imports_user_id.rb
```diff
@@ -0,0 +1,15 @@
+class ChangeProductTaxonomyImportsUserId < ActiveRecord::Migration[7.1]
+  def up
+    safety_assured do
+      remove_column :product_taxonomy_imports, :user_id
```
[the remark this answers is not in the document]
rubenarakelyan: There's no data in this table - it was created in the previous PR to be used here.

support_portal/app/jobs/support_portal/product_taxonomy_import_job.rb
```diff
@@ -0,0 +1,8 @@
+module SupportPortal
+  class ProductTaxonomyImportJob < ApplicationJob
+    def perform(product_taxonomy_import_id)
+      product_taxonomy_import = ::ProductTaxonomyImport.find(product_taxonomy_import_id)
+      UpdateProductTaxonomy.call(product_taxonomy_import: product_taxonomy_import)
```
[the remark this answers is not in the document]
rubenarakelyan: Well...only one thing is happening here, so the scenarios are:

* `ProductTaxonomyImport` lookup fails - job fails and nothing happens so nothing to rollback, default Sidekiq retry mechanism will retry with exponential backoff in case of database connection failure etc
* `UpdateProductTaxonomy` service fails with an exception, is killed etc - everything the service does is wrapped in a database transaction so either it updates the database or nothing happens, there is no middle ground (e.g. the categories are deleted but not re-added)

In the next PR, there will be two more services here - to create the export file and update the bulk upload template. If either of those fails for any reason, I think they should be retried or we should be informed, but I don't think we should necessarily roll back everything we've done and go back to the start.

support_portal/app/jobs/support_portal/product_taxonomy_import_job.rb
```diff
@@ -0,0 +1,8 @@
+module SupportPortal
+  class ProductTaxonomyImportJob < ApplicationJob
```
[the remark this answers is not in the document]
rubenarakelyan: I think most of the previous comment applies here too. Retries will be handled by the default Sidekiq retry mechanism (which I verified locally) so I don't think we need any custom config.

There will be more failure handling in the next PR where there are more things added here that have a higher chance of failing.

support_portal/app/services/support_portal/update_product_taxonomy.rb
```diff
@@ -0,0 +1,39 @@
...
+            subcategory = row[0].value
+            next if category.blank? || subcategory.blank?
+
+            product_category = ::ProductCategory.find_or_create_by!(name: category)
+            subcategories_to_create << { name: subcategory, product_category_id: product_category.id }
+          end
+
+          ::ProductSubcategory.insert_all(subcategories_to_create)
+
+          product_taxonomy_import.mark_as_database_updated!
```
mirotriad, line 34: Should we also implement the creation of export_file and bulk_upload_template_file here? The model has states for these but they don't seem to be generated.

support_portal/app/controllers/support_portal/product_taxonomy_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+          content_type: product_taxonomy_import_params[:import_file].content_type
+        )
+        file.analyze_later
+        @product_taxonomy_import.import_file = file
+      else
+        return render :new
+      end
+
+      # sleep to give the antivirus checks a chance to be completed before running validations
+      sleep 3
```
mirotriad, line 39: Could we replace this sleep with a more reliable approach for checking file analysis completion, perhaps using Active Job callbacks or a separate job?

support_portal/app/services/support_portal/update_product_taxonomy.rb
rubenarakelyan: That functionality is in two separate services that are coming in the next PR because the tickets have been split this way.

support_portal/app/controllers/support_portal/product_taxonomy_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+    # GET /
+    def index
+      imports = ::ProductTaxonomyImport.includes(:user).all.order(created_at: :desc)
+      @records_count = imports.size
+      @pagy, @records = pagy(imports)
+    end
+
+    # GET /taxonomy
+    def taxonomy
+      @records = ::ProductSubcategory.joins(:product_category).all.group_by { |product_subcategory| product_subcategory.product_category.name }.sort
```
mirotriad, line 14: For larger datasets, would it make sense to add pagination to the taxonomy display similar to how the imports list is paginated?

app/models/product_taxonomy_import.rb
```diff
@@ -0,0 +1,136 @@
...
+  # The XLSX file that is created from a template with the latest categories and sub-categories
+  has_one_attached :bulk_upload_template_file
+
+  # This is used on the upload page in case there are no other fields to make strong params work.
+  attribute :_dummy, :string
+
+  validates :import_file, attached: true, size: { less_than: 10.megabytes, message: "The selected file must be smaller than 10MB" }
+  validates :import_file, attached: true, size: { greater_than: 1.byte, message: "The selected file must be larger than 0MB" }
+  validate :file_is_free_of_viruses
+  validate :file_is_an_excel_workbook
```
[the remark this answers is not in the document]
rubenarakelyan: I would argue that content type validation is more reliable than just looking at the file extension. Additionally, the input field in the view has an attribute to only allow XLSX extension files to be selected, and if something manages to upload a file that passes both the file extension and content type checks, the rubyXL file parsing will throw an exception that will be handled as a file format error message.

app/models/product_taxonomy_import.rb
```diff
@@ -0,0 +1,136 @@
+require "aasm"
+
```
mirotriad, line 2: Would it be helpful to add some brief class-level documentation explaining the overall purpose and workflow of this model for future developers?

support_portal/app/jobs/support_portal/product_taxonomy_import_job.rb
```diff
@@ -0,0 +1,8 @@
...
+    def perform(product_taxonomy_import_id)
```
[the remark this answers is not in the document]
rubenarakelyan: Besides retying, which already happens, I don't think there is any other recovery mechanism, since we're just updating the database inside a transaction. At that point, we'll need to investigate the failure and retry by re-uploading the file or re-running the background job.

support_portal/app/controllers/support_portal/product_taxonomy_controller.rb
meeravnagesh: this file would not be changing very often, may be 2 per year

support_portal/app/controllers/support_portal/product_taxonomy_controller.rb
rubenarakelyan: Had call and agreed to go ahead as implemented.

support_portal/app/jobs/support_portal/product_taxonomy_import_job.rb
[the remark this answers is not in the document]
alan-at-work: I would still consider improving this, and take a proactive approach rather that react when something goes a miss. 

I think something like this could work?

```ruby
module SupportPortal
  class ProductTaxonomyImportJob < ApplicationJob
    queue_as :cosmetics_imports
    
    retry_on ActiveRecord::RecordNotFound, wait: 5.seconds, attempts: 3
    retry_on StandardError, wait: 30.seconds, attempts: 3
    
    def perform(product_taxonomy_import_id)
      product_taxonomy_import = ::ProductTaxonomyImport.find(product_taxonomy_import_id)
      
      begin
        result = UpdateProductTaxonomy.call(product_taxonomy_import: product_taxonomy_import)
        
        unless result.success?
          Rails.logger.error "Import failed for ID: #{product_taxonomy_import_id}. Error: #{result.error}"
          product_taxonomy_import.update(error_message: result.error)
        end
      rescue => e
        Rails.logger.error "Error in product taxonomy import #{product_taxonomy_import_id}: #{e.message}"
        product_taxonomy_import.update(error_message: e.message)
        raise
      end
    end
  end
end
```

we'd need to add a queue to sidekiq like `:cosmetics_imports` which would keep things organised

alan-at-work (changes requested):

app/models/product_taxonomy_import.rb
```diff
@@ -0,0 +1,141 @@
...
+private
+
+  def file_is_free_of_viruses
+    # Don't run this validation unless document has been analyzed by antivirus analyzer
+    return unless import_file&.metadata&.key?("safe")
+
+    return if import_file&.metadata&.dig("safe") == true
+
+    errors.add(:import_file, :virus, message: "The selected file must be virus free")
```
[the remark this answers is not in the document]
rubenarakelyan: I looked at `DeleteUnsafeFilesJob` and it will find any unsafe files from this new model automatically and delete them so there's no extra work to be done there.

app/models/product_taxonomy_import.rb
rubenarakelyan: Done

support_portal/app/controllers/support_portal/product_taxonomy_controller.rb
rubenarakelyan: I've added pagination

support_portal/app/controllers/support_portal/product_taxonomy_controller.rb
rubenarakelyan: Yes that would be the best way forward but it will require some refactoring of how we handle uploads and saving the parent models. I will add it as a sub-task.

support_portal/app/jobs/support_portal/product_taxonomy_import_job.rb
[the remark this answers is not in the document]
rubenarakelyan: Done

spec/models/product_taxonomy_import_spec.rb
```diff
@@ -0,0 +1,12 @@
+require "rails_helper"
+
+RSpec.describe ProductTaxonomyImport do
```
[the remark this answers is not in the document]
rubenarakelyan: Done except one for the virus scanning since that needs some more thought and maybe some refactoring of how we do this scanning and wait for it.

support_portal/app/controllers/support_portal/product_taxonomy_controller.rb
mirotriad: Please share the link once you have created the sub-task so this can be marked as resolved :pray:

spec/models/product_subcategory_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      end
+    end
+
+    context "with duplicate name in a different product category" do
+      before do
+        create(:product_subcategory, name:, product_category: create(:product_category))
+      end
+
+      it "returns true" do
+        expect(product_category).to be_valid
```
alan-at-work, line 44: Shouldn't this be `expect(product_subcategory).to be_valid`? 
I would have thought wed want to test the validity of the subcategory, not the category?

support_portal/app/jobs/support_portal/product_taxonomy_import_job.rb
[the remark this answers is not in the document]
alan-at-work: Nicly done @rubenarakelyan

alan-at-work (changes requested):

support_portal/app/controllers/support_portal/product_taxonomy_controller.rb
rubenarakelyan: https://regulatorydelivery.atlassian.net/browse/PSD-5159

spec/models/product_subcategory_spec.rb
rubenarakelyan: Fixed, must have been a copy-paste oversight

spec/models/product_taxonomy_import_spec.rb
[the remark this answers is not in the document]
rubenarakelyan: https://regulatorydelivery.atlassian.net/browse/PSD-5159 is for the refactoring which should also unblock proper testing of virus scanning as part of larger scenarios

alan-at-work (approved):
Nice work mate

PR #3442, opened 2025-04-07
fix/psd-5101-av-code-update

mirotriad:
# PSD-5101: Update antivirus analyzer to match SCPN implementation

## Summary
This PR fixes issues with the antivirus scanning functionality by updating the implementation to match the proven approach used in the SCPN application.

## Changes
- Replace Net::HTTP implementation with RestClient in AntiVirusAnalyzer
- Update API request format to match the expected format of the antivirus service
- Enhance error handling and logging for better diagnostics
- Add `rest-client` gem dependency
- Add diagnostic rake tasks for testing antivirus functionality

## Testing
This change has been tested locally by:
1. Running the AntiVirusAnalyzer on existing blobs
2. Verifying that the proper API requests are made to the antivirus service
3. Confirming correct handling of API responses

## Deployment Notes
Since this adds a new gem dependency, it will require a full deployment with bundle install.

## Related PRs
References [SCPN PR #3580](https://github.com/OfficeForProductSafetyAndStandards/cosmetic-product-notifications/pull/3580) which implemented a similar fix.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,6 +26,7 @@ gem "jsbundling-rails"
 gem "mini_magick"
 gem "opensearch-ruby"
 gem "phonelib", "~> 0.10.6"
+gem "rest-client", "~> 2.1"
 
 # ActionMailer dependencies
 gem "net-imap"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -244,6 +244,7 @@ GEM
       devise (>= 4.3.0)
     diff-lcs (1.6.1)
     docile (1.4.1)
+    domain_name (0.6.20240107)
     dotenv (3.1.7)
     dotenv-rails (3.1.7)
       dotenv (= 3.1.7)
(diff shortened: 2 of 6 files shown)


lib/tasks/antivirus.rake
```diff
@@ -0,0 +1,85 @@
+namespace :antivirus do
```
meeravnagesh, line 1: because the AV is still in staging , there are no Virus files in Prod , so all the files are in staging only , can we run this as a script as supposed to task

meeravnagesh: no need of this file basically

meeravnagesh (changes requested):
no need of the rake task?

lib/tasks/antivirus.rake
mirotriad: Yes, we can run this directly via the console once it is deployed. I've removed it from the PR.

meeravnagesh (approved):

PR #3494, opened 2025-05-13
Refactor antivirus checks to be fully async

rubenarakelyan:
https://regulatorydelivery.atlassian.net/browse/PSD-5159

## Description

Refactors antivirus checks for all uploaded files to be fully async.

* For document/image uploads for notifications/products, we already only displayed safe files and infected files are automatically deleted. Remove the sleep.
* For bulk product uploads, the process continued regardless of the antivirus status. Add a step to force the check to complete before the user can continue.
* For product taxonomy uploads, add a retry mechanism in the existing asynchronous job to keep checking for a safe status before continuing, and expose any errors in the UI.

Also removes the dev antivirus container since it doesn't work with the new chunked scanning method. Defaults to the staging antivirus service instead. Credentials still need to be supplied in the relevant `.env` file.

Screenshot 2025-05-13 at 16 28 18

## Checklist:

### General testing (author)
- [ ] Test without JavaScript
- [ ] Test on small screen

- [ ] Works keyboard only

### DIFF
--- a/.env.development.example
+++ b/.env.development.example
@@ -13,8 +13,8 @@ OPENSEARCH_URL=http://localhost:9200
 
 EMAIL_WHITELIST_ENABLED=false
 
-ANTIVIRUS_URL=http://localhost:3006/safe
-ANTIVIRUS_USERNAME=av
+ANTIVIRUS_URL=https://staging.clamav.uktrade.digital
+ANTIVIRUS_USERNAME=username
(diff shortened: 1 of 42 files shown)


alan-at-work (approved):
Looks good, think we could clean up the redirects on the `documents_controller.rb` and other controllers, but thats not part of this work :)

PR #3506, opened 2025-05-20
fix/psd-5760-handle-nil-severity-in-risk-matrix

mirotriad:
# Fix PSD-5760: NoMethodError in Prism::Tasks::Outcome#show

## Description
This PR fixes a NoMethodError that occurs when viewing the "Review the overall product risk level" page in the Prism risk assessment workflow. The error occurs because `record.severity` can be nil in some cases, but the code tries to call `to_sym` on it.

## Changes
- Added nil checks for `record.severity` in the `overall_risk_level` method, returning an "unknown" risk level when severity is nil
- Added similar nil checks in the `highest_risk_level` and `combined_risk_level` methods to ensure they handle nil severity values gracefully
- All three methods now handle the nil severity case by providing a fallback to "unknown" risk level

## Testing
- Tested by running the Prism risk assessment specs, which passed successfully.
- Tested manually by adding a second harm scenario and then continuing in the form.


prism/app/helpers/prism/tasks/create_helper.rb
```diff
@@ -134,21 +134,33 @@ def overall_risk_level(harm_scenario = nil)
       return unless @harm_scenario || harm_scenario
 
       record = @harm_scenario || harm_scenario
+      return OpenStruct.new(risk_level: "unknown", risk_level_tag_html: Prism::RiskMatrixService.send(:risk_level_tag, risk_level: "unknown")) if record.severity.nil?
+
       Prism::RiskMatrixService.risk_level(probability_frequency: overall_probability_of_harm(record).probability, severity_level: record.severity.to_sym)
     end
```
meeravnagesh, line 140: severity = record.severity.nil? "unknown" :  record.severity.to_sym 

      Prism::RiskMatrixService.risk_level(probability_frequency: overall_probability_of_harm(record).probability, severity_level: severity)

